PR #174, opened 2021-10-04
sql/mysql: update driver to work with sqlspec

yonidavidson:
Update mysql driver to work with sqlspec
Rename mysql driver convert functions


internal/integration/mysql_test.go
```diff
@@ -501,3 +499,20 @@ func (s *mysqlSuite) dropTables(names ...string) {
 	_, err := s.db.Exec("DROP TABLE IF EXISTS " + strings.Join(names, ", "))
 	s.Require().NoError(err, "drop tables %q", names)
 }
+
+func decode(f string) (*db, error) {
+	res, err := schemahcl.Decode([]byte(f))
+	if err != nil {
+		return nil, err
+	}
+	s := db{}
```
a8m, line 508:
```suggestion
	var s db
```

sql/mysql/convert.go
```diff
@@ -85,42 +85,42 @@ func (d *Driver) mustFormat(t schema.Type) string {
 	return s
 }
 
-// ConvertSchema converts a schemaspec.Schema into a schema.Schema.
-func (d *Driver) ConvertSchema(spec *schemaspec.Schema, tables []*schemaspec.Table) (*schema.Schema, error) {
-	return schemautil.ConvertSchema(spec, tables, d.ConvertTable)
+// Schema converts a sqlspec.Schema into a schema.Schema.
+func (d *Driver) Schema(spec *sqlspec.Schema, tables []*sqlspec.Table) (*schema.Schema, error) {
+	return specutil.Schema(spec, tables, d.Table)
 }
```
a8m, line 91: I don't think `Driver.Schema` is a good name.

sql/mysql/convert.go
```diff
@@ -85,42 +85,42 @@ func (d *Driver) mustFormat(t schema.Type) string {
...
 
-// ConvertTable converts a schemaspec.Table to a schema.Table. Table conversion is done without converting
+// Table converts a sqlspec.Table to a schema.Table. Table conversion is done without converting
 // ForeignKeySpecs into ForeignKeys, as the target tables do not necessarily exist in the schema
-// at this point. Instead, the linking is done by the ConvertSchema function.
-func (d *Driver) ConvertTable(spec *schemaspec.Table, parent *schema.Schema) (*schema.Table, error) {
-	return schemautil.ConvertTable(spec, parent, d.ConvertColumn, d.ConvertPrimaryKey, d.ConvertIndex)
+// at this point. Instead, the linking is done by the Schema function.
+func (d *Driver) Table(spec *sqlspec.Table, parent *schema.Schema) (*schema.Table, error) {
```
a8m, line 96: ditto (e.g. `Driver.Table`).

sql/mysql/convert_test.go
```diff
@@ -584,23 +585,23 @@ func TestColumnTypeSpec(t *testing.T) {
...
+//			},
+//		},
+//	}
+//	d := &Driver{version: "8.11"}
+//	c, err := d.Column(s, nil)
+//	it, ok := c.Type.Type.(*schema.IntegerType)
+//	require.True(t, ok)
+//	require.NoError(t, err)
+//	require.Equal(t, "bigint", it.T)
+//}
```
a8m, line 607: It looks like this PR is not ready for review yet, right?

sql/sqlspec/sqlspec.go
```diff
@@ -30,6 +30,7 @@ type (
 		TypeName string                   `spec:"type" override:"type"`
 		Default  *schemaspec.LiteralValue `spec:"default" override:"default"`
 		//Overrides []*Override
```
a8m, line 32:
```suggestion
```

sql/mysql/convert_test.go
yonidavidson: We currently don't support override or I would have migrated that as well.

sql/sqlspec/sqlspec.go
yonidavidson: ditto.

sql/mysql/convert.go
yonidavidson: What would you prefer?, @rotemtam ?

PR #178, opened 2021-10-07
schema/schemaspec: add decoder and implment for HCL

rotemtam:


schema/schemaspec/schemahcl/codec.go
```diff
@@ -0,0 +1,30 @@
...
+	return &Decoder{r: r}
+}
+
+// Decode decodes the Atlas HCL document and stores it in target.
+func (d *Decoder) Decode(target interface{}) error {
+	all, err := ioutil.ReadAll(d.r)
+	if err != nil {
+		return fmt.Errorf("schemahcl: failed reading while decoding: %w", err)
+	}
+	spec, err := Decode(all)
```
yonidavidson, line 25: I know I have less Go experience but I find it confusing that in function Decode we use another function Decode that is also exported but has a different signature.
can't we make it private?

schema/schemaspec/schemahcl/context_test.go
```diff
@@ -34,7 +34,41 @@ config "defaults" {
...
+		Addr  string `spec:"addr"`
+	}
+	type Endpoint struct {
+		Name      string `spec:",name"`
+		Path      string `spec:"path"`
+		Addr      string `spec:"addr"`
+		TimeoutMs int    `spec:"timeout_ms"`
+		Retry     bool   `spec:"retry"`
+		Desc      string `spec:"description"`
+	}
```
a8m, line 49:
```suggestion
	type (
		Backend struct {
			Name  string `spec:",name"`
			Image string `spec:"image"`
			Addr  string `spec:"addr"`
		}
		Endpoint struct {
			Name      string `spec:",name"`
			Path      string `spec:"path"`
			Addr      string `spec:"addr"`
			TimeoutMs int    `spec:"timeout_ms"`
			Retry     bool   `spec:"retry"`
			Desc      string `spec:"description"`
		}
	)
```
Nicer ;)

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -17,12 +17,24 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
+// Unmarshal parses the Atlas HCL-encoded data and stores the result in the target.
+func Unmarshal(body []byte, target interface{}) error {
```
a8m, line 21:
```suggestion
func Unmarshal(data []byte, v interface{}) error {
```

https://pkg.go.dev/encoding/json#Unmarshal
https://pkg.go.dev/encoding/xml#Unmarshal

schema/schemaspec/spec.go
```diff
@@ -43,6 +43,12 @@ type (
 	Ref struct {
 		V string
 	}
+
+	// Unmarshaler is the interface that wraps the Unmarshal function.
+	Unmarshaler interface {
+		// Unmarshal decodes an Atlas DDL document into target.
+		Unmarshal(target interface{}) error
```
a8m, line 50: https://pkg.go.dev/encoding/json#Unmarshaler
https://pkg.go.dev/encoding/xml#Unmarshaler
```suggestion
		UnmarshalSpec(target interface{}) error
```

Can you elaborate on the input type? Why it's not `[]byte`?

a8m (approved):
👌🏻

PR #179, opened 2021-10-10
schema/schemaspec/schemahcl: support Marshal

rotemtam:
also fixing a big in schemaspec.Resource `Scan` where *Ref fields were not properly supported.

### DIFF
--- a/schema/schemaspec/extension.go
+++ b/schema/schemaspec/extension.go
@@ -285,8 +285,7 @@ func scanPtr(key string, r *Resource, field reflect.Value) error {
 	case reflect.TypeOf(LiteralValue{}):
 		attr.V = &LiteralValue{V: field.Elem().FieldByName("V").String()}
 	case reflect.TypeOf(Ref{}):
-		v := field.Elem().FieldByName("V").String()
-		attr.V = &Ref{V: v}
+		attr.V = &Ref{V: field.Elem().FieldByName("V").String()}
 	default:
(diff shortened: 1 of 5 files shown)


schema/schemaspec/extension.go
```diff
@@ -284,6 +279,20 @@ func (r *Resource) Scan(ext interface{}) error {
...
 
+func scanPtr(key string, r *Resource, field reflect.Value) error {
+	attr := &Attr{K: key}
+	switch field.Elem().Type() {
+	case reflect.TypeOf(LiteralValue{}):
+		attr.V = &LiteralValue{V: field.Elem().FieldByName("V").String()}
+	case reflect.TypeOf(Ref{}):
+		attr.V = &Ref{V: field.Elem().FieldByName("V").String()}
+	default:
+		return fmt.Errorf("schemaspec: unsupported pointer to %s", field.Elem().String())
```
a8m, line 290:
```suggestion
	switch e := field.Elem(); e.Type() {
	case reflect.TypeOf(LiteralValue{}):
		attr.V = &LiteralValue{V: e.FieldByName("V").String()}
	case reflect.TypeOf(Ref{}):
		attr.V = &Ref{V: e.FieldByName("V").String()}
	default:
		return fmt.Errorf("schemaspec: unsupported pointer to %s", e)
```

a8m (approved):

PR #180, opened 2021-10-11
doc: update DDL documentation

rotemtam:
updating godoc and ddl.md to reflect recent changes


doc/md/ddl.md
```diff
@@ -116,155 +116,122 @@ playlist "comedy" {
...
-	encode, err := Encode(r)
-	require.NoError(t, err)
-	expected := `show "seinfeld" {
-}
-`
-	require.EqualValues(t, expected, string(encode))
+func ExampleMarshal() {
+  type (
+    Point struct {
+      ID string `spec:",name"`
```
yonidavidson, line 173: I prefer the `atlas` word over `spec`, how can we check that it's not used anywhere else?

doc/md/ddl.md
```diff
@@ -116,155 +116,122 @@ playlist "comedy" {
...
+    Points []*Point `spec:"point"`
+  }{
+    Points: []*Point{
+      {ID: "start", X: 0, Y: 0},
+      {ID: "end", X: 1, Y: 1},
+    },
+  }
+  b, err := Marshal(&test)
+  if err != nil {
+    panic(err)
```
a8m, line 188: Maybe `log.Fatalln` instead of panic?

schema/schemaspec/schemahcl/hcl_test.go
```diff
@@ -70,3 +71,67 @@ func TestResource(t *testing.T) {
...
+	fmt.Println(string(b))
+	// Output: point "start" {
+	//   x = 0
+	//   y = 0
+	// }
+	// point "end" {
+	//   x = 1
+	//   y = 1
+	// }
+}
```
a8m, line 137:
```suggestion
	// Output:
	// point "start" {
	//   x = 0
	//   y = 0
	// }
	// point "end" {
	//   x = 1
	//   y = 1
	// }
}
```

Does this example test work with the previous format?

doc/md/ddl.md
rotemtam: Let's discuss this and do it in a separate PR.

PR #184, opened 2021-10-17
schema/schemaspec/schemahcl: marashaler and unmarshaler interfaces

rotemtam:
…ces and add default impl to package

### DIFF
--- a/schema/schemaspec/schemahcl/hcl.go
+++ b/schema/schemaspec/schemahcl/hcl.go
@@ -17,6 +17,21 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
+// Default is a default implementation of schemaspec.Marshaler and schemaspec.Unmarshaler
+// for Atlas HCL documents.
+var Default = &defaultCodec{}
+
+type defaultCodec struct {
+}
+
+func (*defaultCodec) UnmarshalSpec(data []byte, v interface{}) error {
+	return Unmarshal(data, v)
+}
+
+func (m *defaultCodec) MarshalSpec(v interface{}) ([]byte, error) {
+	return Marshal(v)
+}
+
 // Marshal returns the Atlas HCL encoding of v.
 func Marshal(v interface{}) ([]byte, error) {
 	r := &schemaspec.Resource{}
--- a/schema/schemaspec/schemahcl/hcl_test.go
+++ b/schema/schemaspec/schemahcl/hcl_test.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"testing"
 
+	"ariga.io/atlas/schema/schemaspec"
 	"github.com/stretchr/testify/require"
 )
 
@@ -136,4 +137,11 @@ func ExampleMarshal() {
 	//   x = 1
 	//   y = 1
 	// }
+}
+
+func TestInterface(t *testing.T) {
+	var (
+		_ schemaspec.Unmarshaler = &defaultCodec{}
+		_ schemaspec.Marshaler   = &defaultCodec{}
+	)
 }
--- a/schema/schemaspec/spec.go
+++ b/schema/schemaspec/spec.go
@@ -43,6 +43,18 @@ type (
 	Ref struct {
 		V string
 	}
+
+	// Marshaler is the interface implemented by types that can marshal objects into a
+	// valid Atlas DDL representation.
+	Marshaler interface {
+		MarshalSpec(v interface{}) ([]byte, error)
+	}
+
+	// Unmarshaler is the interface implemented by types that can unmarshal an Atlas DDL
+	// representation into an object.
+	Unmarshaler interface {
+		UnmarshalSpec([]byte, interface{}) error
+	}
 )
 
 // Int returns an integer from the Value of the Attr. If The value is not a LiteralValue or the value


schema/schemaspec/schemahcl/hcl.go
```diff
@@ -17,6 +17,21 @@ import (
...
+var Default = &defaultCodec{}
+
+type defaultCodec struct {
+}
+
+func (*defaultCodec) UnmarshalSpec(data []byte, v interface{}) error {
+	return Unmarshal(data, v)
+}
+
+func (m *defaultCodec) MarshalSpec(v interface{}) ([]byte, error) {
```
yonidavidson, line 31: godoc

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -17,6 +17,21 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
+// Default is a default implementation of schemaspec.Marshaler and schemaspec.Unmarshaler
+// for Atlas HCL documents.
+var Default = &defaultCodec{}
```
a8m, line 22: Why it's needed?

rotemtam: In an upcoming PR, I will add on the driver package a func 

```go
func UnmarshalSpec(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{}) error {
   // .. 
}
```

That can be invoked like:
```go
s := schema.Schema{}
err := mysql.UnmarshalSpec([]byte(f), schemahcl.Default, &s)
```

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -17,6 +17,21 @@ import (
...
+
+type defaultCodec struct {
+}
+
+func (*defaultCodec) UnmarshalSpec(data []byte, v interface{}) error {
```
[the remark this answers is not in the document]
rotemtam: ✅

schema/schemaspec/schemahcl/hcl.go
rotemtam: ✅

schema/schemaspec/schemahcl/hcl_test.go
```diff
@@ -137,3 +138,10 @@ func ExampleMarshal() {
 	//   y = 1
 	// }
 }
+
+func TestInterface(t *testing.T) {
+	var (
+		_ schemaspec.Unmarshaler = &defaultCodec{}
+		_ schemaspec.Marshaler   = &defaultCodec{}
+	)
+}
```
a8m, line 147: This can be replaced with a compile-time check.

```go
var _ interface {
	schemaspec. Marshaler
	schemaspec.Unmarshaler
} = (*defaultCodec)(nil)
```

schema/schemaspec/schemahcl/hcl.go
a8m: So maybe choose a better name, like `DefaultCodec`?

rotemtam: yeah i tried not to export the name codec because i don't like too much in this context. Codec usually is used in streaming contexts. I also didn't want to export something name "MarshalUnmarshaler" 
wdyt about exporting 

```go
var (
  Marshaler schemaspec.Marshaler = &defaultCodec{}
  Unmarshaler schemaspec.Unmarshaler = &defaultCodec{}
) 
```

a8m: `Marshaler` and `Unmarshaler` are used as interfaces in multiple standard packages. I wouldn't use them as variable names. 

I understand your comment regarding the word coded, but I suggested it since you used `decode` and `encode` inside the package. Let's chat about it for a few seconds and find a solution.

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -17,27 +17,44 @@ import (
...
 )
 
-// Marshal returns the Atlas HCL encoding of v.
-func Marshal(v interface{}) ([]byte, error) {
-	r := &schemaspec.Resource{}
-	if err := r.Scan(v); err != nil {
-		return nil, fmt.Errorf("schemahcl: failed scanning %T to resource: %w", v, err)
-	}
-	return encode(r)
+type MarshalerFunc func(interface{}) ([]byte, error)
```
a8m, line 20: Add godoc?

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -17,27 +17,44 @@ import (
...
+		r := &schemaspec.Resource{}
+		if err := r.Scan(v); err != nil {
+			return nil, fmt.Errorf("schemahcl: failed scanning %T to resource: %w", v, err)
+		}
+		return encode(r)
+	})
+	_ schemaspec.Marshaler = MarshalerFunc(nil)
+)
+
+type UnmarshalerFunc func([]byte, interface{}) error
```
a8m, line 38: ditto.

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -17,27 +17,44 @@ import (
...
+
+func (f UnmarshalerFunc) UnmarshalSpec(data []byte, v interface{}) error {
+	return f(data, v)
 }
 
+var (
+	Unmarshal = UnmarshalerFunc(func(data []byte, v interface{}) error {
```
a8m, line 45: ditto

schema/schemaspec/schemahcl/hcl.go
rotemtam: yes will do

PR #185, opened 2021-10-17
sql/internal/specutil: fix bug in Index method

rotemtam:
a bug existed in specutil.Index existed where the method would look for the referenced column literally instead of unpacking the table and column names from the ref address.


sql/internal/specutil/convert.go
```diff
@@ -107,7 +107,10 @@ func Column(spec *sqlspec.Column, conv ConvertTypeFunc) (*schema.Column, error)
 func Index(spec *sqlspec.Index, parent *schema.Table) (*schema.Index, error) {
 	parts := make([]*schema.IndexPart, 0, len(spec.Columns))
 	for seqno, c := range spec.Columns {
-		cn := c.V
+		cn, err := columnName(c)
+		if err != nil {
+			return nil, fmt.Errorf("specutil: failed converting column to index: %w", err)
```
yonidavidson, line 112: why is this log specutil but 3 lines below it is  sqlspec? asking since I think I caused this inconsistency.

a8m: A mistake?

PR #190, opened 2021-10-19
sql/mysql: support UnmarshalSpec

rotemtam:


sql/mysql/sqlspec.go
```diff
@@ -0,0 +1,222 @@
...
+		Tables  []*sqlspec.Table  `spec:"table"`
+		Schemas []*sqlspec.Schema `spec:"schema"`
+	}
+	if err := unmarshaler.UnmarshalSpec(data, &doc); err != nil {
+		return err
+	}
+	switch v := v.(type) {
+	case *schema.Schema:
+		if len(doc.Schemas) != 1 {
+			return fmt.Errorf("mysql: expecting document to contain a single schema, got %d", len(doc.Schemas))
```
yonidavidson, line 26: why ```mysql``` and not ```mysqlspec```?

sql/mysql/sqlspec_test.go
```diff
@@ -0,0 +1,282 @@
...
+			spec:     specutil.NewCol("float", "float", specutil.LitAttr("precision", "10")),
+			expected: &schema.FloatType{T: "float", Precision: 10},
+		},
+		{
+			spec:     specutil.NewCol("float", "float", specutil.LitAttr("precision", "25")),
+			expected: &schema.FloatType{T: "double", Precision: 25},
+		},
+	} {
+		t.Run(tt.spec.Name, func(t *testing.T) {
+			// todo(rotemtam): test uses unexported method, change test after implementing driver Marshal
```
yonidavidson, line 276: Are we into ```todo```ing code?

sql/internal/specutil/spec.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"strconv"
+
+	"ariga.io/atlas/schema/schemaspec"
+	"ariga.io/atlas/sql/sqlspec"
+)
+
+// NewCol is a helper method for constructing *sqlspec.Column instances.
+func NewCol(name, coltype string, attrs ...*schemaspec.Attr) *sqlspec.Column {
```
a8m, line 11:
```suggestion
func Col(name, coltype string, attrs ...*schemaspec.Attr) *sqlspec.Column {
```

Why only this function has a `New` prefix?

sql/mysql/sqlspec.go
```diff
@@ -0,0 +1,222 @@
...
+// UnmarshalSpec unmarshals an Atlas DDL document using an unmarshaler into v.
+func UnmarshalSpec(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{}) error {
+	var doc struct {
+		Tables  []*sqlspec.Table  `spec:"table"`
+		Schemas []*sqlspec.Schema `spec:"schema"`
+	}
+	if err := unmarshaler.UnmarshalSpec(data, &doc); err != nil {
+		return err
+	}
+	switch v := v.(type) {
```
a8m, line 23: Is there any reason for using a type switch and not an if statement? Do you plan to add more cases in the future?

sql/mysql/sqlspec.go
```diff
@@ -0,0 +1,222 @@
...
+		return nconvertDecimal(spec)
+	case sqlspec.TypeFloat:
+		return nconvertFloat(spec)
+	case sqlspec.TypeTime:
+		return nconvertTime(spec)
+	}
+	return parseRawType(spec.TypeName)
+}
+
+// temporarily prefixed with "n" until we complete the refactor of replacing sql/schemaspec with sqlspec.
```
a8m, line 86: 👍

sql/mysql/sqlspec_test.go
a8m: > Are we into `todo`ing code?

Not a fan of it as it usually stays there forever.

sql/mysql/sqlspec_test.go
```diff
@@ -0,0 +1,282 @@
...
+			columnType, err := convertColumnType(tt.spec)
```
a8m, line 277: Is there any reason for testing unexported functions?

a8m (approved):

sql/mysql/sqlspec_test.go
rotemtam: there's are good tests that i want to keep during the refactor
but i need to rewrite the actual test code for them since i made convertColumnType a private func.
there's a bug fixed in #192 that will make the code much nicer so in the meantime this TODO is here and the private method.

sql/mysql/sqlspec.go
```diff
@@ -0,0 +1,220 @@
...
+// UnmarshalSpec unmarshals an Atlas DDL document using an unmarshaler into v.
+func UnmarshalSpec(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{}) error {
+	var doc struct {
+		Tables  []*sqlspec.Table  `spec:"table"`
+		Schemas []*sqlspec.Schema `spec:"schema"`
+	}
+	if err := unmarshaler.UnmarshalSpec(data, &doc); err != nil {
+		return err
+	}
+	if v, ok := v.(*schema.Schema); ok {
```
a8m, line 23: Keep the “happy path” unindented.

sql/mysql/sqlspec_test.go
```diff
@@ -0,0 +1,306 @@
...
+		t.Run(tt.spec.Name, func(t *testing.T) {
+			var s schema.Schema
+			err := UnmarshalSpec(doc(tt.spec), schemahcl.Unmarshal, &s)
+			require.NoError(t, err)
+			tbl, ok := s.Table("table")
+			require.True(t, ok)
+			col, ok := tbl.Column(tt.spec.Name)
+			require.True(t, ok)
+			require.EqualValues(t, tt.expected, col.Type.Type)
+		})
```
a8m, line 286: 👌🏻👌🏻

PR #193, opened 2021-10-20
sql/mysql: support MarshalSpec

rotemtam:


sql/mysql/sqlspec.go
```diff
@@ -34,6 +38,26 @@ func UnmarshalSpec(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{
 	return fmt.Errorf("mysql: failed unmarshaling spec. %T is not supported", v)
 }
 
+// MarshalSpec marshals v into an Atlas DDL document using a schemaspec.Marshaler.
```
yonidavidson, line 41: is v a known name for interface?

rotemtam: https://cs.opensource.google/go/go/+/refs/tags/go1.17.2:src/encoding/json/encode.go;l=158

yonidavidson:
I meant in the go doc.

sql/mysql/sqlspec.go
```diff
@@ -11,12 +13,14 @@ import (
 	"ariga.io/atlas/sql/sqlspec"
 )
 
+type ddlDoc struct {
```
a8m, line 16: why not just `doc`?

sql/mysql/sqlspec.go
```diff
@@ -34,6 +38,26 @@ func UnmarshalSpec(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{
...
+	}
+	spec, tables, err := schemaSpec(s)
+	if err != nil {
+		return nil, fmt.Errorf("mysql: failed converting schema to spec: %w", err)
+	}
+	doc := &ddlDoc{
+		Tables:  tables,
+		Schemas: []*sqlspec.Schema{spec},
+	}
+	return marshaler.MarshalSpec(doc)
```
a8m, line 58:
```suggestion
	return marshaler.MarshalSpec(doc{
		Tables:  tables,
		Schemas: []*sqlspec.Schema{spec},
	})
```

PR #197, opened 2021-10-28
 sql/sqlite: unmarshal spec column types

yonidavidson:
- add tests

### DIFF
--- a/sql/sqlite/sqlspec_test.go
+++ b/sql/sqlite/sqlspec_test.go
@@ -180,13 +180,6 @@ func TestUnmarshalSpecColumnTypes(t *testing.T) {
 				Unsigned: false,
 			},
 		},
-		//{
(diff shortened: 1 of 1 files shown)


yonidavidson:
> > @a8m @rotemtam the unmarshal logic is based on what was in the convert.go , looking at the tests I am not sure if some of the results are intended or not - for example - in blob, not having the size.
> > can you please guide me on this matter? Thanks.
> 
> Looking at it now, I think its wrong not to include size for example. It's true that its meaningless in SQLite (has no impact on how things work), but it can be stored on the db schema. The problem with the way its set up now is that marshal/unmarshal are not symmetrical. e.g if you unmarshal a schema that has a column with a size, and marshal that , that information will be lost.

so as It seems in the driver we are ignoring the size on blob type. therefore the testing is correct.

PR #198, opened 2021-10-31
sql/postgres: unmarshal spec column types

yonidavidson:
- convert column type


sql/postgres/sqlspec.go
```diff
@@ -0,0 +1,189 @@
...
+	Schemas []*sqlspec.Schema `spec:"schema"`
+}
+
+// UnmarshalSpec unmarshals an Atlas DDL document using an unmarshaler into v.
+func UnmarshalSpec(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{}) error {
+	var d doc
+	if err := unmarshaler.UnmarshalSpec(data, &d); err != nil {
+		return err
+	}
+	if v, ok := v.(*schema.Schema); ok {
```
a8m, line 25: Keep the happy path unindented (same comment as for SQLite).

sql/postgres/sqlspec_test.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	tmpl := `
+schema "default" {
+}
+table "table" {
+	schema = "default"
+	%s
+}
+`
+	body := fmt.Sprintf(tmpl, string(buf))
```
a8m, line 156: ditto

sql/postgres/sqlspec.go
yonidavidson: 👍

sql/postgres/sqlspec_test.go
yonidavidson: 👍

a8m (approved):

PR #199, opened 2021-10-31
sql/postgres: convert schema

yonidavidson:
- add test layer


sql/postgres/sqlspec_test.go
```diff
@@ -13,6 +13,142 @@ import (
...
+	column "name" {
+		type = "string"
+		size = 32
+	}
+	primary_key {
+		columns = [table.accounts.column.name]
+	}
+}
+`
+	s := schema.Schema{}
```
a8m, line 63:
```suggestion
	var s schema.Schema
```

a8m (commented):
Please fix the style comment.

a8m (approved):

PR #205, opened 2021-11-03
cmd/atlas/internal/client: implement client open

yonidavidson:


cmd/atlas/internal/client/driver.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"database/sql"
+	"fmt"
+	"strings"
+
+	"ariga.io/atlas/sql/mysql"
+	"ariga.io/atlas/sql/postgres"
+	"ariga.io/atlas/sql/schema"
+	"ariga.io/atlas/sql/sqlite"
```
rotemtam, line 11: I think it's best if the driver implementations aren't coupled to the package, instead, they should be registered at runtime using a pattern similar to the way `sql` drivers are. This way users of the package can use the package for other drivers without modifying this code.

cmd/atlas/internal/client/driver.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+type (
+	// Driver provides to diff, execute and inspect against a Data Source.
+	Driver interface {
+		schema.Differ
+		schema.Execer
+		schema.Inspector
+	}
```
rotemtam, line 20: I don't think this interface is needed, a type that embeds these interfaces as you did below is fine.

cmd/atlas/internal/client/driver.go
```diff
@@ -0,0 +1,109 @@
...
+	// AtlasDriver implements the Driver interface using Atlas.
+	AtlasDriver struct {
+		schema.Differ
+		schema.Execer
+		schema.Inspector
+	}
+
+	dbName string
```
rotemtam, line 28: unless you plan to add methods to this type, using string is perfectly fine. if you accept what I suggested about drivers being registered instead of explicitly imported in compile time, this isn't needed.

cmd/atlas/internal/client/driver.go
```diff
@@ -0,0 +1,109 @@
...
+	mysqlDB    dbName = "mysql"
+	postgresDB dbName = "postgres"
+	sqliteDB   dbName = "sqlite3"
+)
+
+var providers = map[dbName]func(string) (*AtlasDriver, func(), error){
+	mysqlDB:    atlasDriverMysql,
+	postgresDB: atlasDriverPostgres,
+	sqliteDB:   atlasDriverSqlite,
+}
```
rotemtam, line 41: same for this.

cmd/atlas/internal/client/driver.go
```diff
@@ -0,0 +1,109 @@
...
+
+// NewAtlasDriver connects a new Atlas Driver returns AtlasDriver and a closer.
+func NewAtlasDriver(dsn string) (*AtlasDriver, func(), error) {
```
rotemtam, line 44: the close function can be a public method of `*AtlasDriver`

cmd/atlas/internal/client/driver.go
```diff
@@ -0,0 +1,109 @@
...
+
+type (
+	// Driver provides to diff, execute and inspect against a Data Source.
+	Driver interface {
+		schema.Differ
+		schema.Execer
+		schema.Inspector
+	}
+	// AtlasDriver implements the Driver interface using Atlas.
+	AtlasDriver struct {
```
rotemtam, line 22: consider what will happen when you have more subcommand in the client, this struct will become huge and have a lot of dependencies. dependencies lead to coupling, and coupling leads to complexity. (complexity leads to the dark side, said master yoda)

It's better to have many specialized types for each group of operations. i think it would make sense to have a struct for `Schema` related commands (currently inspect and apply). later when we add other domains of functionality, add more specific structs.

cmd/atlas/internal/client/driver.go
```diff
@@ -0,0 +1,109 @@
...
+
+var providers = map[dbName]func(string) (*AtlasDriver, func(), error){
+	mysqlDB:    atlasDriverMysql,
+	postgresDB: atlasDriverPostgres,
+	sqliteDB:   atlasDriverSqlite,
+}
+
+// NewAtlasDriver connects a new Atlas Driver returns AtlasDriver and a closer.
+func NewAtlasDriver(dsn string) (*AtlasDriver, func(), error) {
+	a := strings.Split(dsn, "://")
```
rotemtam, line 45: use the `url` package and take the `Scheme` value 
https://cs.opensource.google/go/go/+/refs/tags/go1.17.2:src/net/url/url.go;l=359

cmd/atlas/internal/client/driver.go
```diff
@@ -0,0 +1,109 @@
...
+	if len(a) != 2 {
+		return nil, nil, fmt.Errorf("failed to parse %s", dsn)
+	}
+	p := providers[dbName(a[0])]
+	if p == nil {
+		return nil, nil, fmt.Errorf("failed to parse %s", dsn)
+	}
```
rotemtam, line 52:
```suggestion
	p, ok := providers[dbName(a[0])]
	if !ok {
		return nil, nil, fmt.Errorf("failed to parse %s", dsn)
	}
```

cmd/atlas/internal/client/driver.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	drv, err := mysql.Open(db)
+	if err != nil {
+		return nil, nil, err
+	}
+	return &AtlasDriver{
+		drv.Diff(),
```
rotemtam, line 69: initialization of fields is better done explicitly `Differ: drv.Diff()`

cmd/atlas/internal/client/driver.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, nil, err
+	}
+	drv, err := postgres.Open(db)
+	if err != nil {
+		return nil, nil, err
+	}
+	return &AtlasDriver{
+		drv.Diff(),
+		drv.Migrate(),
+		drv,
```
rotemtam, line 89: ditto

cmd/atlas/internal/client/schema.go
```diff
@@ -0,0 +1,15 @@
+package client
+
+import (
+	"context"
+)
+
+// Inspect returns the schema description by its name.
+func Inspect(ctx context.Context, d Driver) ([]byte, error) {
+	return nil, nil
+}
```
rotemtam, line 10: these types should be associated with your struct. 
also if they return []byte, you must specify a `schemaspec.Marshaler` to specify under which syntax the transformation to bytes should happen.

cmd/atlas/internal/client/schema.go
```diff
@@ -0,0 +1,15 @@
...
+
+// Apply migrates the data source to match the requested schema.
+func Apply(ctx context.Context, d Driver, s []byte) error {
+	return nil
```
rotemtam, line 14: ditto

rotemtam (changes requested):

cmd/atlas/internal/client/doc.go
```diff
@@ -0,0 +1,7 @@
+// Package client provides abstraction over the usage of Atlas.
+// Example usage of client:
+//		d,close,_ := client.NewAtlasDriver(dsn)
+//		defer close()
+//		data := client.Inspect(ctx, d)
+//
+package client
```
a8m, line 7: `cmd` is not importable. Can we start with developing stuff on `cmd/atlas` and move them to a shared place when it's necessary? I also think that the package name `client` is too generic (like `util`) and should not exist as a namespace.

cmd/atlas/internal/client/driver.go
a8m: A reference for go-cloud implementation 
https://github.com/google/go-cloud/blob/master/internal/openurl/openurl.go

yonidavidson:
Decided to implement first the CLI.

PR #207, opened 2021-11-07
cmd/atlas: add atlas driver provider

yonidavidson:


cmd/atlas/internal/url/provider.go
```diff
@@ -0,0 +1,52 @@
+package url
```
a8m, line 1: I'd avoid using the word `url` as a package, because it's widely used as an std package. Also, keep it simple (for now) and move this logic to a single file under the `cmd/atlas` package.

cmd/atlas/internal/url/provider.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &Atlas{
+		DB:        db,
+		Differ:    drv.Diff(),
+		Execer:    drv.Migrate(),
+		Inspector: drv,
+	}, nil
+}
```
a8m, line 49: I wonder if the code duplication can be avoided here. Maybe after we'll have a standard interface for all drivers.

cmd/atlas/internal/url/url.go
```diff
@@ -0,0 +1,85 @@
...
+
+type (
+	providerMap struct {
+		m map[string]func(string) (*Atlas, error)
+	}
+
+	// Mux is used for routing dsn to correct provider.
+	Mux struct {
+		providers *providerMap
+	}
```
a8m, line 19:
```suggestion
	// Mux is used for routing dsn to correct provider.
	Mux struct {
		providers map[string]func(string) (*Atlas, error)
	}
```

Why the `providerMap` is needed?

cmd/atlas/internal/url/url.go
```diff
@@ -0,0 +1,85 @@
...
+		Differ    schema.Differ
+		Execer    schema.Execer
+		Inspector schema.Inspector
+	}
+)
+
+// Close release all resources taken by Driver.
+func (a *Atlas) Close() error {
+	return a.DB.Close()
+}
```
a8m, line 33: Can't you just embed `DB`?

cmd/atlas/internal/url/url.go
```diff
@@ -0,0 +1,85 @@
...
+		m map[string]func(string) (*Atlas, error)
+	}
+
+	// Mux is used for routing dsn to correct provider.
+	Mux struct {
+		providers *providerMap
+	}
+
+	// Atlas implements the Driver interface using Atlas.
+	Atlas struct {
```
a8m, line 22: Driver?

cmd/atlas/internal/url/url.go
```diff
@@ -0,0 +1,85 @@
...
+
+var defaultURLMux *Mux
+
+// DefaultURLMux returns the default system "Mux".
+func DefaultURLMux() *Mux {
+	if defaultURLMux == nil {
+		defaultURLMux = NewMux()
+	}
+	return defaultURLMux
+}
```
a8m, line 52:
```suggestion
var defaultURLMux = NewMux()

// DefaultURLMux returns the default system "Mux".
func DefaultURLMux() *Mux {
	return defaultURLMux
}
```

Why do you need the `DefaultURLMux` func? Also, decide if it's `URLMux` or just `Mux`.

cmd/atlas/internal/url/url.go
```diff
@@ -0,0 +1,85 @@
...
+	m.m[key] = p
+}
+
+// RegisterProvider is used to register an Atlas provider by key..
+func (u *Mux) RegisterProvider(key string, p func(string) (*Atlas, error)) {
+	u.providers.register(key, p)
+}
+
+// OpenAtlas is used for opening an atlas driver on a specific data source.
+func (u *Mux) OpenAtlas(dsn string) (*Atlas, error) {
```
a8m, line 67: 👍

cmd/atlas/internal/url/provider.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"ariga.io/atlas/sql/mysql"
+	"ariga.io/atlas/sql/postgres"
+)
+
+func init() {
+	d := DefaultURLMux()
+	d.RegisterProvider("mysql", mysqlProvider)
+	d.RegisterProvider("postgres", postgresProvider)
+	d.RegisterProvider("sqlite3", sqliteProvider)
```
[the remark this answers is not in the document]
yonidavidson: 👍

cmd/atlas/internal/url/provider.go
yonidavidson: 👍

cmd/atlas/internal/url/provider.go
yonidavidson: Agree, but I want to see how the other drivers will init.

cmd/atlas/internal/url/url.go
yonidavidson: 👍

cmd/atlas/internal/url/url.go
yonidavidson: Mux,
This pattern is used by the go-cloud package, I think it's to avoid using a var straight from the package.

cmd/atlas/internal/mux/mux.go
```diff
@@ -0,0 +1,72 @@
...
+	if !ok {
+		return nil, fmt.Errorf("could not find provider, %s", err)
+	}
+	return p(dsn)
+}
+
+func parseDSN(url string) (string, string, error) {
+	a := strings.Split(url, "://")
+	if len(a) != 2 {
+		return "", "nil", fmt.Errorf("failed to parse dsn")
```
a8m, line 69:
```suggestion
		return "", "", fmt.Errorf("failed to parse dsn")
``

??

File: cmd/atlas/internal/mux/mux.go [hunk 10]
```diff
@@ -0,0 +1,72 @@
+package mux
```
a8m, line 1: Move it to atlas as well. Or, use a different naming such as `mux.URL`, `openurl.Mux`,

cmd/atlas/internal/mux/mux.go
```diff
@@ -0,0 +1,72 @@
...
+		Differ    schema.Differ
+		Execer    schema.Execer
+		Inspector schema.Inspector
+	}
+)
+
+// Close release all resources taken by Driver.
+func (a *Driver) Close() error {
+	return a.DB.Close()
+}
```
a8m, line 29: not needed.

cmd/atlas/internal/mux/mux.go
```diff
@@ -0,0 +1,72 @@
...
+		providers: make(map[string]func(string) (*Driver, error)),
+	}
+}
+
+var defaultURLMux = NewMux()
+
+// DefaultMux returns the default system "Mux".
+func DefaultMux() *Mux {
+	return defaultURLMux
+}
```
a8m, line 43: See `http.DefaultClient` as an example.

cmd/atlas/internal/mux/mux.go
yonidavidson: moved to atlas

cmd/atlas/internal/mux/mux.go
yonidavidson: 👍 
why? cli closing is good enough?

yonidavidson: or just using the API directly?

cmd/atlas/internal/mux/mux.go
yonidavidson: so to remove it that users won't get confused and think they get a default configuration?

cmd/atlas/internal/mux/mux.go
rotemtam: when you embed a type you can call its methods directly

cmd/atlas/mux.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+// NewMux returns a new Mux.
+func NewMux() *Mux {
+	return &Mux{
+		providers: make(map[string]func(string) (*Driver, error)),
+	}
+}
+
+var defaultURLMux = NewMux()
```
a8m, line 33:
```suggestion
var defaultMux = NewMux()
```

cmd/atlas/provider.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"database/sql"
+
+	"ariga.io/atlas/sql/mysql"
+	"ariga.io/atlas/sql/postgres"
+)
+
+func init() {
+	defaultURLMux.RegisterProvider("mysql", mysqlProvider)
+	defaultURLMux.RegisterProvider("postgres", postgresProvider)
```
a8m, line 12:
```suggestion
	defaultMux.RegisterProvider("mysql", mysqlProvider)
	defaultMux.RegisterProvider("postgres", postgresProvider)
```

a8m (approved):

PR #208, opened 2021-11-08
cmd/atlas/inspect: implement

yonidavidson:


cmd/atlas/inspect.go
```diff
@@ -12,16 +16,17 @@ var (
 	inspectCmd = &cobra.Command{
 		Use:   "inspect",
 		Short: "Inspect atlas schema.",
-		Run:   func(cmd *cobra.Command, args []string) {},
+		Run:   func(cmd *cobra.Command, args []string) { runInspect(inspectFlags.dsn) },
```
rotemtam, line 19:
```suggestion
		Run:   func(cmd *cobra.Command, args []string) { 
			runInspect(inspectFlags.dsn) 
		},
```

cmd/atlas/inspect.go
```diff
@@ -12,16 +16,17 @@ var (
...
 		Example: `
-atlas schema inspect -d mysql://user:pass@host:port/dbname
-atlas schema inspect -d postgres://user:pass@host:port/dbname
-atlas schema inspect --dsn sqlite3://path/to/dbname.sqlite3`,
+atlas schema inspect -d "mysql://user:pass@host:port/dbname"
+atlas schema inspect -d "postgres://user:pass@host:port/dbname"
+atlas schema inspect --dsn "sqlite3://path/to/dbname.sqlite3"`,
```
rotemtam, line 23: why is this necessary?

cmd/atlas/mux.go
```diff
@@ -17,9 +18,10 @@ type (
 	// Driver implements the Atlas interface.
 	Driver struct {
 		*sql.DB
-		Differ    schema.Differ
-		Execer    schema.Execer
-		Inspector schema.Inspector
+		Differ      schema.Differ
+		Execer      schema.Execer
+		Inspector   schema.Inspector
+		MarshalSpec func(v interface{}, marshaler schemaspec.Marshaler) ([]byte, error)
```
rotemtam, line 24: 1. use `schemaspec.Marshaler`
2. i dont think the driver should deal with the syntax that its caller wants to see. the layer above should deal w/formatting

cmd/atlas/inspect.go
```diff
@@ -31,3 +36,16 @@ func init() {
 	)
 	cobra.CheckErr(inspectCmd.MarkFlagRequired("dsn"))
 }
+
+func runInspect(dsn string) {
+	ctx := context.Background()
+	a, err := defaultMux.OpenAtlas(dsn)
+	cobra.CheckErr(err)
```
rotemtam, line 43: feels weird to me to use a library for such a simple thing why not 
```suggestion
	if a, err := defaultMux.OpenAtlas(dsn); err != nil {
	  log.Fatalf("Error: %s", err)
	}
```
Or similar?

cmd/atlas/inspect.go
```diff
@@ -12,16 +16,17 @@ var (
...
-atlas schema inspect --dsn sqlite3://path/to/dbname.sqlite3`,
+atlas schema inspect -d "mysql://user:pass@host:port/dbname"
+atlas schema inspect -d "postgres://user:pass@host:port/dbname"
+atlas schema inspect --dsn "sqlite3://path/to/dbname.sqlite3"`,
 	}
 )
 
 func init() {
 	schemaCmd.AddCommand(inspectCmd)
+	schemaCmd.SetOut(os.Stdout)
```
rotemtam, line 29: What's the benefit of going through cobra and coupling ourselves to it instead of just writing to stdout? 
Remember, complexity comes from two things: 1. coupling 2. obscurity - here you get both

cmd/atlas/inspect.go
```diff
@@ -31,3 +36,16 @@ func init() {
...
+
+func runInspect(dsn string) {
+	ctx := context.Background()
+	a, err := defaultMux.OpenAtlas(dsn)
+	cobra.CheckErr(err)
+	name, err := schemaNameFromDSN(dsn)
+	cobra.CheckErr(err)
+	schema, err := a.Inspector.InspectSchema(ctx, name, nil)
+	cobra.CheckErr(err)
+	ddl, err := a.MarshalSpec(schema, schemahcl.Marshal)
```
rotemtam, line 48: see comment on  coupling driver to serialization. 
also the syntax (HCl) will be a parameter in the near future. guessing people will want JSON soon to build tooling on top

rotemtam (changes requested):

cmd/atlas/inspect.go
yonidavidson: zsh shell has problems with url that is domain without string.

cmd/atlas/inspect.go
yonidavidson: it's the cobra std way, and its less lines.

cmd/atlas/inspect.go
yonidavidson: I'll move it to a printer module but it's important to make sure what goes via std error and what in stdout. by using cobra the cli is consistent. (think about users doing atlas something something > atlas.hcl

cmd/atlas/inspect.go
yonidavidson: agree but for now I'll make it default.

cmd/atlas/inspect.go
yonidavidson: https://github.com/ohmyzsh/ohmyzsh/issues/31

a8m: I work with zsh and it works for me. 

```go
// go run main.go --dsn sqlite3://path/to/dbname.sqlite3
// prints: [.../exe/main --dsn sqlite3://path/to/dbname.sqlite3]
func main() {
	fmt.Println(os.Args)
}
```

cmd/atlas/mux.go
```diff
@@ -60,3 +62,12 @@ func parseDSN(url string) (string, string, error) {
...
 }
+
+func schemaNameFromDSN(url string) (string, error) {
+	_, dsn, err := parseDSN(url)
+	if err != nil {
+		return "", err
+	}
+	a := strings.SplitAfter(dsn, "/")
+	return a[len(a)-1], nil
+}
```
a8m, line 73: qq: why not use a proper parser here? Both MySQL and PostgreSQL drivers have diff. Let me know if you need help with this.

cmd/atlas/inspect.go
a8m: btw, remove sqlite from the examples as it's not supported atm

cmd/atlas/mux.go
a8m: https://pkg.go.dev/github.com/jackc/pgconn#Config

https://pkg.go.dev/github.com/go-sql-driver/mysql#Config

yonidavidson: 👍

cmd/atlas/mux.go
yonidavidson: For now, in atlas this part is in the driver so it's attached here.

a8m (approved):

cmd/atlas/mux.go
```diff
@@ -17,9 +21,19 @@ type (
...
-		Execer    schema.Execer
-		Inspector schema.Inspector
+		Differ      schema.Differ
+		Execer      schema.Execer
+		Inspector   schema.Inspector
+		MarshalSpec func(v interface{}, marshaler schemaspec.Marshaler) ([]byte, error)
+	}
+
+	schemaMarshal struct {
+		driver    *Driver
```
rotemtam, line 31: printer? 
also dont need access to the full driver (coupling) use the MarshalSpec method directly

cmd/atlas/inspect.go
```diff
@@ -31,3 +38,14 @@ func init() {
 	)
 	cobra.CheckErr(inspectCmd.MarkFlagRequired("dsn"))
 }
+
+func inspectRun(a *Driver, m schemaMarshaler, dsn string) {
+	ctx := context.Background()
+	name, err := schemaNameFromDSN(dsn)
+	cobra.CheckErr(err)
+	s, err := a.Inspector.InspectSchema(ctx, name, nil)
```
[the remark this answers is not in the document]
yonidavidson: it won't, you asked me to put the inspector on a field and not embed it.

cmd/atlas/mux.go
yonidavidson: Done

PR #215, opened 2021-11-11
schema/schemaspec/hcl: fix list of ref encode

yonidavidson:


schema/schemaspec/schemahcl/hcl_test.go
```diff
@@ -109,17 +110,18 @@ show "seinfeld" {
 func ExampleMarshal() {
 	type (
 		Point struct {
-			ID string `spec:",name"`
-			X  int    `spec:"x"`
-			Y  int    `spec:"y"`
+			ID string            `spec:",name"`
+			X  int               `spec:"x"`
+			Y  int               `spec:"y"`
```
rotemtam, line 115: Please leave the example test as is (it doesn't make sense for a point to have a Z value that is a ref).

Instead add a specific test in https://github.com/ariga/atlas/blob/master/schema/schemaspec/schemahcl/context_test.go

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -238,9 +238,16 @@ func writeAttr(attr *schemaspec.Attr, body *hclwrite.Body) error {
 	case *schemaspec.ListValue:
 		lst := make([]string, 0, len(v.V))
 		for _, item := range v.V {
-			val, err := schemaspec.StrVal(item)
-			if err != nil {
-				return err
+			var val string
+			var err error
+			switch v := item.(type) {
+			case *schemaspec.Ref:
```
rotemtam, line 244: Can we use an if else (or if-return) block (with type assertion) if we only have 2 cases ?

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -238,9 +238,16 @@ func writeAttr(attr *schemaspec.Attr, body *hclwrite.Body) error {
 	case *schemaspec.ListValue:
 		lst := make([]string, 0, len(v.V))
 		for _, item := range v.V {
-			val, err := schemaspec.StrVal(item)
-			if err != nil {
-				return err
+			var val string
+			var err error
```
a8m, line 242:
```suggestion
			var (
				v string
				err error
			)
```

See Rotem's comment, but group vars into one block.

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -238,9 +238,18 @@ func writeAttr(attr *schemaspec.Attr, body *hclwrite.Body) error {
...
+			)
+			v, ok := item.(*schemaspec.Ref)
+			if ok {
+				val = strings.ReplaceAll(v.V, "$", "")
+			} else {
+				val, err = schemaspec.StrVal(item)
+				if err != nil {
+					return err
+				}
 			}
```
a8m, line 253:
```suggestion
			if v, ok := item.(*schemaspec.Ref); ok {
				val = strings.ReplaceAll(v.V, "$", "")
			} else if val, err = schemaspec.StrVal(item); err != nil {
				return err
			}
```

schema/schemaspec/schemahcl/context_test.go
```diff
@@ -301,3 +301,30 @@ person "jane" {
 	}
 	require.EqualValues(t, jane, test.People[1])
 }
+
+func TestSchemaRefParse(t *testing.T) {
+	type (
+		Point struct {
+			Z []*schemaspec.Ref `spec:"z"`
+		}
+	)
```
a8m, line 310:
```suggestion
	type Point struct {
		Z []*schemaspec.Ref `spec:"z"`
	}
```

a8m (approved):

rotemtam:
Just realized we both fixed the same bug
https://github.com/ariga/atlas/pull/219/files

PR #216, opened 2021-11-11
sql/mysql/sqlspec: fix type conversion

yonidavidson:


sql/mysql/sqlspec_test.go
```diff
@@ -331,6 +331,13 @@ func TestMarshalSpecColumnType(t *testing.T) {
 			},
 			expected: specutil.NewCol("column", "int8"),
 		},
+		{
+			schem: &schema.IntegerType{
+				T:        tMediumInt,
+				Unsigned: false,
+			},
+			expected: specutil.NewCol("column", "int24"),
```
rotemtam, line 339: No such virtual type "uint24", this is a specific mysql int size that doesn't have a corresponding setting on other dbs so we chose to omit it

sql/mysql/sqlspec.go
```diff
@@ -322,13 +322,15 @@ func integerSpec(t *schema.IntegerType) (*sqlspec.Column, error) {
...
+	case tMediumInt:
+		return &sqlspec.Column{TypeName: "int24"}, nil
 	case tBigInt:
 		if t.Unsigned {
 			return specutil.NewCol("", "uint64"), nil
 		}
 		return &sqlspec.Column{TypeName: "int64"}, nil
 	}
-	return nil, errors.New("mysql: schema integer failed to convert")
+	return nil, fmt.Errorf("mysql: schema integer failed to convert %s", t.T)
```
rotemtam, line 333: I think %q (quoted) is easier to read

sql/mysql/sqlspec.go
```diff
@@ -322,13 +322,15 @@ func integerSpec(t *schema.IntegerType) (*sqlspec.Column, error) {
 		return &sqlspec.Column{TypeName: "int"}, nil
 	case tTinyInt:
 		return &sqlspec.Column{TypeName: "int8"}, nil
+	case tMediumInt:
+		return &sqlspec.Column{TypeName: "int24"}, nil
```
a8m, line 326: What is `int24`?

yonidavidson: https://dev.mysql.com/doc/c-api/8.0/en/c-api-prepared-statement-type-codes.html

it's a type in mysql

sql/mysql/sqlspec_test.go
yonidavidson: You can't omit this type it's used, for example:
```mysql://rfamro@tcp(mysql-rfam-public.ebi.ac.uk:4497)/Rfam```
without supporting this type we won't be able to inspect any db using it.

rotemtam: let me clarify, the Atlas DDL has the concept of [Virtual Types](https://atlasgo.io/sql-hcl#virtual-types), it has a list of types that a user can use and Atlas will project into the db specific type. We chose not to include `uint24` in this list. Instead a user can always use the database-specific type name. In this case it would be  `mediumint`. 

The bug is actually here:
https://sourcegraph.com/github.com/ariga/atlas@master/-/blob/sql/mysql/sqlspec.go?L330-331

the type of the column should be `mediumint`.

rotemtam (approved):

PR #220, opened 2021-11-14
schema/schemaspec: support string and bool lists

rotemtam:
More work towards supporting primitive types in HCL lists. 

Next PR will include support for numeric types.

### DIFF
--- a/schema/schemaspec/extension.go
+++ b/schema/schemaspec/extension.go
@@ -358,7 +358,7 @@ func scanSliceAttr(key string, r *Resource, field reflect.Value) error {
 			lst.V = append(lst.V, &LiteralValue{V: strconv.FormatBool(item)})
 		}
 	case reflect.Ptr:
-		if typ.Kind() != reflect.Slice || typ.Elem() != reflect.TypeOf(&Ref{}) {
+		if typ.Elem() != reflect.TypeOf(&Ref{}) {
 			return fmt.Errorf("schemaspec: currently on ref slice values supported, got %s", typ)
 		}
 		for i := 0; i < field.Len(); i++ {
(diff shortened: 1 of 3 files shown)


schema/schemaspec/extension.go
```diff
@@ -320,17 +341,32 @@ func scanAttr(key string, r *Resource, field reflect.Value) error {
...
+			item := field.Index(i).Interface().(string)
+			lst.V = append(lst.V, &LiteralValue{V: strconv.Quote(item)})
+		}
+	case reflect.Bool:
+		for i := 0; i < field.Len(); i++ {
+			item := field.Index(i).Interface().(bool)
+			lst.V = append(lst.V, &LiteralValue{V: strconv.FormatBool(item)})
+		}
+	case reflect.Ptr:
+		if typ.Kind() != reflect.Slice || typ.Elem() != reflect.TypeOf(&Ref{}) {
```
yonidavidson, line 361: not a fan of this kind of logic, I prefer to break it but make it more readble.

yonidavidson: even better, error log by case.

sql/mysql/sqlspec_test.go
```diff
@@ -454,6 +454,7 @@ func TestMarshalSpecColumnType(t *testing.T) {
 			}
 			s.Tables[0].Schema = &s
 			ddl, err := MarshalSpec(&s, schemahcl.Marshal)
+			fmt.Println(string(ddl))
```
yonidavidson, line 457: why print?

schema/schemaspec/extension.go
rotemtam: good point the assertion that its a slice is not needed at all, dropping it

sql/mysql/sqlspec_test.go
rotemtam: correct!~

a8m (approved):

PR #223, opened 2021-11-14
sql/postgres: support dropping foreign keys and skip auto-dropped changes

a8m:


sql/postgres/migrate.go
```diff
@@ -421,6 +421,30 @@ func (m *migrate) fks(b *sqlx.Builder, fks ...*schema.ForeignKey) {
...
+	dropC := make(map[string]bool)
+	for _, c := range changes {
+		if c, ok := c.(*schema.DropColumn); ok {
+			dropC[c.C.Name] = true
+		}
+	}
+	for i, c := range changes {
+		switch c := c.(type) {
+		case *schema.DropIndex:
+			if len(c.I.Parts) == 1 && c.I.Parts[0].C != nil && dropC[c.I.Parts[0].C.Name] {
```
yonidavidson, line 436: can you elaborate here a bit in doc? this is not the easiest code to read

rotemtam (approved):

sql/postgres/migrate.go
a8m: Done. Thanks 🙏

PR #227, opened 2021-11-15
internal/integration: multi column key changes

a8m:


sql/postgres/migrate.go
```diff
@@ -421,34 +421,43 @@ func (m *migrate) fks(b *sqlx.Builder, fks ...*schema.ForeignKey) {
...
+// skipAutoChanges filters unnecessary changes that are automatically
+// happened by the database when ALTER TABLE is executed.
+func skipAutoChanges(changes []schema.Change) []schema.Change {
 	dropC := make(map[string]bool)
 	for _, c := range changes {
 		if c, ok := c.(*schema.DropColumn); ok {
 			dropC[c.C.Name] = true
 		}
 	}
+search:
```
yonidavidson, line 439: no possibility for an infinite loop? maybe add max counter?

yonidavidson (approved):

sql/postgres/migrate.go
a8m: `continue` is not `goto`

https://play.golang.org/p/3PzJ51D5rkZ

PR #232, opened 2021-11-16
cli/atlas: product review

yonidavidson:
- includes all changes in review (doesn't include postgress errors found today).

### DIFF
--- a/cmd/atlas/mux.go
+++ b/cmd/atlas/mux.go
@@ -71,7 +71,7 @@ func (u *Mux) OpenAtlas(dsn string) (*Driver, error) {
 	}
 	p, ok := u.providers[key]
 	if !ok {
-		return nil, fmt.Errorf("could not find provider, %s", err)
+		return nil, fmt.Errorf("could not find provider, %s", key)
 	}
 	return p(dsn)
 }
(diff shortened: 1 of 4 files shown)


cmd/atlas/mux.go
```diff
@@ -71,7 +71,7 @@ func (u *Mux) OpenAtlas(dsn string) (*Driver, error) {
 	}
 	p, ok := u.providers[key]
 	if !ok {
-		return nil, fmt.Errorf("could not find provider, %s", err)
+		return nil, fmt.Errorf("could not find provider, %s", key)
```
a8m, line 74:
```suggestion
		return nil, fmt.Errorf("could not find provider: %s", key)
```

cmd/atlas/schema.go
```diff
@@ -7,8 +7,8 @@ import (
 // schemaCmd represents the schema command
 var schemaCmd = &cobra.Command{
 	Use:   "schema",
-	Short: "Work with atlas schemas.",
-	Long:  `Provides ability to interact with schema and data source.`,
+	Short: "Work with atlas schemas",
+	Long:  `Interact with the schema and data source`,
```
a8m, line 11:
```suggestion
	Long:  "Interact with the schema and data source",
```

Why the inconsistency between the 2 strings?

a8m (approved):

cmd/atlas/schema.go
yonidavidson: short is when it's a sub command.
long is when you are -h on a specific command.

yonidavidson:
@a8m  - short command - is the text you see when you are looking at a sub command  (for example inspect under help).
lonh -is the help you see when quering a specific command for example "atlas schema inspect -h

PR #233, opened 2021-11-17
cmd/atlas/doc: generate docs

yonidavidson:
image
image


cmd/atlas/doc/doc.go
```diff
@@ -0,0 +1,26 @@
+//go:build ignore
+// +build ignore
+
+package main
```
a8m, line 4: No need for this package. Move it to `cmd/atlas`.

a8m (approved):

PR #242, opened 2021-11-23
sql/sqlite: bug fixes and migration support

a8m:


internal/integration/sqlite_test.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (t *liteTest) dropTables(names ...string) {
+	t.Cleanup(func() {
+		for i := range names {
+			_, err := t.db.Exec("DROP TABLE IF EXISTS " + names[i])
+			require.NoError(t.T, err, "drop tables %q", names[i])
+		}
+	})
+}
```
yonidavidson, line 164: Are you adding TestSqlite_Ent later?

sql/sqlite/migrate.go
```diff
@@ -0,0 +1,299 @@
...
+	if alterable(modify) {
+		return m.alterTable(ctx, modify)
+	}
+	if err := m.skipConstraints(ctx, func(ctx context.Context) error {
+		newT := *modify.T
+		indexes := newT.Indexes
+		newT.Indexes = nil
+		newT.Name = "new_" + newT.Name
+		// Create a new table with a temporary name, and copy the existing rows to it.
+		if err := m.addTable(ctx, &schema.AddTable{T: &newT}); err != nil {
```
yonidavidson, line 94: Wow!

internal/integration/sqlite_test.go
a8m: Yes. It’s WIP 🙃

a8m:
> Looks great, just missing the ent integration tests, is that for another PR?

Yes.

PR #245, opened 2021-11-24
Feat/hcl interfaces

rotemtam:
This PR adds support for reading from `schemaspec.Resource` trees into structs that contain slices of interface types. 

### Example 

Consider this block of HCL:
```hcl
zoo "ramat_gan" {
	lion "simba" {
		friend = "rafiki"
	}
	parrot "iago" {
		boss = "jafar"
	}
}
```

Both `lion` and `parrot` represent animals, and we would like to read them into a struct that contains an `[]Animal` slice such as:

```go
type (
  Animal interface {
	  animal()
  }
  Parrot struct {
	  Animal
	  Boss string `spec:"boss"`
  }
  Lion struct {
	  Animal
	  Friend string `spec:"friend"`
  }
  Zoo struct {
	  Animals []Animal `spec:""`
  }
)
```


schema/schemaspec/extension.go
```diff
@@ -94,6 +108,30 @@ func (r *Resource) As(target interface{}) error {
 				return err
 			}
 			delete(existingAttrs, attr.K)
+		case ft.isInterfaceSlice:
```
yonidavidson, line 111: why isInterfaceSlice is a struct attribute by isResourceSlice() requires a func?

yonidavidson (approved):
💯

schema/schemaspec/extension.go
```diff
@@ -53,6 +53,20 @@ func (r registry) lookup(ext interface{}) (string, bool) {
 	return "", false
 }
 
+// implementors returns a slice of the names of the extensions that implement i.
+func (r registry) implementors(i reflect.Type) ([]string, error) {
```
a8m, line 57:
```suggestion
func (r registry) implementers(i reflect.Type) ([]string, error) {
```

implementers?

schema/schemaspec/extension.go
```diff
@@ -388,24 +426,27 @@ func specFields(ext interface{}) []fieldTag {
 		}
 		parts := strings.Split(lookup, ",")
 		fields = append(fields, fieldTag{
-			field:  f.Name,
-			tag:    lookup,
-			isName: len(parts) > 1 && parts[1] == "name",
+			field:            f.Name,
+			tag:              lookup,
+			isName:           len(parts) > 1 && parts[1] == "name",
+			isInterfaceSlice: f.Type.Kind() == reflect.Slice && f.Type.Elem().Kind() == reflect.Interface,
```
a8m, line 432: What if it's not a slice? An array or just an interface?

schema/schemaspec/schemahcl/hcl_test.go
```diff
@@ -143,3 +144,49 @@ func ExampleMarshal() {
...
+			Animal
+			Boss string `spec:"boss"`
+		}
+		Lion struct {
+			Animal
+			Friend string `spec:"friend"`
+		}
+		Zoo struct {
+			Animals []Animal `spec:""`
+		}
```
a8m, line 173: Can you add a test case for something like:

```go
type U struct {
	Pet Animal `spec:""`
}
```

schema/schemaspec/extension.go
rotemtam: yep

schema/schemaspec/schemahcl/hcl_test.go
rotemtam: done

schema/schemaspec/extension.go
rotemtam: added support for just an interface. will support array in a subsequent diff

a8m (approved):
👍

PR #246, opened 2021-11-24
sql/migratespec: support modify_table and add_column

rotemtam:
Stacked on #245 

Adding first migrate spec types for initial review of the API. 

Starting to add types that capture the different change types from `sql/migrate`.

In HCL, it will look something like:

```hcl
modify_table {
	table = "users"
	add_column {
		column "id" {
			type = "int"
		}
	}
}
```

After building this layer, we can start exploring how the migration document looks like. For example:

```hcl
migration "xyz" {
  up {
     modify_table {
        // ...
     }
  }
  down {
     // ...
  }
}
```


yonidavidson (approved):
💯 
Just fix linter issues

sql/migratespec/migratespec.go
```diff
@@ -0,0 +1,26 @@
+package migratespec
```
a8m, line 1: Let's keep this code in `sqlspec`?

a8m (changes requested):
See my comment.

sql/migratespec/migratespec.go
rotemtam: done

sql/sqlspec/migratespec.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"ariga.io/atlas/schema/schemaspec"
+)
+
+type (
+	// Change is the interface implemented by change specifications. Change instances are supposed
+	// to be mappable to schema.Change instances.
+	Change interface {
+		c()
```
a8m, line 11: https://github.com/ariga/atlas/blob/master/sql/schema/migrate.go#L24
```suggestion
		change()
```

internal/integration/hclsqlspec/migrate_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"testing"
+
+	"ariga.io/atlas/schema/schemaspec/schemahcl"
+	"ariga.io/atlas/sql/sqlspec"
+	"github.com/stretchr/testify/require"
+)
+
+func Test(t *testing.T) {
```
a8m, line 11:
```suggestion
func TestMigrate(t *testing.T) {
```

a8m (approved):
Please fix the above comments.

PR #252, opened 2021-11-27
sql/postgres: support marshaling array types in sqlspec

a8m:
Note, apply still fails due to some wrong parsing logic in unmarshaling

### DIFF
--- a/sql/postgres/sqlspec.go
+++ b/sql/postgres/sqlspec.go
@@ -257,12 +257,16 @@ func columnTypeSpec(t schema.Type) (*sqlspec.Column, error) {
 		return &sqlspec.Column{TypeName: t.T}, nil
 	case *schema.UnsupportedType:
 		return &sqlspec.Column{TypeName: t.T}, nil
-	case *NetworkType:
-		return &sqlspec.Column{TypeName: t.T}, nil
-	case *CurrencyType:
+	case *ArrayType:
 		return &sqlspec.Column{TypeName: t.T}, nil
 	case *BitType:
 		return bitSpec(t)
+	case *CurrencyType:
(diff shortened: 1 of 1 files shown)


sql/postgres/sqlspec.go
```diff
@@ -257,12 +257,20 @@ func columnTypeSpec(t schema.Type) (*sqlspec.Column, error) {
...
+	case *NetworkType:
+		return &sqlspec.Column{TypeName: t.T}, nil
+	case *SerialType:
+		return &sqlspec.Column{TypeName: t.T}, nil
+	case *UUIDType:
+		return &sqlspec.Column{TypeName: t.T}, nil
+	case *XMLType:
+		return &sqlspec.Column{TypeName: t.T}, nil
 	default:
 		return nil, fmt.Errorf("failed to convert column type %T to spec", t)
```
rotemtam, line 275: Went over this with @yonidavidson on the MySQL package a little while ago, I think the correct approach is:
```suggestion
	default:
		return &sqlspec.Column{TypeName: t.T}, nil
```

rotemtam (changes requested):

sql/postgres/sqlspec.go
rotemtam: And same goes for all that types that this function is merely a pass-through for `&sqlspec.Column{TypeName: t.T}, nil`

a8m: A type assertion is required for getting `.T`.

rotemtam: got it 👍

rotemtam (approved):

PR #262, opened 2021-11-29
sql/sqlspec: add support for add_table and drop_table specs

rotemtam:


internal/integration/hclsqlspec/migrate_test.go
```diff
@@ -18,6 +18,21 @@ modify_table {
 		}
 	}
 }
+
+add_table {
+	table "products" {
+		column "sku" {
+			type = "string"
+		}
+		index "sku" {
```
yonidavidson, line 27: can you change the name of the index? better coverage.

yonidavidson (approved):

internal/integration/hclsqlspec/migrate_test.go
```diff
@@ -18,6 +18,21 @@ modify_table {
...
+			unique = true
+		}
+	}
+}
+
+drop_table {
+	table = "products"
+}
```
a8m, line 35: I wonder, if something like the following looks better:

```suggestion
drop_table "products" {}

drop_table "products" {
	// attrs.
}

modify_table "products" {
	add_column "..." {
		...
	}
}
```

rotemtam: for drop/add table I agree, changing.

for add_column, i think its nice that the user basically embeds a `*sqlspec.Column` and we can reuse that component and syntax without much magic.

yonidavidson (approved):

internal/integration/hclsqlspec/migrate_test.go
rotemtam: Looking at an option to support 

```go
type AddTable struct {
  *sqlspec.Table
}
```

currently `Resource.As` breaks on this, but I'm investigating why.

a8m: > for add_column, i think its nice that the user basically embeds a *sqlspec.Column and we can reuse that component and syntax without much magic.

Who's the user here? Also, what about `drop` and `modify`? Do you think they need to have a different structure?

```hcl
modify_table "t" {
    add_column "c1" {
        type = "int"
        ...
    }

    drop_column "c2" {}

    modify_column "c3" {
        ...
    }
}
```

PR #268, opened 2021-11-29
internal/integration: cli mysql

yonidavidson:


internal/integration/mysql_test.go
```diff
@@ -502,6 +505,34 @@ schema "test" {
...
+		myRun(t, func(t *myTest) {
+			defer os.RemoveAll("atlas")
+			cmd := exec.Command("go", "run", "ariga.io/atlas/cmd/atlas",
+				"schema",
+				"inspect",
+				"-d",
+				t.dsn(),
+			)
+			stderr := bytes.NewBuffer(nil)
+			stdout := bytes.NewBuffer(nil)
```
rotemtam, line 519:
```suggestion
			var (
			  stdout, stderr bytes.Buffer
			)
```

internal/integration/mysql_test.go
```diff
@@ -502,6 +505,34 @@ schema "test" {
 	})
 }
 
+func TestMySQL_CLI(t *testing.T) {
+	t.Run("Inspect", func(t *testing.T) {
```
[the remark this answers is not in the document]
yonidavidson: There are going to be a few cases (inspect, apply etc..)

internal/integration/mysql_test.go
```diff
@@ -502,6 +505,37 @@ schema "test" {
...
+		myRun(t, func(t *myTest) {
+			defer os.RemoveAll("atlas")
+			c := exec.Command("go", "get", "-d", "ariga.io/atlas/cmd/atlas")
```
a8m, line 512: ?

internal/integration/mysql_test.go
```diff
@@ -502,6 +505,37 @@ schema "test" {
...
+			require.NoError(t, c.Run())
+			cmd := exec.Command("go", "run", "ariga.io/atlas/cmd/atlas",
+				"schema",
+				"inspect",
+				"-d",
+				t.dsn(),
+			)
+			var (
+			  stdout, stderr bytes.Buffer
+			)
```
a8m, line 522: Grouping is not needed here.

internal/integration/mysql_test.go
```diff
@@ -26,6 +28,7 @@ type myTest struct {
 	db      *sql.DB
 	drv     *mysql.Driver
 	version string
+	port    int
```
a8m, line 31: Store and use the dsn instead?

yonidavidson: I like the port,
a. it can be used for something else,
b. it requires "calculating" the string and all the other params are just "passed" not "calculated"

internal/integration/mysql_test.go
yonidavidson: I need a way for the integrating test to get this packages on the first time, I moved it to the top of the test.

internal/integration/mysql_test.go
yonidavidson: grouped as a result of another comment.

a8m (approved):

PR #271, opened 2021-11-30
sql/sqlite: extract foreign-key constraint names from create statement

a8m:


sql/sqlite/driver.go
```diff
@@ -566,6 +566,77 @@ func autoinc(t *schema.Table) {
...
+	}
+	// Loop over inlined column constraints.
+	for _, m := range reConstC.FindAllStringSubmatch(c.S, -1) {
+		if len(m) != 5 {
+			return fmt.Errorf("unexpected number of matches for a column constraint: %q", m)
+		}
+		// Pattern matches "column", "constraint_name", "ref_table" and "ref_columns".
+		for _, fk := range t.ForeignKeys {
+			// Found a foreign-key match for the constraint.
+			if matchFK(fk, columns(m[1]), m[3], columns(m[4])) {
```
yonidavidson, line 604: I would add 
You need to document what returns from the regex, this code starts from a hard regex and then m[part] is used along with the function.

rotemtam (approved):
LGTM. would put an example matching string in the comment above each regexp

a8m:
Thanks for the review @rotemtam and @yonidavidson. I added unit tests for the regex as @yonidavidson recommended.

yonidavidson (approved):
Excellent regx testing!

PR #272, opened 2021-11-30
internal/integration: pg-cli test

yonidavidson:


internal/integration/postgres_test.go
```diff
@@ -421,6 +427,94 @@ schema "test" {
...
+			f := "atlas.hcl"
+			err := ioutil.WriteFile(f, []byte(h), 0644)
+			require.NoError(t, err)
+			defer os.Remove(f)
+			cmd := exec.Command("go", "run", "-mod=mod", "ariga.io/atlas/cmd/atlas",
+				"schema",
+				"apply",
+				"-d",
+				t.dsn(),
+				"-f",
```
rotemtam, line 494: feels like these tests are similar enough to extract to a helper fn now?

rotemtam: i would keep the assertions part of the test body, and not couple them to the helper though.

rotemtam (changes requested):

internal/integration/postgres_test.go
yonidavidson: You are right!, I merged

a8m (approved):

PR #273, opened 2021-12-01
internal/integration: merge hcl testing

yonidavidson:


internal/integration/integration_test.go
```diff
@@ -25,6 +25,7 @@ type T interface {
 	dropTables(...string)
 	migrate(...schema.Change)
 	diff(*schema.Table, *schema.Table) []schema.Change
+	applyHcl(spec string)
```
a8m, line 28:
```suggestion
	applyHcl(string)
```

a8m (approved):
💯

PR #275, opened 2021-12-03
schema/schemaspec/schemahcl: supporting injecting variables into an Atlas DDL HCL unmarshaling

rotemtam:
Allows users to inject a `map[string]interface{}` of values into the HCL document that are available as 

```hcl
x = vars.x
```

### DIFF
--- a/schema/schemaspec/schemahcl/context.go
+++ b/schema/schemaspec/schemahcl/context.go
@@ -27,9 +27,13 @@ import (
 //		title = "{greeting.hebrew.word}, world!"
 //	}
 //
-func evalCtx(f *hcl.File) (*hcl.EvalContext, error) {
+func evalCtx(f *hcl.File, cfg *unmarshalConfig) (*hcl.EvalContext, error) {
 	c := &container{}
-	if diag := gohcl.DecodeBody(f.Body, &hcl.EvalContext{}, c); diag.HasErrors() {
+	ctx, err := ctxFromVars(cfg.Variables)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


schema/schemaspec/schemahcl/hcl.go
```diff
@@ -30,24 +30,49 @@ var (
...
+	}
+}
 
 type container struct {
 	Body hcl.Body `hcl:",remain"`
 }
 
 // decode decodes the input Atlas HCL document and returns a *schemaspec.Resource representing it.
-func decode(body []byte) (*schemaspec.Resource, error) {
+func decode(body []byte, cfg *unmarshalConfig) (*schemaspec.Resource, error) {
```
yonidavidson, line 75: is cfg the right word? isn't it hclContext?

schema/schemaspec/schemahcl/context.go
```diff
@@ -27,9 +27,13 @@ import (
 //		title = "{greeting.hebrew.word}, world!"
 //	}
 //
-func evalCtx(f *hcl.File) (*hcl.EvalContext, error) {
+func evalCtx(f *hcl.File, cfg *unmarshalConfig) (*hcl.EvalContext, error) {
```
yonidavidson, line 30: is unmarshalConfig the right name? isn't it hclContext?

rotemtam:
@yonidavidson , thx for the review. This branch is old and outdated, I will re-do it after all the work we did for custom types.

PR #276, opened 2021-12-03
sql/mysql: add support for charset attributes in sqlspec

a8m:


sql/mysql/sqlspec.go
```diff
@@ -307,8 +352,10 @@ func binarySpec(t *schema.BinaryType) (*sqlspec.Column, error) {
 func stringSpec(t *schema.StringType) (*sqlspec.Column, error) {
 	switch t.T {
 	case tVarchar, tMediumText, tLongText, tTinyText, tText, tChar:
-		s := strconv.Itoa(t.Size)
-		return specutil.NewCol("", "string", specutil.LitAttr("size", s)), nil
+		if t.Size == 0 {
+			return &sqlspec.Column{Type: t.T}, nil
+		}
+		return specutil.NewCol("", "string", specutil.LitAttr("size", strconv.Itoa(t.Size))), nil
```
a8m, line 358: A tiny patch to support text fields. Will remove after we decide to tackle virtual fields.

sql/mysql/sqlspec_test.go
```diff
@@ -151,6 +151,134 @@ table "accounts" {
...
+    null      = false
+    type      = "text"
+    charset   = "utf8mb4"
+    collation = "utf8mb4_0900_ai_ci"
+  }
+}
+schema "test" {
+  charset   = "utf8mb4"
+  collation = "utf8mb4_0900_ai_ci"
+}
```
yonidavidson, line 251: Excellent test

yonidavidson (approved):
💯 
Not a fan that we need to apply charset on each logical part of the DB (schema, table, column ) but it better be explicit.

a8m:
> 💯 Not a fan that we need to apply charset on each logical part of the DB (schema, table, column ) but it better be explicit.

We don't, it's inherited from the parent element and you can override it for specific elements.

PR #281, opened 2021-12-05
internal/integration: add sanity

yonidavidson:


internal/integration/mysql_test.go
```diff
@@ -636,6 +811,20 @@ func (t *myTest) posts() *schema.Table {
...
+func (t *myTest) valueByVersion(v56 string, v57 string, v8 string) string {
+	switch t.version {
+	case "56":
+		return v56
+	case "57":
+		return v57
+	case "8":
+		return v8
+	default:
+		t.Fail()
```
a8m, line 823: Maybe Fatal/Error with a message? 

```suggestion
		t.Fatalf("unrecognized version: %q", t.version)
```

a8m (approved):

internal/integration/mysql_test.go
yonidavidson: thanks, updated

PR #283, opened 2021-12-05
internal/integration: add mariadb to mysql test cases

a8m:


internal/integration/mysql_test.go
```diff
@@ -637,7 +647,14 @@ create table atlas_types_sanity
...
 						Raw: "timestamp", Null: true},
-					Default: &schema.RawExpr{X: "CURRENT_TIMESTAMP"},
+					Default: &schema.RawExpr{
+						X: func() string {
+							if t.mariadb() {
+								return "current_timestamp()"
+							}
+							return "CURRENT_TIMESTAMP"
+						}(),
+					},
```
yonidavidson, line 657: why not t.valueByVersion(?

yonidavidson (approved):

internal/integration/mysql_test.go
a8m: because I have 3 versions, and the map feels too big.

PR #284, opened 2021-12-05
schema/schemaspec/schemahcl: adding support for describing types using functions

rotemtam:


schema/schemaspec/schemahcl/opts.go
```diff
@@ -37,10 +41,48 @@ func EvalContext(ctx *hcl.EvalContext) Option {
...
+						T: typeSpec.T,
+					}
+					for i, arg := range args {
+						a := schemaspec.Attr{K: typeSpec.Attributes[i].Name}
+						v, err := extractLiteralValue(arg)
+						if err != nil {
+							return cty.NilVal, err
+						}
+						a.V = v
+						t.Attributes = append(t.Attributes, &a)
```
a8m, line 80:
```suggestion
						v, err := extractLiteralValue(arg)
						if err != nil {
							return cty.NilVal, err
						}
						t.Attributes = append(t.Attributes, schemaspec.Attr{K: typeSpec.Attributes[i].Name, V: v})
```

It will be easier to read it this way.

a8m (approved):
💯

PR #291, opened 2021-12-07
schema/schemaspec/schemahcl: don't break on type attributes that aren't defined as type arguments in sql

rotemtam:
this is infra to support `TypeSpec`s that have attributes that aren't represented as "type arguments" (for example `unsigned` in integer types

### DIFF
--- a/schema/schemaspec/schemahcl/hcl.go
+++ b/schema/schemaspec/schemahcl/hcl.go
@@ -292,11 +292,15 @@ func (s *state) findTypeSpec(t string) (*schemaspec.TypeSpec, bool) {
 }
 
 func hclType(spec *schemaspec.TypeSpec, typ *schemaspec.Type) (string, error) {
-	if len(spec.Attributes) == 0 {
+	if len(typeFuncArgs(spec)) == 0 {
 		return typ.T, nil
 	}
 	args := make([]string, 0, len(spec.Attributes))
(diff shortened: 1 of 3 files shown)


schema/schemaspec/schemahcl/hcl.go
```diff
@@ -306,6 +310,15 @@ func hclType(spec *schemaspec.TypeSpec, typ *schemaspec.Type) (string, error) {
...
 }
 
+func findAttr(attrs []*schemaspec.Attr, k string) (*schemaspec.Attr, bool) {
+	for _, atr := range attrs {
+		if atr.K == k {
+			return atr, true
+		}
+	}
+	return nil, false
+}
```
a8m, line 320:
```suggestion
func hasAttr(attrs []*schemaspec.Attr, k string) (*schemaspec.Attr, bool) {
	for _, attr := range attrs {
		if attr.K == k {
			return attr, true
		}
	}
	return nil, false
}
```

Let's stay consistent with other parts.

a8m (approved):

schema/schemaspec/schemahcl/hcl.go
rotemtam: I think it's odd for a function that returns the value to be called `hasX` which usually only returns a bool. Don't you think?

PR #298, opened 2021-12-11
sql/internal/specutil: adding TypeRegistry for TypeSpecs

rotemtam:
Stacked on #297 

Adding a new container type for working with TypeSpecs. This is added to make the needed changes to mysql/postgres for type-safe type definitions in Atlas DDL.

### DIFF
--- a/schema/schemaspec/spec.go
+++ b/schema/schemaspec/spec.go
@@ -67,8 +67,8 @@ type (
 
 	// TypeSpec represents a specification for defining a Type.
 	TypeSpec struct {
-		Name       string
-		T          string
+		Name       string // Name is the identifier for the type in an Atlas DDL document.
+		T          string // T is the database identifier for the type.
 		Attributes []*TypeAttr
 	}
 
--- a/schema/schemaspec/spec.go
+++ b/schema/schemaspec/spec.go
@@ -67,8 +67,10 @@ type (
 
 	// TypeSpec represents a specification for defining a Type.
 	TypeSpec struct {
-		Name       string // Name is the identifier for the type in an Atlas DDL document.
-		T          string // T is the database identifier for the type.
+		// Name is the identifier for the type in an Atlas DDL document.
+		Name       string
+		 // T is the database identifier for the type.
+		T          string
 		Attributes []*TypeAttr
 	}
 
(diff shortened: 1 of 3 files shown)


schema/schemaspec/spec.go
```diff
@@ -67,8 +67,8 @@ type (
 
 	// TypeSpec represents a specification for defining a Type.
 	TypeSpec struct {
-		Name       string
-		T          string
+		Name       string // Name is the identifier for the type in an Atlas DDL document.
+		T          string // T is the database identifier for the type.
```
a8m, line 71:
```suggestion
		// Name is the identifier for the type in an Atlas DDL document.
		Name       string
		 // T is the database identifier for the type.
		T          string
```
Comments are too long IMO for inline.

sql/internal/specutil/types.go
```diff
@@ -40,3 +41,42 @@ func PrintType(typ *schemaspec.Type, spec *schemaspec.TypeSpec) (string, error)
...
+type TypeRegistry struct {
+	r []*schemaspec.TypeSpec
+}
+
+// Register adds one or more TypeSpec to the registry.
+func (r *TypeRegistry) Register(specs ...*schemaspec.TypeSpec) error {
+	for _, s := range specs {
+		if _, exists := r.Find(s.T); exists {
+			return fmt.Errorf("specutil: type with T of %q already registered", s.T)
+		}
```
a8m, line 55: If the Name is the Atlas identifier, why checking also the type?

rotemtam: I have some concerns on how to handle multiple identifiers that map to the same db type. Namely, how to disambiguate between them. Because I don't have a solution there yet, I'm defensive and enforcing the variant that currently only one identifier  can represent `T`.

a8m: Why do we care about the database type in the DDL part? Why not allow registering both "int" and "integer" on MySQL, even though both of them map to `int? See https://dev.mysql.com/doc/refman/8.0/en/integer-types.html

> MySQL supports the SQL standard integer types INTEGER (or INT) ...

rotemtam: b/c you need to be able to go back and forth from the db to hcl without losing info
if a user used `int` you want it to marshal/unmarshal/inspect the same.

a8m: If you use `integer` in the DDL (atlas or create stmt), you'll get back an `int` type on inspection.

a8m (approved):

PR #299, opened 2021-12-11
Feat/convert schema type

rotemtam:
Stacked on #298 

Adding method to convert schema.Type the a schemaspec.Type generically, using reflection. 
In the future, should probably be solved by expanding the schema.Type interface.

### DIFF
--- a/sql/internal/specutil/types.go
+++ b/sql/internal/specutil/types.go
@@ -2,6 +2,7 @@ package specutil
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 
 	"ariga.io/atlas/schema/schemaspec"
@@ -39,4 +40,43 @@ func PrintType(typ *schemaspec.Type, spec *schemaspec.TypeSpec) (string, error) 
 		mid = "(" + strings.Join(args, ",") + ")"
 	}
 	return typ.T + mid + suffix, nil
+}
+
+// TypeRegistry is a collection of *schemaspec.TypeSpec.
+type TypeRegistry struct {
+	r []*schemaspec.TypeSpec
+}
+
+// Register adds one or more TypeSpec to the registry.
+func (r *TypeRegistry) Register(specs ...*schemaspec.TypeSpec) error {
+	for _, s := range specs {
+		if _, exists := r.Find(s.T); exists {
+			return fmt.Errorf("specutil: type with T of %q already registered", s.T)
+		}
+		if _, exists := r.FindByName(s.Name); exists {
+			return fmt.Errorf("specutil: type with name of %q already registered", s.T)
+		}
+	}
+	r.r = append(r.r, specs...)
+	return nil
+}
+
+// FindByName searches the registry for types that have the provided name.
+func (r *TypeRegistry) FindByName(name string) (*schemaspec.TypeSpec, bool) {
+	for _, current := range r.r {
+		if current.Name == name {
+			return current, true
+		}
+	}
+	return nil, false
+}
+
+// Find searches the registry for types that have the provided T.
+func (r *TypeRegistry) Find(t string) (*schemaspec.TypeSpec, bool) {
+	for _, current := range r.r {
+		if current.T == t {
+			return current, true
+		}
+	}
+	return nil, false
 }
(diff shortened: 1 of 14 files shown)


sql/internal/specutil/types.go
```diff
@@ -80,3 +84,81 @@ func (r *TypeRegistry) Find(t string) (*schemaspec.TypeSpec, bool) {
...
+		}
+		switch tat.Kind {
+		case reflect.Int:
+			i := strconv.Itoa(int(field.Int()))
+			s.Attributes = append(s.Attributes, LitAttr(tat.Name, i))
+		case reflect.Bool:
+			b := strconv.FormatBool(field.Bool())
+			s.Attributes = append(s.Attributes, LitAttr(tat.Name, b))
+		case reflect.Slice:
+			var lits []string
```
a8m, line 124: make

sql/internal/specutil/types.go
```diff
@@ -80,3 +84,81 @@ func (r *TypeRegistry) Find(t string) (*schemaspec.TypeSpec, bool) {
...
+	}
+	if tf.Kind() != reflect.String {
+		return nil, fmt.Errorf("specutil: cannot convert non-string T field for %T", typ)
+	}
+	s.T = tf.String()
+	typeSpec, ok := r.Find(s.T)
+	if !ok {
+		return nil, fmt.Errorf("specutil: type %q not found in registry", s.T)
+	}
+	for _, tat := range typeSpec.Attributes {
```
a8m, line 107:
```suggestion
	for _, attr := range typeSpec.Attributes {
```

sql/internal/specutil/types.go
```diff
@@ -80,3 +84,81 @@ func (r *TypeRegistry) Find(t string) (*schemaspec.TypeSpec, bool) {
...
+		n := inflect.Camelize(tat.Name)
+		field := vof.FieldByName(n)
+		if !field.IsValid() {
+			return nil, fmt.Errorf("invalid field name %q for attr %q on type spec %q", n, tat.Name, typeSpec.T)
+		}
+		if field.Kind() != tat.Kind {
+			return nil, errors.New("incompatible kinds on typespec attr and typefield ")
```
a8m, line 114:
```suggestion
			return nil, errors.New("incompatible kinds on typespec attr and typefield")
```

sql/internal/specutil/types.go
```diff
@@ -80,3 +84,81 @@ func (r *TypeRegistry) Find(t string) (*schemaspec.TypeSpec, bool) {
 	}
 	return nil, false
 }
+
+// Convert converts the schema.Type to a *schemaspec.Type.
+func (r *TypeRegistry) Convert(typ schema.Type) (*schemaspec.Type, error) {
+	s := &schemaspec.Type{}
+	vof := reflect.ValueOf(typ)
```
a8m, line 91:
```suggestion
	rv := reflect.ValueOf(typ)
```

rv (reflect value) is more common.

sql/internal/specutil/types.go
```diff
@@ -80,3 +84,81 @@ func (r *TypeRegistry) Find(t string) (*schemaspec.TypeSpec, bool) {
...
+	if vof.Kind() == reflect.Ptr {
```
a8m, line 92: Since it's an interface, you may want to check if it's valid/not-nil.

sql/internal/specutil/types.go
```diff
@@ -80,3 +84,81 @@ func (r *TypeRegistry) Find(t string) (*schemaspec.TypeSpec, bool) {
...
+	}
+}
+
+// UnsignedTypeAttr returns a TypeAttr for an `unsigned` attribute relevant for integer types.
+func UnsignedTypeAttr() *schemaspec.TypeAttr {
+	return &schemaspec.TypeAttr{
+		Name: "unsigned",
+		Kind: reflect.Bool,
+	}
+}
```
a8m, line 164: why it's here and not in sql/mysql?

a8m (approved):
looks good. minor comments

sql/internal/specutil/types.go
rotemtam: Wanted to use it in tests, and also didn't want to commit to the API yet, so put it here under internal.

rotemtam: its actually for a test, moved it to the test file.
maybe we'll add it to the mysql package in the next diff.

PR #301, opened 2021-12-12
sql: add support for creating and dropping schemas in migrate

a8m:


sql/mysql/migrate.go
```diff
@@ -47,6 +51,33 @@ func (m *migrate) Exec(ctx context.Context, changes []schema.Change) error {
 	return nil
 }
 
+// topLevel executes first the changes for creating or dropping schemas (top-level schema elements).
+func (m *migrate) topLevel(ctx context.Context, changes []schema.Change) ([]schema.Change, error) {
```
a8m, line 55: I thought about moving it out to `sqlx`, but meh. too many abstractions for saving 10 lines of code.

image

image

rotemtam (approved):

PR #304, opened 2021-12-13
sql/mysql: add type registry

rotemtam:
This adds the TypeRegistry for the mysql driver. It does not yet replace the older string-based processing for DDL marshal/unmarshal.

### DIFF
--- a/schema/schemaspec/spec.go
+++ b/schema/schemaspec/spec.go
@@ -67,15 +67,18 @@ type (
 
 	// TypeSpec represents a specification for defining a Type.
 	TypeSpec struct {
-		Name       string // Name is the identifier for the type in an Atlas DDL document.
-		T          string // T is the database identifier for the type.
+		// Name is the identifier for the type in an Atlas DDL document.
+		Name string
+		// T is the database identifier for the type.
+		T          string
 		Attributes []*TypeAttr
 	}
 
 	// TypeAttr describes an attribute of a TypeSpec, for example `varchar` fields
 	// can have a `size` attribute.
 	TypeAttr struct {
-		Name     string // Name should be a snake_case of related the schema.Type struct field.
+		// Name should be a snake_case of related the schema.Type struct field.
+		Name     string
 		Kind     reflect.Kind
 		Required bool
 	}
--- a/schema/schemaspec/spec.go
+++ b/schema/schemaspec/spec.go
@@ -67,8 +67,8 @@ type (
 
 	// TypeSpec represents a specification for defining a Type.
 	TypeSpec struct {
-		Name       string
-		T          string
+		Name       string // Name is the identifier for the type in an Atlas DDL document.
+		T          string // T is the database identifier for the type.
 		Attributes []*TypeAttr
 	}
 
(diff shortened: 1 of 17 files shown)


schema/schemaspec/spec.go
```diff
@@ -68,16 +68,17 @@ type (
 	// TypeSpec represents a specification for defining a Type.
 	TypeSpec struct {
 		// Name is the identifier for the type in an Atlas DDL document.
-		Name       string
-		 // T is the database identifier for the type.
+		Name string
+		// T is the database identifier for the type.
 		T          string
 		Attributes []*TypeAttr
```
a8m, line 74:
```suggestion
		Attrs []*TypeAttr
```

We use Attrs in all other places in atlas.

sql/internal/specutil/types.go
```diff
@@ -38,11 +39,25 @@ func (r *TypeRegistry) PrintType(typ *schemaspec.Type) (string, error) {
...
-		if !ok {
-			return "", errors.New("expecting literal value")
+		switch v := arg.V.(type) {
+		case *schemaspec.LiteralValue:
+			args = append(args, v.V)
+		case *schemaspec.ListValue:
+			for _, li := range v.V {
+				lit, ok := li.(*schemaspec.LiteralValue)
+				if !ok {
+					return "", errors.New("expecting literal value")
```
a8m, line 49:
```suggestion
					return "", fmt.Errorf("expecting literal value. got: %T", li)
```

sql/internal/specutil/types.go
```diff
@@ -38,11 +39,25 @@ func (r *TypeRegistry) PrintType(typ *schemaspec.Type) (string, error) {
...
+				}
+				uq, err := strconv.Unquote(lit.V)
+				if err != nil {
+					return "", fmt.Errorf("expecting list items to be quoted strings: %w", err)
+				}
+
```
a8m, line 55:
```suggestion
```

sql/internal/specutil/types.go
```diff
@@ -38,11 +39,25 @@ func (r *TypeRegistry) PrintType(typ *schemaspec.Type) (string, error) {
...
+				args = append(args, `'`+uq+`'`)
```
a8m, line 56:
```suggestion
				args = append(args, "'"+uq+"'")
```

Why backticks? 

I would leave it for the driver btw. Because there are cases where these strings need to be escaped ("a'b"), and replacing `"` with `'` is not always necessary.

sql/mysql/sqlspec_test.go
```diff
@@ -5,14 +5,20 @@ import (
...
 	"ariga.io/atlas/sql/schema"
 	"ariga.io/atlas/sql/sqlspec"
+	"github.com/hashicorp/hcl/v2/hclwrite"
 
 	"github.com/stretchr/testify/require"
 )
 
+var (
+	hclState = schemahcl.New(schemahcl.WithTypes(TypeRegistry.Specs()))
+)
```
a8m, line 20:
```suggestion
var hclState = schemahcl.New(schemahcl.WithTypes(TypeRegistry.Specs()))
```

a8m (approved):

PR #308, opened 2021-12-14
schema/schemaspec/schemahcl: support optional type attrs

rotemtam:
for types with non required attributes, support both:
```hcl
type = int
// AND
type = int(10)
```

### DIFF
--- a/schema/schemaspec/schemahcl/hcl.go
+++ b/schema/schemaspec/schemahcl/hcl.go
@@ -314,7 +314,7 @@ func hclType(spec *schemaspec.TypeSpec, typ *schemaspec.Type) (string, error) {
 			}
 		}
 	}
-	// if no args were chosen and the type can be described without a function
+	// If no args were chosen and the type can be described without a function.
 	if len(args) == 0 && len(typeFuncReqArgs(spec)) == 0 {
 		return spec.Name, nil
 	}
--- a/schema/schemaspec/schemahcl/hcl.go
+++ b/schema/schemaspec/schemahcl/hcl.go
@@ -314,6 +314,10 @@ func hclType(spec *schemaspec.TypeSpec, typ *schemaspec.Type) (string, error) {
 			}
 		}
 	}
+	// if no args were chosen and the type can be described without a function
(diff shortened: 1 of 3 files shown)


schema/schemaspec/schemahcl/hcl.go
```diff
@@ -314,6 +314,10 @@ func hclType(spec *schemaspec.TypeSpec, typ *schemaspec.Type) (string, error) {
 			}
 		}
 	}
+	// if no args were chosen and the type can be described without a function
```
a8m, line 317:
```suggestion
	// If no args were chosen and the type can be described without a function.
```

schema/schemaspec/schemahcl/opts.go
```diff
@@ -47,62 +47,75 @@ func WithTypes(typeSpecs []*schemaspec.TypeSpec) Option {
 		for _, ts := range typeSpecs {
 			typeSpec := ts
 			config.types = append(config.types, typeSpec)
-			if len(typeFuncArgs(typeSpec)) == 0 {
+			// if no required args exist, register the type as a variable in the HCL context.
```
a8m, line 50:
```suggestion
			// If no required args exist, register the type as a variable in the HCL context.
```

schema/schemaspec/schemahcl/opts.go
```diff
@@ -47,62 +47,75 @@ func WithTypes(typeSpecs []*schemaspec.TypeSpec) Option {
...
+			if len(typeFuncReqArgs(typeSpec)) == 0 {
 				typ := &schemaspec.Type{T: typeSpec.T}
 				config.ctx.Variables[typeSpec.Name] = cty.CapsuleVal(ctyTypeSpec, typ)
-				continue
 			}
-			spec := &function.Spec{
-				Type: function.StaticReturnType(ctyTypeSpec),
+			// if func args exist, register the type as a function in HCL.
```
a8m, line 55:
```suggestion
			// If func args exist, register the type as a function in HCL.
```

a8m (approved):

PR #309, opened 2021-12-14
sql/mysql: aligning type registry with sanity test

rotemtam:


a8m (approved):
Looks good, but CI is red

PR #314, opened 2021-12-15
sql/internal/specutil: support omitting right-most zero-value optional param

rotemtam:
this fixes a bug where a typespec has an optional (non-required) type attribute, and the resulting DDL representation was too verbose. For instance, in some dialects both `varchar` and `varchar(20)` are valid. In these cases the `size` attribute is optional. In these cases if the size was not specified it would be rendered as `varchar(0)`. 

This PR adds a fix where the right-most optional args that have the zero value are omitted. so default:
`varchar(0)` => `varchar`, `x(10,0)` => `x(10)`, `x(0,10)` => remains `x(0,10)`

### DIFF
--- a/sql/internal/specutil/types_test.go
+++ b/sql/internal/specutil/types_test.go
@@ -109,12 +109,30 @@ func TestRegistryConvert(t *testing.T) {
 			typ:      &schema.IntegerType{T: "int", Unsigned: true},
 			expected: &schemaspec.Type{T: "int", Attrs: []*schemaspec.Attr{LitAttr("unsigned", "true")}},
 		},
+
 		{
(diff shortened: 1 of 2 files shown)


sql/internal/specutil/types_test.go
```diff
@@ -109,13 +109,31 @@ func TestRegistryConvert(t *testing.T) {
 			typ:      &schema.IntegerType{T: "int", Unsigned: true},
 			expected: &schemaspec.Type{T: "int", Attrs: []*schemaspec.Attr{LitAttr("unsigned", "true")}},
 		},
+
```
a8m, line 112: rm

a8m (approved):
Fix lint errors

PR #315, opened 2021-12-18
sql/internal/specutil: sanity tests for all drivers

rotemtam:
Adding a sanity test for all driver type registries and fixing some bugs along the way. 

postgres enum needs some more love, but ill do that separately.

### DIFF
--- a/sql/internal/spectest/spectest.go
+++ b/sql/internal/spectest/spectest.go
@@ -12,17 +12,12 @@ import (
 
 // RegistrySanityTest runs a sanity for a TypeRegistry, generated a dummy *schemaspec.Type
 // then converting it to a schema.Type and back to a *schemaspec.Type.
-func RegistrySanityTest(t *testing.T,
-	registry *specutil.TypeRegistry,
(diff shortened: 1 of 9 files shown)


sql/internal/spectest/spectest.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func contains(needle string, haystack []string) bool {
+	for _, item := range haystack {
+		if item == needle {
+			return true
+		}
+	}
+	return false
+}
```
a8m, line 41: What are these var names?
```suggestion
func contains(s string, l []string) bool {
	for i := range l {
		if s == l[i] {
			return true
		}
	}
	return false
}
```

sql/internal/spectest/spectest.go
```diff
@@ -0,0 +1,62 @@
+package spectest
```
a8m, line 1: Add atlas license header? I'll add a CI test for it later.

a8m (approved):
looks good. change please the naming

yonidavidson (approved):

PR #317, opened 2021-12-19
sql/internal/specutil: disallow required arg after optional

rotemtam:


sql/internal/specutil/types.go
```diff
@@ -83,6 +86,17 @@ func (r *TypeRegistry) Register(specs ...*schemaspec.TypeSpec) error {
 	return nil
 }
 
+func validSpec(typeSpec *schemaspec.TypeSpec) error {
+	var seenOptional bool
```
a8m, line 90:
```suggestion
	var optionalSeen bool
```

you may know what is more correct in terms of grammar.

a8m (approved):

sql/internal/specutil/types.go
rotemtam: i think both are ok. 
i meant it as "I've already seen the optional" and the way you put it is "an optional param was seen" 
leaving as is

PR #320, opened 2021-12-19
schema/schemaspec/schemahcl: support partial type arguments

rotemtam:
bugfix: for types that have "optional" parameters, such as float on MySQL - ie in the DDL both float, float(10), and float(10,2) are legit, only `float` and `float(10,2)` were supported. this PR adds support for the partial invocation `float(10)`, to mimick the db DDL.

### DIFF
--- a/schema/schemaspec/schemahcl/context_test.go
+++ b/schema/schemaspec/schemahcl/context_test.go
@@ -6,6 +6,7 @@ import (
 
 	"ariga.io/atlas/schema/schemaspec"
 	"ariga.io/atlas/schema/schemaspec/internal/schemautil"
+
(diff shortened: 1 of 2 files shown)


schema/schemaspec/schemahcl/context_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"ariga.io/atlas/schema/schemaspec"
+	"ariga.io/atlas/schema/schemaspec/internal/schemautil"
```
a8m, line 8: A line break between internal imports and external.

```suggestion
	"ariga.io/atlas/schema/schemaspec/internal/schemautil"

```

a8m (approved):
💯

PR #321, opened 2021-12-20
sql/postgres: implement the migrate.planner interface

a8m:


sql/postgres/migrate_test.go
```diff
@@ -0,0 +1,152 @@
...
+				Changes: []*migrate.Change{
+					{Cmd: `ALTER TABLE "users" DROP COLUMN "name"`},
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		db, m, err := sqlmock.New()
+		require.NoError(t, err)
+		mock{m}.version("130000")
```
rotemtam, line 140: whats 130000?

sql/postgres/migrate.go
```diff
@@ -9,32 +9,81 @@ import (
...
+		if c.Reverse == "" {
+			s.Reversible = false
+		}
+	}
+	return &s.Plan, nil
+}
+
+// ApplyChanges applies the changes on the database. An error is returned
+// if the driver is unable to produce a plan to it, or one of the statements
+// is failed or unsupported.
```
rotemtam, line 44:
```suggestion
// if the driver is unable to produce a plan to do so, or one of the statements
// is failed or unsupported.
```

rotemtam (approved):

sql/postgres/migrate_test.go
a8m: PostgreSQL version format.

PR #324, opened 2021-12-20
all: remove schema.Execer

a8m:
Replaced with migrate.ApplyChanges

### DIFF
--- a/sql/migrate/migrate.go
+++ b/sql/migrate/migrate.go
@@ -75,7 +75,12 @@ type (
 	// PlanApplier wraps the methods for planning and applying changes
 	// on the database.
 	PlanApplier interface {
+		// PlanChanges returns a migration plan for applying the given changeset.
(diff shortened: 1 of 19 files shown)


sql/migrate/migrate.go
```diff
@@ -75,7 +75,12 @@ type (
 	// PlanApplier wraps the methods for planning and applying changes
 	// on the database.
 	PlanApplier interface {
+		// PlanChanges returns a migration plan for applying the given changeset.
 		PlanChanges(context.Context, string, []schema.Change) (*Plan, error)
+
+		// ApplyChanges responsible for applying the given changeset.
+		// An error may return from ApplyChanges if the driver does not
+		// know how to execute a change.
```
rotemtam, line 83: i think less "personification" of components is better :-)
```suggestion
		// An error may return from ApplyChanges if the driver is
		// unable to execute a change.
```

rotemtam (approved):
💯

PR #329, opened 2021-12-21
sql/internal/specutil: configure type registry with func opts

rotemtam:
some refac in prep  for the `raw("int[]")` support i'm working on


a8m (approved):
looks good, just fix the lint

PR #332, opened 2021-12-21
sql/internal/specutil: handle unsupported types

rotemtam:


schema/schemaspec/schemahcl/opts.go
```diff
@@ -64,6 +65,20 @@ func WithTypes(typeSpecs []*schemaspec.TypeSpec) Option {
...
+		Params: []function.Parameter{
+			{Name: "def", Type: cty.String, AllowNull: false},
+		},
+		Type: function.StaticReturnType(ctyTypeSpec),
+		Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+			t := &schemaspec.Type{T: args[0].AsString()}
+			return cty.CapsuleVal(ctyTypeSpec, t), nil
+		},
+	}
+	return function.New(s)
```
a8m, line 79:
```suggestion
	return function.New(&function.Spec{
		Params: []function.Parameter{
			{Name: "def", Type: cty.String, AllowNull: false},
		},
		Type: function.StaticReturnType(ctyTypeSpec),
		Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
			t := &schemaspec.Type{T: args[0].AsString()}
			return cty.CapsuleVal(ctyTypeSpec, t), nil
		},
	})
```

Consider avoiding this temporary variable.

a8m (approved):

PR #334, opened 2021-12-21
cmd/atlas: use migrate plan in apply cmd

a8m:


yonidavidson (approved):

cmd/action/diff.go
```diff
@@ -1,69 +0,0 @@
-package action
```
rotemtam, line 1: 😍

rotemtam (approved):

PR #336, opened 2021-12-22
sql/internal/specutil: sharing from column to columnspec code

rotemtam:
found a bug with handling column default. small PR before so all drivers share the fix

### DIFF
--- a/sql/internal/specutil/convert.go
+++ b/sql/internal/specutil/convert.go
@@ -22,6 +22,7 @@ type (
 	ConvertPrimaryKeyFunc func(*sqlspec.PrimaryKey, *schema.Table) (*schema.Index, error)
 	ConvertIndexFunc      func(*sqlspec.Index, *schema.Table) (*schema.Index, error)
 	ColumnSpecFunc        func(*schema.Column, *schema.Table) (*sqlspec.Column, error)
+	ColumnTypeSpecFunc    func(t schema.Type) (*sqlspec.Column, error)
 	TableSpecFunc         func(*schema.Table) (*sqlspec.Table, error)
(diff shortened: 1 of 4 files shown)


sql/internal/specutil/convert.go
```diff
@@ -22,6 +22,7 @@ type (
 	ConvertPrimaryKeyFunc func(*sqlspec.PrimaryKey, *schema.Table) (*schema.Index, error)
 	ConvertIndexFunc      func(*sqlspec.Index, *schema.Table) (*schema.Index, error)
 	ColumnSpecFunc        func(*schema.Column, *schema.Table) (*sqlspec.Column, error)
+	ColumnTypeSpecFunc    func(t schema.Type) (*sqlspec.Column, error)
```
a8m, line 25:
```suggestion
	ColumnTypeSpecFunc    func(schema.Type) (*sqlspec.Column, error)
```

yonidavidson (approved):

sql/internal/specutil/convert.go
```diff
@@ -279,6 +280,22 @@ func FromPrimaryKey(s *schema.Index) (*sqlspec.PrimaryKey, error) {
 	}, nil
 }
 
+// FromColumn converts a *schema.Column into a *sqlspec.Column using the ColumnTypeSpecFunc.
+func FromColumn(col *schema.Column, columnTypeSpec ColumnTypeSpecFunc) (*sqlspec.Column, error) {
```
a8m, line 284:
```suggestion
func FromColumn(col *schema.Column, convertType ColumnTypeSpecFunc) (*sqlspec.Column, error) {
```

The rest of the functions name it as `convert<T>`.

rotemtam: actually we have <X>Spec for the ones going from schema to spec and convert<X> for other direction

PR #339, opened 2021-12-23
cmd/version: add version command

yonidavidson:


cmd/action/internal/build/build.go
```diff
@@ -0,0 +1,5 @@
+package build
+
+// Version is the atlas CLI build version
+// Should be set by build script "-X 'ariga.io/atlas/cmd/action/internal/build.Version=${version}'"
+var Version string
```
a8m, line 5: Why do we need a separate package for this variable? i.e. why not hold it in the version.go file?

cmd/action/version.go
```diff
@@ -0,0 +1,30 @@
+package action
+
+import (
+	"fmt"
+
+	"ariga.io/atlas/cmd/action/internal/build"
```
a8m, line 6:
```suggestion
	"ariga.io/atlas/cmd/action/internal/build"

```

a8m (approved):

PR #340, opened 2021-12-23
sql/mysql: support constant literals in default values

a8m:


sql/mysql/migrate.go
```diff
@@ -424,6 +416,30 @@ func (*state) attr(b *sqlx.Builder, attrs ...schema.Attr) {
...
+		if !hasNumericDefault(c.Type.Type) && !isHex(v) {
+			v = quote(v)
+		}
+		b.P("DEFAULT", v)
+	case *schema.RawExpr:
+		v := x.X
+		// For backwards compatibility, quote raw expressions that are not wrapped
+		// with parens for non-numeric column types (i.e. literals).
+		switch t := c.Type.Type; {
+		case isHex(v), hasNumericDefault(t), strings.HasPrefix(v, "(") && strings.HasSuffix(v, ")"):
```
yonidavidson, line 433: is this just for future use?

yonidavidson (approved):
LGTM, 1 comment

rotemtam (approved):

sql/mysql/migrate.go
a8m: No, it is to avoid adding their negation in the if below. It will make it too long.

PR #341, opened 2021-12-23
cmd/action: update version

yonidavidson:


cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,152 @@
...
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/mitchellh/go-homedir"
+	"github.com/pelletier/go-toml/v2"
+	"golang.org/x/mod/semver"
+)
+
+// Update logic:
```
rotemtam, line 17: 1. Fix the documentation text (more than 24h, hcl)
2. Put this is a Godoc comment on the relevant funcs

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,152 @@
...
+// 1. Check release file ~/.local/atlas/release.hcl for latest known release and poll time
+// 2. If last poll was less than 24h, poll GitHub public API https://docs.github.com/en/rest/reference/releases#get-the-latest-release
+// 3. Store the latest release metadata
+// 4. If current build Version, that is not development, is lower than the latest release, notify user
+
+type (
+	// Store is where the latest release data is stored on CLI host
+	Store struct {
```
rotemtam, line 25: Name is ambiguous and doesn't mean anything obvious to me , something else ?

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,152 @@
...
+	// LatestRelease implements the required fields from https://api.github.com/repos/ariga/atlas/releases/latest
+	LatestRelease struct {
+		Version string `json:"tag_name"`
+		URL     string `json:"html_url"`
+	}
+)
+
+// CheckForUpdate implements a notification to the user when a later release is available
+func CheckForUpdate(version string, logF func(i ...interface{})) {
+	if ok := enabled(version); !ok {
```
rotemtam, line 40: Assignment is redundant

cmd/action/internal/update/update_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+func TestCheckForUpdate(t *testing.T) {
+	url := "https://github.com/ariga/atlas/releases/tag/test"
+	tests := []struct {
+		name           string
+		currentVersion string
+		latestVersion  string
+		store          *LatestRelease
+		checkedAt      time.Time
+		latestReleaseF func() (LatestRelease, error)
```
rotemtam, line 19: Any reason this is a func? Use the value for defining the test and of you need instantiate a func in the test run body

cmd/action/internal/update/update_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestCheckForUpdate(t *testing.T) {
+	url := "https://github.com/ariga/atlas/releases/tag/test"
+	tests := []struct {
+		name           string
+		currentVersion string
+		latestVersion  string
+		store          *LatestRelease
```
rotemtam, line 17: Unit tests should be  testing the public api, put the state file text instead of the results of its parsing

cmd/action/internal/update/update.go
yonidavidson: I like the name, store as tokenStore, configStore, and part of update package is update.Store

cmd/action/internal/update/update_test.go
yonidavidson: changed that store is saved externally.

cmd/action/internal/update/update_test.go
yonidavidson: changed

cmd/action/internal/update/update.go
```diff
@@ -1,123 +1,150 @@
...
-// 2. Poll GitHub public CLI for latest release if last poll was less than 24h
-// 3. Store a release file in  with the data
-// 4. If current build version, that is not development, is lower than the latest release, notify user
-func CheckForUpdate(version string) {
-	if _, ok := os.LookupEnv("ATLAS_NO_UPDATE_NOTIFIER"); ok {
+// 1. Check release file ~/.local/atlas/release.toml for latest known release and poll time
+// 2. If last poll was more than 24h, poll GitHub public API https://docs.github.com/en/rest/reference/releases#get-the-latest-release
+// 3. Store the latest release metadata
+// 4. If current build Version, that is not development, is lower than the latest release, notify user
+func CheckForUpdate(version string, logF func(i ...interface{})) {
```
rotemtam, line 37: this signature is a bit unusual for a public method. method should return an err and its caller should decide what to do with the result. Also, this implementation couples the location of the config file, which IMO should be injected to the method. WDYT? 
in addition, the godoc is usually written as a fluent sentence and begins with the method name.
```go
// CheckForUpdate checks if a new version of Atlas CLI exists. To reduce the frequency in which
// this check is made.. yadda yadda.
```

yonidavidson: For now, I prefer to keep the logic in the package and not expose it outside to the caller, this might change in the near future on other integrations

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,150 @@
...
+	if !enabled(version) {
+		return
+	}
+	p, err := homedir.Expand("~/.config/atlas")
+	if err != nil {
+		return
+	}
+	ok, message, err := shouldUpdate(version, p, latestReleaseFromGithub)
```
a8m, line 45:
```suggestion
	ok, message, err := shouldUpdate(version, p, latestReleaseFromGithub)
```

Why not call `latestReleaseFromGithub` directly?

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,150 @@
...
+		return false
+	}
+	return true
+}
+
+func shouldUpdate(version string, path string, latestReleaseF func() (LatestRelease, error)) (bool, string, error) {
+	r, _ := localStore(path)
+	if shouldSkip(r) {
+		return false, "", nil
+	}
```
a8m, line 69: Why are you ignoring the error here?

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,150 @@
...
+		return false, "", err
+	}
+	if semver.Compare(version, l.Version) != -1 {
+		return false, "", nil
+	}
+	return true, fmt.Sprintf("new atlas version %s, %s", l.Version, l.URL), nil
+}
+
+func latestReleaseFromGithub() (LatestRelease, error) {
+	req, err := http.NewRequest("GET", "https://api.github.com/repos/ariga/atlas/releases/latest", nil)
```
a8m, line 84:
```suggestion
	req, err := http.NewRequest(http.MethodGet, "https://api.github.com/repos/ariga/atlas/releases/latest", nil)
```

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return LatestRelease{}, err
+	}
+	defer resp.Body.Close()
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return LatestRelease{}, err
+	}
+	ok := resp.StatusCode >= 200 && resp.StatusCode < 300
+	if !ok {
```
a8m, line 103:
```suggestion
	if resp.StatusCode != http.StatusOK {
```

Are there any possible status codes?

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,150 @@
...
+		return LatestRelease{}, fmt.Errorf("update: failed to fetch latest release version")
+	}
+	var b LatestRelease
+	err = json.Unmarshal(data, &b)
+	if err != nil {
+		return LatestRelease{}, err
+	}
```
a8m, line 110:
```suggestion
	if err := json.Unmarshal(data, &b); err != nil {
		return LatestRelease{}, err
	}
```

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return err
+	}
+	err = ioutil.WriteFile(fileLocation(path), b, 0600)
+	return err
+}
+
+func fileLocation(p string) string {
+	return filepath.Join(p, "release.toml")
+}
```
a8m, line 150: Why not inline this function?

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,150 @@
...
+	err = toml.Unmarshal(b, &s)
+	if err != nil {
+		return nil, err
+	}
+	return &s, nil
+}
+
+func saveStore(path string, l LatestRelease, t time.Time) error {
+	s := Store{Version: l.Version, URL: l.URL, CheckedAt: t}
+	err := os.MkdirAll(path, 0755)
```
a8m, line 136: Why not `os.MkdirAll(dir, os.ModePerm)`. here you limit the write to the owner.

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return err
+	}
+	b, err := toml.Marshal(&s)
+	if err != nil {
+		return err
+	}
+	err = ioutil.WriteFile(fileLocation(path), b, 0600)
+	return err
```
a8m, line 145: 0600 - owner can read and write
0644 - owner can read and write, all can read

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,150 @@
+package update
```
a8m, line 1: add license header?

cmd/action/internal/update/update.go
yonidavidson: if the file can't be read, we prefer to recreate it - we only "pay" with an extra polling to github api.

cmd/action/internal/update/update.go
yonidavidson: the cli is stored in user permission in the user's home

cmd/action/internal/update/update.go
yonidavidson: if I don't use DI for this function it will make testing more clunky, I'll have to use httpmock (and it's an overkill)

cmd/action/internal/update/update.go
yonidavidson: checked other CLI's in my home folder all are 0755

cmd/action/internal/update/update.go
yonidavidson: it's used in a few places.

cmd/action/internal/update/update.go
yonidavidson: changed

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,152 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return LatestRelease{}, err
+	}
+	defer resp.Body.Close()
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return LatestRelease{}, err
+	}
+	if !(resp.StatusCode == http.StatusOK) {
```
a8m, line 105:
```suggestion
	if resp.StatusCode != http.StatusOK {
```

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,152 @@
...
+	err := os.MkdirAll(path, 0755)
+	if err != nil {
+		return err
+	}
+	b, err := json.Marshal(&s)
+	if err != nil {
+		return err
+	}
+	err = ioutil.WriteFile(fileLocation(path), b, 0644)
+	return err
```
a8m, line 147:
```suggestion
	return ioutil.WriteFile(fileLocation(path), b, 0644)
```

cmd/action/internal/update/update.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func localStore(path string) (*Store, error) {
+	b, err := ioutil.ReadFile(fileLocation(path))
+	if err != nil {
+		return nil, err
+	}
+	var s Store
+	err = json.Unmarshal(b, &s)
+	if err != nil {
```
a8m, line 130:
```suggestion
	if err := json.Unmarshal(b, &s); err != nil {
```

a8m (approved):
Approved, but please fix the rest of the comments 🎉

PR #342, opened 2021-12-26
sql/sqlite: support literal expressions

a8m:
In SQLite, almost any value is valid for DEFAULT. Use https://www.sqlite.org/syntax/literal-value.html to distinguish literals from raw expressions.


yonidavidson (approved):
💯

PR #345, opened 2021-12-28
sql/mysql: support auto_increment attribute in migration

a8m:


rotemtam (approved):
needs support in sqlspec, i can add in a subsequent diff

PR #346, opened 2021-12-28
sql: add MarshalHCL and UnmarshalHCL helpers to all drivers

rotemtam:
As discussed with @a8m and @zeevmoney, adding in each driver helper methods:
```go
var (
	hclState = schemahcl.New(schemahcl.WithTypes(TypeRegistry.Specs()))
	// UnmarshalHCL unmarshals an Atlas HCL DDL document into v.
	UnmarshalHCL = schemaspec.UnmarshalerFunc(func(bytes []byte, i interface{}) error {
		return UnmarshalSpec(bytes, hclState, i)
	})
	// MarshalHCL marshals v into an Atlas HCL DDL document.
	MarshalHCL = schemaspec.MarshalerFunc(func(v interface{}) ([]byte, error) {
		return MarshalSpec(v, hclState)
	})
)
```
This should greatly reduce the appearance of this lovely snippet:
```go
schemahcl.New(schemahcl.WithTypes(mysql.TypeRegistry.Specs()))
```

### DIFF
--- a/sql/postgres/sqlspec.go
+++ b/sql/postgres/sqlspec.go
@@ -8,6 +8,7 @@ import (
 	"unicode"
 
 	"ariga.io/atlas/schema/schemaspec"
+	"ariga.io/atlas/schema/schemaspec/schemahcl"
 	"ariga.io/atlas/sql/internal/specutil"
 	"ariga.io/atlas/sql/internal/sqlx"
 	"ariga.io/atlas/sql/schema"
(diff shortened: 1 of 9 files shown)


sql/postgres/sqlspec.go
```diff
@@ -226,3 +227,15 @@ var TypeRegistry = specutil.NewRegistry(
 		specutil.TypeSpec("sql", &schemaspec.TypeAttr{Name: "def", Required: true, Kind: reflect.String}),
 	),
 )
+
+var (
+	hclState = schemahcl.New(schemahcl.WithTypes(TypeRegistry.Specs()))
```
a8m, line 232: The TypeRegistry can be private imo.

a8m (approved):

sql/postgres/sqlspec.go
rotemtam: you're probably right. let me consider this a bit more

PR #350, opened 2021-12-29
sql/mysql: support realm marshal and unmarshal

rotemtam:
first PR towards multi-schema support.

### DIFF
--- a/internal/integration/mysql_test.go
+++ b/internal/integration/mysql_test.go
@@ -529,15 +529,23 @@ func TestMySQL_CLI(t *testing.T) {
 	})
 }
 
-func TestMySQL_Realm(t *testing.T) {
+func TestMySQL_HCL_Realm(t *testing.T) {
 	myRun(t, func(t *myTest) {
+		t.dropDB("second")
 		realm := t.loadRealm()
 		hcl, err := mysql.MarshalHCL(realm)
 		require.NoError(t, err)
-		var r schema.Realm
-		err = mysql.UnmarshalHCL(hcl, &r)
+		wa := string(hcl) + `
+schema "second" {
+}
+`
+		t.applyRealmHcl(wa)
+		realm, err = t.drv.InspectRealm(context.Background(), &schema.InspectRealmOption{})
 		require.NoError(t, err)
-		require.EqualValues(t, realm.Schemas[0].Name, r.Schemas[0].Name)
+		_, ok := realm.Schema("test")
+		require.True(t, ok)
+		_, ok = realm.Schema("second")
+		require.True(t, ok)
 	})
 }
 
@@ -961,6 +969,17 @@ func (t *myTest) applyHcl(spec string) {
 	require.NoError(t, err)
 }
 
+func (t *myTest) applyRealmHcl(spec string) {
+	realm := t.loadRealm()
+	var desired schema.Realm
+	err := mysql.UnmarshalHCL([]byte(spec), &desired)
+	require.NoError(t, err)
+	diff, err := t.drv.RealmDiff(realm, &desired)
+	require.NoError(t, err)
+	err = t.drv.ApplyChanges(context.Background(), diff)
+	require.NoError(t, err)
+}
+
 func (t *myTest) diff(t1, t2 *schema.Table) []schema.Change {
 	changes, err := t.drv.TableDiff(t1, t2)
 	require.NoError(t, err)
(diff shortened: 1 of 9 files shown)


internal/integration/mysql_test.go
```diff
@@ -529,6 +529,18 @@ func TestMySQL_CLI(t *testing.T) {
...
 
+func TestMySQL_Realm(t *testing.T) {
+	myRun(t, func(t *myTest) {
+		realm := t.loadRealm()
+		hcl, err := mysql.MarshalHCL(realm)
+		require.NoError(t, err)
+		var r schema.Realm
+		err = mysql.UnmarshalHCL(hcl, &r)
+		require.NoError(t, err)
+		require.EqualValues(t, realm.Schemas[0].Name, r.Schemas[0].Name)
```
yonidavidson, line 540: How many schemas do you have in this realm? maybe just the first one passed

sql/internal/specutil/convert.go
```diff
@@ -222,6 +245,23 @@ func resolveCol(ref *schemaspec.Ref, sch *schema.Schema) (*schema.Column, error)
...
+	)
+	for _, sch := range r.Schemas {
+		schemaSpec, tableSpecs, err := FromSchema(sch, fn)
+		if err != nil {
+			return nil, nil, fmt.Errorf("specutil: cannot convert schema %q: %w", sch.Name, err)
+		}
+		schemas = append(schemas, schemaSpec)
+		tables = append(tables, tableSpecs...)
+	}
+	return schemas, tables, nil
```
yonidavidson, line 262: why are we breaking it down? why not just pass the realms ? aren't the tables part of the realms?

sql/internal/specutil/spec.go
```diff
@@ -45,3 +47,64 @@ func ListAttr(k string, litValues ...string) *schemaspec.Attr {
...
+
+// Marshal marshals v into an Atlas DDL document using a schemaspec.Marshaler. Marshal uses the given
+// schemaSpec function to convert a *schema.Schema into *sqlspec.Schema and []*sqlspec.Table.
+func Marshal(v interface{}, marshaler schemaspec.Marshaler, schemaSpec func(schem *schema.Schema) (*sqlspec.Schema, []*sqlspec.Table, error)) ([]byte, error) {
+	d := &doc{}
+	switch s := v.(type) {
+	case *schema.Schema:
+		spec, tables, err := schemaSpec(s)
+		if err != nil {
+			return nil, fmt.Errorf("mysql: failed converting schema to spec: %w", err)
```
yonidavidson, line 64: should be sql no? mysql is driver specific.

sql/internal/specutil/spec.go
```diff
@@ -45,3 +47,64 @@ func ListAttr(k string, litValues ...string) *schemaspec.Attr {
...
+		}
+		d.Tables = tables
+		d.Schemas = []*sqlspec.Schema{spec}
+	case *schema.Realm:
+		for _, s := range s.Schemas {
+			spec, tables, err := schemaSpec(s)
+			if err != nil {
+				return nil, fmt.Errorf("mysql: failed converting schema to spec: %w", err)
```
yonidavidson, line 72: here too

sql/internal/specutil/spec.go
```diff
@@ -45,3 +47,64 @@ func ListAttr(k string, litValues ...string) *schemaspec.Attr {
...
+func Unmarshal(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{}, convertTable func(spec *sqlspec.Table, parent *schema.Schema) (*schema.Table, error)) error {
+	var d doc
+	if err := unmarshaler.UnmarshalSpec(data, &d); err != nil {
+		return err
+	}
+	switch v := v.(type) {
+	case *schema.Realm:
+		realm, err := Realm(d.Schemas, d.Tables, convertTable)
+		if err != nil {
+			return fmt.Errorf("postgres: failed converting to *schema.Realm: %w", err)
```
yonidavidson, line 94: replace postgres with sql

sql/internal/specutil/spec.go
```diff
@@ -45,3 +47,64 @@ func ListAttr(k string, litValues ...string) *schemaspec.Attr {
...
+		}
+		*v = *realm
+	case *schema.Schema:
+		if len(d.Schemas) != 1 {
+			return fmt.Errorf("postgres: expecting document to contain a single schema, got %d", len(d.Schemas))
```
yonidavidson, line 99: postgres->sql

sql/internal/specutil/convert.go
```diff
@@ -378,6 +418,14 @@ func FromForeignKey(s *schema.ForeignKey) (*sqlspec.ForeignKey, error) {
 	}, nil
 }
 
+func schemaName(ref *schemaspec.Ref) (string, error) {
+	parts := strings.Split(ref.V, ".")
+	if len(parts) < 2 || parts[0] != "$schema" {
+		return "", fmt.Errorf("expected ref format of $schema.name")
```
yonidavidson, line 424: is the error clear enough? maybe add some prefix like specutil:

rotemtam: good point. 
this error is always wrapped by callers that already give the package prefix. to avoid stuttering I'm keeping as is

sql/internal/specutil/convert.go
rotemtam: good point. but, currently no "realm" object in sqlspec. 
In any case, this is one of the reasons why specutil is internal, the API is not finalized and still evolving.

a8m (approved):

PR #352, opened 2021-12-30
cmd/action: support sqlite

yonidavidson:


cmd/action/mux.go
```diff
@@ -76,7 +79,28 @@ func schemaNameFromDSN(url string) (string, error) {
...
+	drv, err := sqlite.Open(db)
+	if err != nil {
+		return "", err
+	}
+	r, err := drv.InspectRealm(context.Background(), nil)
+	if err != nil {
+		return "", err
+	}
+	if len(r.Schemas) != 1 {
+		return "", fmt.Errorf("number of schemas larger then 1, n=%d", len(r.Schemas))
```
a8m, line 103:
```suggestion
		return "", fmt.Errorf("number of schemas larger than 1, n=%d", len(r.Schemas))
```

or just > 1

cmd/action/provider.go
```diff
@@ -45,3 +47,19 @@ func postgresProvider(dsn string) (*Driver, error) {
...
+	}
+	drv, err := sqlite.Open(db)
+	if err != nil {
+		return nil, err
+	}
+	return &Driver{
+		Driver:      drv,
+		Marshaler:   sqlite.MarshalHCL,
+		Unmarshaler: sqlite.UnmarshalHCL,
+		Types:       sqlite.TypeRegistry.Specs(),
```
a8m, line 63: Do we still need this after adding the UnmarshalHCL ones?

internal/integration/sqlite_test.go
```diff
@@ -422,6 +426,34 @@ create table atlas_defaults
 	})
 }
 
+func TestSQLite_CLI(t *testing.T) {
+	h := `
+			schema "main" {
+				charset   = "%s"
+				collation = "%s"
```
a8m, line 433: pretty sure this is mysql specific

internal/integration/sqlite_test.go
```diff
@@ -792,3 +824,24 @@ func (t *liteTest) dropTables(names ...string) {
...
 	})
 }
+
+// defaultConfig returns the default charset and
+// collation configuration for testing.
+func (t *liteTest) defaultAttrs() []schema.Attr {
+	var (
+		charset   = "latin1"
+		collation = "latin1_swedish_ci"
+	)
```
a8m, line 834: ditto

cmd/action/provider.go
yonidavidson: It's in all the other providers

a8m (approved):
looks good, but please change os.Mkdir to `strings.Replace("/", "_").

PR #353, opened 2021-12-30
cmd/action: add maria-db

yonidavidson:


cmd/action/mux.go
```diff
@@ -70,7 +70,7 @@ func schemaNameFromDSN(url string) (string, error) {
 		return "", err
 	}
 	switch key {
-	case "mysql":
+	case "mysql", "maria":
```
zeevmoney, line 73: maybe also `mariadb://` ?

https://mariadb.com/kb/en/about-mariadb-connector-j/#using-drivermanager

internal/integration/mysql_test.go
```diff
@@ -934,6 +934,9 @@ create table atlas_types_sanity
 }
 
 func (t *myTest) dsn() string {
+	if t.mariadb() {
+		return fmt.Sprintf("maria://root:pass@tcp(localhost:%d)/test", t.port)
+	}
 	return fmt.Sprintf("mysql://root:pass@tcp(localhost:%d)/test", t.port)
```
a8m, line 940:
```suggestion
	d := "mysql"
	if t.mariadb() {
		d = "mariadb"
	}
	return fmt.Sprintf("%s://root:pass@tcp(localhost:%d)/test", d, t.port)
```

a8m (approved):

cmd/action/mux.go
a8m: Agree with @zeevmoney

zeevmoney (approved):

PR #355, opened 2021-12-30
cmd/action: document no update env param

yonidavidson:
add atlas env command.
The long help describes all the available env' parameters.
When activating the cmd it prints the env params (when set).

This design is similar to what Go CLI does.

### DIFF
--- a/cmd/action/env.go
+++ b/cmd/action/env.go
@@ -1,0 +1,28 @@
+package action
+
+import (
+	"fmt"
+	"os"
+
+	"ariga.io/atlas/cmd/action/internal/update"
+	"github.com/spf13/cobra"
+)
+
+var envCmd = &cobra.Command{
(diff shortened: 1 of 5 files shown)


rotemtam (commented):
Yoni if this message pertains to _any_ command, I don't think it should appear in this subcommand but in a more general place describing general behavior.

yonidavidson:
It's the version command, the cli doesn't have an update command or a place that prints all env related params since this is the only environment params the cli supports. The logic for keeping it here is for the user to look to how to cancel a version check under the version command. Do you have a suggestion to place it somewhere else ?

rotemtam:
I'm sure there's an elegant solution. It doesn't make sense this way, we can figure it out together if you like

rotemtam (approved):

PR #357, opened 2022-01-01
sql/sqlite: support inspecting CHECK constraints

a8m:


sql/sqlite/inspect_test.go
```diff
@@ -307,14 +314,100 @@ CREATE TABLE t1 (
...
+		{
+			input: "CREATE TABLE t(x char check)",
+		},
+		{
+			input: "CREATE TABLE t(x char constraint x check)",
+		},
+	}
+	for _, tt := range tests {
+		table := &schema.Table{Attrs: []schema.Attr{&CreateStmt{S: tt.input}}}
+		require.NoError(t, fillChecks(table))
```
rotemtam, line 403: In general, I think unit tests should test public methods. Is there a way to check this logic without directly to an unexported method?

rotemtam (approved):
Looks good. See comment on testing via unexported method

sql/sqlite/inspect_test.go
a8m: It was @yonidavidson request the last time I’ve added a regex.

PR #361, opened 2022-01-02
sql/postgres: handle identity attributes on column creation

a8m:


sql/postgres/migrate.go
```diff
@@ -434,19 +435,31 @@ func (s *state) column(b *sqlx.Builder, c *schema.Column) {
...
 		case *schema.Comment:
 		case *schema.Collation:
-			b.P("COLLATE").Ident(attr.V)
+			b.P("COLLATE").Ident(a.V)
 		case *Identity:
-			if attr.Generation == "" {
-				attr.Generation = defaultIdentityGen
-			}
-			b.P("GENERATED", attr.Generation, "AS IDENTITY")
+			// Handled below.
```
yonidavidson, line 443: so either one attr is Identity and then we ignore it or the entire Attrs is an identity?

yonidavidson (approved):

sql/postgres/migrate.go
a8m: We (see it and ) ignore it, because we don't accept unknown attributes. However, it's handled below with a helper function.

PR #369, opened 2022-01-05
sql/schema: initial implementation for fluent api

a8m:
This DSL is still in an "experiment mode", but I plan to stabilize it after finishing the integration with ent.

### DIFF
--- a/sql/schema/dsl.go
+++ b/sql/schema/dsl.go
@@ -444,7 +444,7 @@ func (i *Index) SetTable(t *Table) *Index {
 	return i
 }
 
-// AddAttrs adds and additional attributes to the index.
+// AddAttrs adds additional attributes to the index.
 func (i *Index) AddAttrs(attrs ...Attr) *Index {
 	i.Attrs = append(i.Attrs, attrs...)
 	return i
--- a/sql/schema/dsl.go
+++ b/sql/schema/dsl.go
@@ -9,7 +9,7 @@ import (
 )
 
 // The functions and methods below provide a DSL for creating schema resources using
-// a fluent interface. Note, methods named "Add<T>" links the arguments in both directions.
+// a fluent interface. Note that some methods create links between the schema elements.
 
 // New creates a new Schema.
 func New(name string) *Schema {
@@ -162,8 +162,7 @@ func NewBoolColumn(name, typ string) *Column {
 
 // NewNullBoolColumn creates a new nullable BoolType column.
 func NewNullBoolColumn(name, typ string) *Column {
-	return NewColumn(name).
-		SetType(&BoolType{T: typ}).
+	return NewBoolColumn(name, typ).
 		SetNull(true)
 }
 
@@ -175,8 +174,7 @@ func NewIntColumn(name, typ string) *Column {
 
 // NewNullIntColumn creates a new nullable IntegerType column.
 func NewNullIntColumn(name, typ string) *Column {
-	return NewColumn(name).
-		SetType(&IntegerType{T: typ}).
+	return NewIntColumn(name, typ).
 		SetNull(true)
 }
 
@@ -188,8 +186,7 @@ func NewUintColumn(name, typ string) *Column {
 
 // NewNullUintColumn creates a new nullable unsigned IntegerType column.
 func NewNullUintColumn(name, typ string) *Column {
-	return NewColumn(name).
-		SetType(&IntegerType{T: typ, Unsigned: true}).
+	return NewUintColumn(name, typ).
 		SetNull(true)
 }
 
@@ -223,7 +220,8 @@ func NewEnumColumn(name string, opts ...EnumOption) *Column {
 
 // NewNullEnumColumn creates a new nullable EnumType column.
 func NewNullEnumColumn(name string, opts ...EnumOption) *Column {
-	return NewEnumColumn(name, opts...).SetNull(true)
+	return NewEnumColumn(name, opts...).
+		SetNull(true)
 }
 
 // BinaryOption allows configuring BinaryType using functional options.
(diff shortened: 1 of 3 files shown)


rotemtam (approved):

sql/schema/dsl.go
```diff
@@ -0,0 +1,583 @@
...
+}
+
+// SetComment sets or appends the Comment attribute
+// to the schema with the given value.
+func (s *Schema) SetComment(v string) *Schema {
+	replaceOrAppend(&s.Attrs, &Comment{Text: v})
+	return s
+}
+
+// AddAttrs adds and additional attributes to the schema.
```
masseelch, line 40:
```suggestion
// AddAttrs adds additional attributes to the schema.
```

sql/schema/dsl.go
```diff
@@ -0,0 +1,583 @@
...
+// AddChecks appends the given checks to the attribute list.
+func (t *Table) AddChecks(checks ...*Check) *Table {
+	for _, c := range checks {
+		t.Attrs = append(t.Attrs, c)
+	}
+	return t
+}
+
+// SetSchema sets the schema (named-database) of the table.
+func (t *Table) SetSchema(s *Schema) *Table {
```
masseelch, line 96: Why the decision to only link back on `AddT` and not here? If not combined with `SetTable` on the Schema this can lead to "incomplete" data, correct?

sql/schema/dsl.go
```diff
@@ -0,0 +1,583 @@
...
+	t.Schema = s
+	return t
+}
+
+// SetPrimaryKey sets the primary-key of the table.
+func (t *Table) SetPrimaryKey(pk *Index) *Table {
```
masseelch, line 102: In this case you do a "link-back" even though the method is not named `AddT`

sql/schema/dsl.go
```diff
@@ -0,0 +1,583 @@
...
+	pk.Table = t
+	t.PrimaryKey = pk
+	for _, p := range pk.Parts {
+		if p.C == nil {
+			continue
+		}
+		if _, ok := t.Column(p.C.Name); !ok {
+			t.AddColumns(p.C)
```
masseelch, line 110: Here again. This is inconsistent behaviour imo :thinking:

sql/schema/dsl.go
```diff
@@ -0,0 +1,583 @@
...
+// NewColumn creates a new column with the given name.
+func NewColumn(name string) *Column {
+	return &Column{Name: name}
+}
+
+// NewNullColumn creates a new nullable column with the given name.
+func NewNullColumn(name string) *Column {
+	c := &Column{Name: name}
+	c.SetNull(true)
+	return c
```
masseelch, line 155:
```suggestion
	return NewColumn(name).
		SetNull(true)
```

sql/schema/dsl.go
```diff
@@ -0,0 +1,583 @@
...
+
+// NewBoolColumn creates a new BoolType column.
+func NewBoolColumn(name, typ string) *Column {
+	return NewColumn(name).
+		SetType(&BoolType{T: typ})
+}
+
+// NewNullBoolColumn creates a new nullable BoolType column.
+func NewNullBoolColumn(name, typ string) *Column {
+	return NewColumn(name).
```
masseelch, line 166: I'd use `NewNullColumn` inside the `NewNull...` builders

sql/schema/dsl.go
```diff
@@ -0,0 +1,583 @@
...
+
+// NewIntColumn creates a new IntegerType column.
+func NewIntColumn(name, typ string) *Column {
+	return NewColumn(name).
+		SetType(&IntegerType{T: typ})
+}
+
+// NewNullIntColumn creates a new nullable IntegerType column.
+func NewNullIntColumn(name, typ string) *Column {
+	return NewColumn(name).
```
masseelch, line 179: dito

sql/schema/dsl.go
```diff
@@ -0,0 +1,583 @@
...
+
+// NewUintColumn creates a new unsigned IntegerType column.
+func NewUintColumn(name, typ string) *Column {
+	return NewColumn(name).
+		SetType(&IntegerType{T: typ, Unsigned: true})
+}
+
+// NewNullUintColumn creates a new nullable unsigned IntegerType column.
+func NewNullUintColumn(name, typ string) *Column {
+	return NewColumn(name).
```
masseelch, line 192: dito

sql/schema/dsl.go
```diff
@@ -0,0 +1,583 @@
...
+}
+
+// SetComment sets or appends the Comment attribute
+// to the column with the given value.
+func (c *Column) SetComment(v string) *Column {
+	replaceOrAppend(&c.Attrs, &Comment{Text: v})
+	return c
+}
+
+// AddAttrs adds and additional attributes to the column.
```
masseelch, line 408:
```suggestion
// AddAttrs adds additional attributes to the column.
```

sql/schema/dsl.go
```diff
@@ -0,0 +1,583 @@
...
+	return i
+}
+
+// SetTable configures the table of the index.
+func (i *Index) SetTable(t *Table) *Index {
+	i.Table = t
+	return i
+}
+
+// AddAttrs adds and additional attributes to the index.
```
masseelch, line 448:
```suggestion
// AddAttrs adds additional attributes to the index.
```

masseelch (approved):

sql/schema/dsl.go
a8m: No real reason, I use "AddTables" in most cases (and not "SetSchema") and wanted to avoid checking if the table exists in the schema for each "Add" call. However, I still can add this, no strong opinion here.

sql/schema/dsl.go
a8m: Right. The comment at the top was written before I wrote this method.

sql/schema/dsl.go
a8m: 👍

a8m: let's talk about it f2f and I'll handle this in future PRs if needed.

PR #372, opened 2022-01-05
cmd/action: sqlite make sure file exist

zeevmoney:


cmd/action/internal/uri/uri.go
```diff
@@ -0,0 +1,44 @@
...
+func SqliteExists(dsn string) error {
+	if !inMemory.MatchString(dsn) {
+		return fileExists(dsn)
+	}
+	return nil
+}
+
+func fileExists(dsn string) error {
+	s := strings.Split(dsn, "?")
+	fn := dsn
```
a8m, line 28: fn usually means "functions". use the dsn instead
```suggestion
```

cmd/action/internal/uri/uri.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	fn = filepath.Clean(fn)
+	_, err := os.Stat(fn)
+	if errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("file %s does not exist", fn)
+	}
+	if errors.Is(err, os.ErrPermission) {
+		return fmt.Errorf("no permission to access file %s", fn)
+	}
+	return nil
```
a8m, line 43:
```suggestion
	if _, err := os.Stat(fn); err != nil {
		return fmt.Errorf("failed opening %q: %w", dsn, err)
	}
	return nil
```

cmd/action/internal/uri/uri.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2021-present The Atlas Authors. All rights reserved.
+// This source code is licensed under the Apache 2.0 license found
+// in the LICENSE file in the root directory of this source tree.
+
+package uri
```
a8m, line 5: don't create another package for one function. leave private in the `action` package.

cmd/action/internal/uri/uri.go
```diff
@@ -0,0 +1,44 @@
...
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+var inMemory = regexp.MustCompile("^file:.*:memory:$|:memory:|^file:.*mode=memory.*")
+
+// SqliteExists returns nil if the sqlite dsn is in memory or file exists, otherwise error.
+func SqliteExists(dsn string) error {
```
a8m, line 19:
```suggestion
func SQLiteExists(dsn string) error {
```

cmd/action/internal/uri/uri.go
zeevmoney: it's a file name (potential), I'll use f

cmd/action/internal/uri.go
```diff
@@ -0,0 +1,39 @@
+// Copyright 2021-present The Atlas Authors. All rights reserved.
+// This source code is licensed under the Apache 2.0 license found
+// in the LICENSE file in the root directory of this source tree.
+
+package action
```
a8m, line 5: Why it’s here? Leave it in cmd/action

cmd/action/mux_test.go
```diff
@@ -1,9 +1,13 @@
 package action_test
 
 import (
+	"io/ioutil"
+	"os"
 	"testing"
 
 	"ariga.io/atlas/cmd/action"
+
+	_ "github.com/mattn/go-sqlite3"
```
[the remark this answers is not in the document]
zeevmoney: The function `schemaName` which is called from `SchemaNameFromSN` is using the sqlite3 driver.

a8m (approved):

PR #373, opened 2022-01-05
sql/mysql: export column types for external usage

a8m:


rotemtam (approved):
Cc @zeevmoney i think these are duplicated in entimport

PR #389, opened 2022-01-10
sql/internal/specutil: support marshaling unsupported types as sql() …

rotemtam:
…expressions in HCL for MySQL and SQLite

this brings support for bi-directional conversion of columns with 

```
type = sql("custom")
```

this PR solves the issue for MySQL and SQLite.
With Postgres there's a small issue @a8m and I will discuss tommorow


a8m (approved):
💯

PR #393, opened 2022-01-11
sql/mysql: avoid panic on migrate failure

a8m:
Unsupported types should be rejected in MySQL, but should
not cause the migration process to fail/panic


rotemtam (approved):
Thanks !

PR #404, opened 2022-01-13
Add support for SQL comments

masseelch:
This PR adds support for adding comments to the HCL. Opposed to the previous implementation where there was an `Comment` field added the the `sqlspec` structs this is no longer the case since SQLite for example does not support comments.

The current implementation also makes sure that no `commen = ""` attributes are rendered to the marshaled HCL if there are no comments.

### DIFF
--- a/sql/internal/specutil/convert.go
+++ b/sql/internal/specutil/convert.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"ariga.io/atlas/schema/schemaspec"
+	"ariga.io/atlas/sql/internal/sqlx"
 	"ariga.io/atlas/sql/schema"
 	"ariga.io/atlas/sql/sqlspec"
 )
@@ -106,6 +107,9 @@ func Table(spec *sqlspec.Table, parent *schema.Schema, convertColumn ConvertColu
 		}
 		tbl.Indexes = append(tbl.Indexes, i)
 	}
+	if err := convertCommentFromSpec(spec, &tbl.Attrs); err != nil {
+		return nil, err
+	}
 	return tbl, nil
 }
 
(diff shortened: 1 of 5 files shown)


sql/internal/specutil/convert.go
```diff
@@ -457,3 +472,25 @@ func colRef(cName string, tName string) *schemaspec.Ref {
 func schemaRef(n string) *schemaspec.Ref {
 	return &schemaspec.Ref{V: "$schema." + n}
 }
+
+// convertCommentFromSpec converts a spec comment attribute to a schema element attribute.
+func convertCommentFromSpec(spec interface {
+	Attr(string) (*schemaspec.Attr, bool)
+}, attrs *[]schema.Attr) error {
```
a8m, line 479:
```suggestion
func convertCommentFromSpec(spec interface {Attr(string) (*schemaspec.Attr, bool) }, attrs *[]schema.Attr) error {
```

Not that long for adding this line break (I'm not fan of it tbh)

sql/mysql/sqlspec_test.go
```diff
@@ -272,6 +284,79 @@ schema "test" {
...
+			Unique: true,
+			Parts:  []*schema.IndexPart{{SeqNo: 0, C: s.Tables[0].Columns[0]}},
+			Attrs: []schema.Attr{
+				&schema.Comment{Text: "index comment"},
+			},
+		},
+	}
+	buf, err := MarshalSpec(s, hclState)
+	require.NoError(t, err)
+	// We expect a zero value comment to not bre present in the marshaled HCL.
```
a8m, line 332:
```suggestion
	// We expect a zero value comment to not be present in the marshaled HCL.
```

a8m (commented):
The code looks good to me, but let's wait for @rotemtam to approve this.

rotemtam (approved):

sql/internal/specutil/convert.go
masseelch: `gfmt` thinks otherwise. `gofmt` is king :man_shrugging:

PR #405, opened 2022-01-13
beautify code

masseelch:
Put inline interface into its own type

### DIFF
--- a/sql/internal/specutil/convert.go
+++ b/sql/internal/specutil/convert.go
@@ -473,10 +473,13 @@ func schemaRef(n string) *schemaspec.Ref {
 	return &schemaspec.Ref{V: "$schema." + n}
 }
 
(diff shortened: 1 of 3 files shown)


sql/internal/specutil/convert.go
```diff
@@ -473,10 +473,12 @@ func schemaRef(n string) *schemaspec.Ref {
 	return &schemaspec.Ref{V: "$schema." + n}
 }
 
-// convertCommentFromSpec converts a spec comment attribute to a schema element attribute.
-func convertCommentFromSpec(spec interface {
+type Attr interface {
```
rotemtam, line 476:
```suggestion
// Attr is the interface that wraps the Attr method.
type Attr interface {
```

a8m (approved):

PR #407, opened 2022-01-13
sql/mysql: reject CHARSET/COLLATE attributes for non-char columns on migrate

a8m:
- Reject CHARSET/COLLATE attributes for non-char columns on migration
- "Normalize" inspected JSON columns on MariaDB
- Add UnsetCollate and UnsetCharset


rotemtam (approved):
💯

PR #412, opened 2022-01-14
sql/postgres: minor bug fixes for modify-enum

a8m:


sql/postgres/migrate.go
```diff
@@ -208,20 +208,23 @@ func (s *state) modifyTable(ctx context.Context, modify *schema.ModifyTable) err
...
 			case ok1 && ok2 && from.T == to.T:
 				if err := s.alterType(from, to); err != nil {
 					return err
 				}
-			// Not an enum, or was dropped.
-			default:
-				changes = append(changes, change)
+				// If only the enum values were changed,
+				// there is no need to ALTER the table.
+				if change.Change == schema.ChangeType {
```
masseelch, line 218: I am confused. There is no need to alter the table but it is altered just above this block

sql/postgres/migrate.go
```diff
@@ -208,20 +208,23 @@ func (s *state) modifyTable(ctx context.Context, modify *schema.ModifyTable) err
...
+				}
+			// Enum was added (and column type was changed).
+			case !ok1 && ok2:
+				if err := s.addTypes(ctx, change.To); err != nil {
+					return err
+				}
 			}
+			changes = append(changes, change)
 		default:
 			changes = append(changes, change)
```
masseelch, line 229: Why did you lose the `// Not an enum, or was dropped.` comment?

sql/postgres/migrate.go
a8m: If only the enum was changed (e.g. add values), there's no need to modify the column (on ALTER TABLE). See tests, or let me know if it needs to phrasing work.

sql/postgres/migrate.go
a8m: It's not needed. It was replaced with the comment of the negated case.

masseelch: Missed that, sorry.

masseelch (approved):

PR #415, opened 2022-01-14
sql/postgres: finish handling comments creation/modification for indexes, tables and columns

a8m:


sql/postgres/migrate.go
```diff
@@ -166,28 +166,43 @@ func (s *state) dropTable(drop *schema.DropTable) {
 	})
 }
 
-// modifyTable builds and executes the queries for bringing the table into its modified state.
+// modifyTable builds the statements that bring the table into its modified state.
```
masseelch, line 169:
```suggestion
// modifyTable builds the statements that carry the table into its modified state.
```

sql/postgres/migrate.go
```diff
@@ -166,28 +166,43 @@ func (s *state) dropTable(drop *schema.DropTable) {
...
 			dropI = append(dropI, change.I)
 		case *schema.ModifyIndex:
+			k := change.Change
+			if change.Change.Is(schema.ChangeComment) {
+				from, to, err := commentChange(sqlx.CommentDiff(change.From.Attrs, change.To.Attrs))
+				if err != nil {
+					return err
+				}
+				comments = append(comments, s.indexComment(modify.T, change.To, to, from))
+				// If only the index comment was changed.
```
masseelch, line 201:
```suggestion
				// If only the index' comment was changed.
```

masseelch (approved):

sql/postgres/migrate.go
a8m: Not a native speaker, but not sure that means the same. Am I wrong?

a8m: cc @rotemtam

sql/postgres/migrate.go
a8m: 👍

PR #416, opened 2022-01-14
sql/mysql: support modifying schema charset and collation attributes

a8m:
We can consider to add schema changes as ModifyCharset and ModifyCollation
in the future to avoid all the type assertion.


masseelch (commented):
LGTM, but CI is failing

rotemtam (approved):
🙏

PR #417, opened 2022-01-15
sql/mysql: reject CHECK constraints when it is not supported by the d…

a8m:
…atabase

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,3 +1,6 @@
+run:
+  timeout: 3m
+
 issues:
   include:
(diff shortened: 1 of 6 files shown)


masseelch:
What about working around this by adding a trigger instead if a check constraint prior to version 8?

a8m:
> What about working around this by adding a trigger instead if a check constraint prior to version 8?

Not at this stage. I think we should add support in HCL for triggers, events, etc.

masseelch (approved):
💯

PR #433, opened 2022-01-18
sql/schema: simplify the ExecQuerier interface.

a8m:
Make it simpler to implement it by tools like Ent

### DIFF
--- a/sql/internal/sqlx/sqlx.go
+++ b/sql/internal/sqlx/sqlx.go
@@ -20,6 +20,18 @@ func ValidString(s sql.NullString) bool {
 	return s.Valid && s.String != "" && strings.ToLower(s.String) != "null"
 }
 
+// ScanOne scans one record and closes the rows at the end.
+func ScanOne(rows *sql.Rows, dest ...interface{}) error {
+	defer rows.Close()
+	if !rows.Next() {
+		return sql.ErrNoRows
+	}
+	if err := rows.Scan(dest...); err != nil {
+		return err
+	}
+	return rows.Close()
+}
+
 // ScanFKs scans the rows and adds the foreign-key to the table.
 // Reference elements are added as stubs and should be linked
 // manually by the caller.
--- a/sql/postgres/inspect.go
+++ b/sql/postgres/inspect.go
@@ -51,8 +51,12 @@ func (i *inspect) InspectSchema(ctx context.Context, name string, opts *schema.I
 	var schemas []*schema.Schema
 	switch name {
 	case "":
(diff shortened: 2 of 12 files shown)


sql/postgres/inspect.go
```diff
@@ -121,9 +125,14 @@ func (i *inspect) table(ctx context.Context, name string, opts *schema.InspectTa
 		query = tableSchemaQuery
 		args = append(args, opts.Schema)
 	}
-	row := i.QueryRowContext(ctx, query, args...)
-	var tSchema, comment sql.NullString
-	if err := row.Scan(&tSchema, &comment); err != nil {
+	var (
+		tSchema, comment sql.NullString
+		rows, err        = i.QueryContext(ctx, query, args...)
```
yonidavidson, line 130: is this correct formatting?

yonidavidson (approved):

sql/schema/inspect.go
```diff
@@ -27,11 +27,8 @@ func IsNotExistError(err error) bool {
 	return errors.As(err, &e)
 }
 
-// ExecQuerier wraps the standard sql.DB methods.
+// ExecQuerier wraps the 2 standard sql.DB methods.
```
masseelch, line 30:
```suggestion
// ExecQuerier wraps the two standard sql.DB methods.
```

sql/internal/sqlx/sqlx.go
```diff
@@ -20,6 +20,18 @@ func ValidString(s sql.NullString) bool {
 	return s.Valid && s.String != "" && strings.ToLower(s.String) != "null"
 }
 
+// ScanOne scans one record and closes the rows at the end.
+func ScanOne(rows *sql.Rows, dest ...interface{}) error {
```
masseelch, line 24: This should error out if there are more than 1 row, should't it?

sql/postgres/inspect.go
a8m: Yes.

sql/internal/sqlx/sqlx.go
a8m: I tried to keep the same behavior as `sql.Row`.

masseelch: I was thinking about [database/sql](https://pkg.go.dev/database/sql#DB.QueryRow). It errors if there is more than one row in the result set.

masseelch: Sorry, it behaves exactly like you said.

PR #441, opened 2022-01-19
atlas: preparations for referenced column types

masseelch:
This PR introdues first steps to enable referenced column types. This is needed for the postgres implementation of enums.

The goal is to support HCL like the following:

```hcl
schema "sch" {}
table "tbl" {
  column "col" {
    type = enum.my_enum
  }
}
enum "my_enum" {
  values = ["on", "off"]
}
```

The first approach was to change `sqlspec.Table.Colum` from `*schemaspec.Type` to `schemaspec.Value` but handling a `Value` required different handling on `sqlspec.Table.Column.Type` and `sqlspec.Table.Column.Default` and the current structure didn't allow for this.

The current approach is valid too and all the implementation details are hidden from the actual `schema.Schema` the atlas internals work with.

Please note, that implementation for converting between `sqlspec.Schema` and `schema.Schema` will follow in another PR.

### DIFF
--- a/sql/internal/specutil/spec.go
+++ b/sql/internal/specutil/spec.go
@@ -79,7 +79,8 @@ func Marshal(v interface{}, marshaler schemaspec.Marshaler, schemaSpec func(sche
 
 // Unmarshal unmarshals an Atlas DDL document using an unmarshaler into v. Unmarshal uses the
 // given convertTable function to convert a *sqlspec.Table into a *schema.Table.
-func Unmarshal(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{}, convertTable func(spec *sqlspec.Table, parent *schema.Schema) (*schema.Table, error)) error {
+func Unmarshal(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{},
+	convertTable ConvertTableFunc) error {
 	var d doc
 	if err := unmarshaler.UnmarshalSpec(data, &d); err != nil {
 		return err
--- a/sql/mysql/sqlspec_test.go
+++ b/sql/mysql/sqlspec_test.go
@@ -661,10 +661,6 @@ func TestTypes(t *testing.T) {
 			expected: &schema.StringType{T: TypeLongText},
 		},
 		{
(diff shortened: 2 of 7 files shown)


sql/mysql/sqlspec_test.go
```diff
@@ -660,10 +660,6 @@ func TestTypes(t *testing.T) {
 			typeExpr: "longtext",
 			expected: &schema.StringType{T: TypeLongText},
 		},
-		{
-			typeExpr: `enum("a","b")`,
-			expected: &schema.EnumType{T: "enum", Values: []string{"a", "b"}},
-		},
```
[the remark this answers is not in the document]
masseelch: https://github.com/ariga/atlas/blob/e2d7dad198ad537e6045e37ea13f6924c1f893a5/sql/mysql/sqlspec_test.go#L528

Was a duplicate test. There still is one.

sql/sqlspec/sqlspec.go
```diff
@@ -25,6 +25,14 @@ type (
...
 	}
 
+	// Enum holds a specification for an enum, that can be referenced as a column type.
+	Enum struct {
+		Name   string          `spec:",name"`
+		Schema *schemaspec.Ref `spec:"schema"`
+		Values []string        `spec:"values"`
+		schemaspec.DefaultExtension
+	}
+
```
rotemtam, line 35: shouldn't this be in the postgres package?

sql/internal/specutil/spec.go
```diff
@@ -48,6 +48,7 @@ func ListAttr(k string, litValues ...string) *schemaspec.Attr {
 type doc struct {
 	Tables  []*sqlspec.Table  `spec:"table"`
 	Schemas []*sqlspec.Schema `spec:"schema"`
+	Enums   []*sqlspec.Enum   `spec:"enum"`
```
rotemtam, line 51: should be specific to postgres imo

masseelch:
@rotemtam moved it into `postgres`. Please have a look again.

rotemtam (approved):

PR #443, opened 2022-01-19
sqlx: fix potential panic when array element is nil

zeevmoney:


a8m (approved):

sql/internal/sqlx/diff.go
```diff
@@ -383,7 +383,7 @@ func Has(elements, target interface{}) bool {
 		panic("target must be a non-nil pointer")
 	}
 	for i := 0; i < ev.Len(); i++ {
-		if e := ev.Index(i).Elem(); e.Type().AssignableTo(tv.Type()) {
+		if e := ev.Index(i).Elem(); !e.IsNil() && e.Type().AssignableTo(tv.Type()) {
```
a8m, line 386:
```suggestion
		if e := ev.Index(i).Elem(); !e.IsValid() && e.Type().AssignableTo(tv.Type()) {
```

Use IsValid instead?

PR #444, opened 2022-01-19
sql/mysql: use DROP CHECK version that works both on MariaDB and MySQL

a8m:


sql/mysql/diff.go
```diff
@@ -50,9 +50,26 @@ func (d *diff) TableAttrDiff(from, to *schema.Table) ([]schema.Change, error) {
...
+	// or by Atlas for older versions. These CHECK constraints (inlined on the columns)
+	// also cannot be dropped using "DROP CONSTRAINTS", but can be modified and dropped
+	// using "MODIFY COLUMN".
+	var checks []schema.Change
+	for _, c := range sqlx.CheckDiff(from, to, func(c1, c2 *schema.Check) bool {
 		return c1.Expr != c2.Expr || sqlx.Has(c1.Attrs, &Enforced{}) != sqlx.Has(c2.Attrs, &Enforced{})
-	})...), nil
+	}) {
+		drop, ok := c.(*schema.DropCheck)
+		if !ok || !strings.HasPrefix(drop.C.Expr, "json_valid") {
```
masseelch, line 62: Is atlas already that spread we have to do this?

masseelch (approved):

PR #445, opened 2022-01-19
sql/postgres: unmarshal HCL enum

masseelch:
This is the `Unmarshal` part of the postgres support for enums.

### DIFF
--- a/sql/postgres/sqlspec.go
+++ b/sql/postgres/sqlspec.go
@@ -40,7 +40,7 @@ func UnmarshalSpec(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{
 	}
 	switch v := v.(type) {
 	case *schema.Realm:
-		realm, err := Realm(d.Schemas, d.Tables)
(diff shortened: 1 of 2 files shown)


sql/postgres/sqlspec.go
```diff
@@ -65,12 +66,15 @@ func MarshalSpec(v interface{}, marshaler schemaspec.Marshaler) ([]byte, error)
 	return specutil.Marshal(v, marshaler, schemaSpec)
 }
 
-// Realm converts the schemas and tables into a schema.Realm.
-func Realm(schemas []*sqlspec.Schema, tables []*sqlspec.Table) (*schema.Realm, error) {
+// Realm converts the schemas and tables of the doc into a schema.Realm.
+func Realm(d doc) (*schema.Realm, error) {
```
rotemtam, line 70: I prefer not to use an unexported type as input to an exported method

masseelch: True

a8m (approved):

PR #447, opened 2022-01-20
sql/postgres: marshal HCL enum

masseelch:
This is the `Marshal` part of the postgres support for enums.

### DIFF
--- a/sql/postgres/sqlspec.go
+++ b/sql/postgres/sqlspec.go
@@ -63,7 +63,30 @@ func UnmarshalSpec(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{
 
 // MarshalSpec marshals v into an Atlas DDL document using a schemaspec.Marshaler.
 func MarshalSpec(v interface{}, marshaler schemaspec.Marshaler) ([]byte, error) {
-	return specutil.Marshal(v, marshaler, schemaSpec)
+	d := &doc{}
+	switch s := v.(type) {
+	case *schema.Schema:
+		spec, tables, enums, err := schemaSpec(s)
+		if err != nil {
+			return nil, fmt.Errorf("specutil: failed converting schema to spec: %w", err)
+		}
+		d.Tables = tables
+		d.Schemas = []*sqlspec.Schema{spec}
+		d.Enums = enums
+	case *schema.Realm:
+		for _, s := range s.Schemas {
+			spec, tables, enums, err := schemaSpec(s)
+			if err != nil {
+				return nil, fmt.Errorf("specutil: failed converting schema to spec: %w", err)
+			}
+			d.Tables = append(d.Tables, tables...)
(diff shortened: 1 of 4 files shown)


sql/postgres/sqlspec.go
```diff
@@ -63,7 +63,30 @@ func UnmarshalSpec(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{
 
 // MarshalSpec marshals v into an Atlas DDL document using a schemaspec.Marshaler.
 func MarshalSpec(v interface{}, marshaler schemaspec.Marshaler) ([]byte, error) {
-	return specutil.Marshal(v, marshaler, schemaSpec)
+	d := &doc{}
```
yonidavidson, line 66: var d doc

sql/postgres/sqlspec.go
```diff
@@ -63,7 +63,30 @@ func UnmarshalSpec(data []byte, unmarshaler schemaspec.Unmarshaler, v interface{
...
+	switch s := v.(type) {
+	case *schema.Schema:
+		spec, tables, enums, err := schemaSpec(s)
```
yonidavidson, line 69: I am not a fan of this API extending the number of returned values.
maybe we should return (struct,err) ?

sql/postgres/sqlspec.go
```diff
@@ -216,9 +239,32 @@ func enumName(ref *schemaspec.Type) (string, error) {
...
+func enumRef(n string) *schemaspec.Ref {
+	return &schemaspec.Ref{
+		V: "$enum." + n,
+	}
+}
+
 // schemaSpec converts from a concrete Postgres schema to Atlas specification.
-func schemaSpec(schem *schema.Schema) (*sqlspec.Schema, []*sqlspec.Table, error) {
-	return specutil.FromSchema(schem, tableSpec)
+func schemaSpec(schem *schema.Schema) (*sqlspec.Schema, []*sqlspec.Table, []*Enum, error) {
```
yonidavidson, line 250: https://github.com/ariga/atlas/pull/447/files#r788467466

sql/postgres/sqlspec.go
masseelch: `var d doc` <- not a pointer

sql/postgres/sqlspec.go
masseelch: schemaSpec is unimported and therefore I think this is okay. If it had been an exported function, I'd agree with you. Let's see what @a8m thinks :-)

sql/postgres/sqlspec.go
a8m: I'd return a doc instead.

a8m (approved):
Looks good, besides one comment. Returning more than 2 values from a function is not that common in this project.

rotemtam (approved):

PR #452, opened 2022-01-20
sql/mysql: add reverse to DropIndex

hedwigz:


sql/mysql/migrate_test.go
```diff
@@ -502,9 +543,9 @@ func TestPlanChanges(t *testing.T) {
 		require.NotNil(t, plan)
 		require.Equal(t, tt.wantPlan.Reversible, plan.Reversible)
 		require.Equal(t, tt.wantPlan.Transactional, plan.Transactional)
-		for i, c := range plan.Changes {
-			require.Equal(t, tt.wantPlan.Changes[i].Cmd, c.Cmd)
-			require.Equal(t, tt.wantPlan.Changes[i].Reverse, c.Reverse)
+		for i, c := range tt.wantPlan.Changes {
```
a8m, line 546: Maybe add a check that `len` is equal?

a8m (approved):
Looks good, besides a minor commnet. Thanks 🙏

sql/mysql/migrate_test.go
hedwigz: :+1:

Issue #457, opened 2022-01-23
Postgres schema selection


cthackers:
Hi, 
Is there a way to define what schema I want it to inspect/migrate ? 
From [here](https://github.com/ariga/atlas/blob/11ba264cfa606ce987375e0a08e7888e40711475/cmd/action/mux.go#L93) it seems is hardcoded to only use `public` schema. The project I'm working on uses 3 different schemas and `public` is not used at all. Running inspect on my database gives me an empty result.

CGA1123:
@cthackers I wonder whether a patch like this fixes your issue? https://github.com/cga1123/atlas/commit/50991749f48dccbbd073c405dac91e4697f18a1a

a8m:
CGA1123@5099174 looks great to me.

cc @rotemtam, @yonidavidson

rotemtam:
hey @CGA1123 ,
that looks great, do you mind sending a PR so we can add it to the next version?

CGA1123:
Yep can roll it into a PR this evening, unless you want to cherry pick yourself earlier?

rotemtam:
No rush, when you have the time :-)

PR #470, opened 2022-01-23
cmd/action: Parse schema name in postgres dsn

CGA1123:
`lib/pq` supports parsing runtime parameters from the DSN. Allowing for
configuration of the schema to inspect or apply on using the
`search_path` parameter.

Closes: #457

### DIFF
--- a/cmd/action/mux.go
+++ b/cmd/action/mux.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


a8m (approved):
Looks good to me. Thanks for the contribution @CGA1123 🙏

cmd/action/mux.go
```diff
@@ -90,14 +91,34 @@ func SchemaNameFromDSN(url string) (string, error) {
...
+	// lib/pq supports setting default schemas via the `search_path` parameter
+	// in a dsn.
+	//
+	// See: https://github.com/lib/pq/blob/8446d16b8935fdf2b5c0fe333538ac395e3e1e4b/conn.go#L1155-L1165
+	queries := url.Query()
+	if queries.Has("search_path") {
+		return queries.Get("search_path"), nil
+	}
+
+	return "public", nil
```
a8m, line 119: One comment, maybe skip it if the "search_path" is empty or return an error? (e.g. `"postgres://localhost:5432/dbname?search_path="`).

```suggestion
	url, err := url.Parse(dsn)
	if err != nil {
		// For backwards compatibility, we default to "public" when failing to
		// parse.
		return "public", nil
	}
	// lib/pq supports setting default schemas via the `search_path` parameter
	// in a dsn.
	//
	// See: https://github.com/lib/pq/blob/8446d16b8935fdf2b5c0fe333538ac395e3e1e4b/conn.go#L1155-L1165
	if schema := url.Query().Get("search_path"); schema != "" {
		return schema, nil
	}
```

a8m: I approved the PR but then thought about the "empty value" case. 
cc @CGA1123 @masseelch

rotemtam (approved):
Thanks for your contribution @CGA1123 , if you haven't yet, please join our community on our discord server!

https://discord.com/invite/QhsmBAWzrC

PR #474, opened 2022-01-24
sql/mysql: inspect: support fractional seconds precision in timestamp…

### LINKED ISSUE #471: Support schema inspect for Mysql table with column "Timestamp/Datetime default on update"
We have a column that is in Datetime type and has the [default on update definition](https://dev.mysql.com/doc/refman/5.7/en/timestamp-initialization.html). For example:

```
CREATE TABLE `foo` (
 `id`			INT,
 `update_time`	DATETIME(6)  NOT NULL DEFAULT CURRENT_TIMESTAMP(6) ON UPDATE CURRENT_TIMESTAMP(6)

  PRIMARY KEY (`id`),
)
```

When I try to inspect the schema, I got the following error: 
```
atlas schema inspect -d "mysql://root:secret@tcp(localhost:12334)/my_database"
Error: mysql: unknown attribute "on update current_timestamp(6)"
```

Mysql version: 5.7.35

Atlas version: 
```
atlas CLI version v0.3.1
https://github.com/ariga/atlas/releases/tag/v0.3.1
```

masseelch:
… type, its default and on update clause

Fix #471

### DIFF
--- a/sql/mysql/inspect.go
+++ b/sql/mysql/inspect.go
@@ -450,7 +450,7 @@ func (i *inspect) tableNames(ctx context.Context, schema string, opts *schema.In
 	return names, nil
 }
 
-var crrTsReg = regexp.MustCompile(`^(?:default_generated )?on update current_timestamp(?:\(\d?\))?$`)
+var reTimeOnUpdate = regexp.MustCompile(`^(?:default_generated )?on update current_timestamp(?:\(\d?\))?$`)
 
 // extraAttr parses the EXTRA column from the INFORMATION_SCHEMA.COLUMNS table
 // and appends its parsed representation to the column.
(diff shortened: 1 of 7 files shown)


sql/mysql/inspect.go
```diff
@@ -514,6 +516,8 @@ func (i *inspect) createStmt(ctx context.Context, t *schema.Table) error {
 	return nil
 }
 
+var crrTsDefReg = regexp.MustCompile(`(?i)^current_timestamp(?:\(\d?\))?$`)
```
a8m, line 519:
```suggestion
var reCurrTS = regexp.MustCompile(`(?i)^current_timestamp(?:\(\d?\))?$`)
```

a8m (approved):
Looks goooooood ❤️

PR #476, opened 2022-01-24
sql/postgres: inspect: support fractional seconds precision in timest…

masseelch:
…amp type


sql/postgres/inspect.go
```diff
@@ -228,7 +232,7 @@ func (i *inspect) addColumn(t *schema.Table, rows *sql.Rows) error {
 
 func columnType(c *columnDesc) schema.Type {
 	var typ schema.Type
-	switch t := c.typ; strings.ToLower(t) {
+	switch t := strings.Split(c.typ, "(")[0]; strings.ToLower(t) {
```
a8m, line 235: Do you have examples in which the `data_type` returned with `(`? I'm pretty sure the test examples are not correct (e.g. `timestamp(4)`).

masseelch: I'll check

a8m (approved):
fantastic 💯

PR #477, opened 2022-01-24
sql/postgres: add timestamptz to type-registry

a8m:


masseelch (approved):
What about `timetz`?

a8m:
> timetz

Missed it. Will add 🙏

a8m:
OK, found this https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timetz

PR #482, opened 2022-01-24
sql/mysql: add support for bool/boolean types

a8m:
See first 2 rows in https://dev.mysql.com/doc/refman/8.0/en/other-vendor-data-types.html.
The rest are not common.


internal/integration/mysql_test.go
```diff
@@ -1209,3 +1215,137 @@ func rmCreateStmt(t *schema.Table) {
...
+			},
+			Cmds: map[string]func(ts *testscript.TestScript, neg bool, args []string){
+				"apply":   t.cmdApply,
+				"exist":   t.cmdExist,
+				"synced":  t.cmdSynced,
+				"cmpshow": t.cmdCmpShow,
+			},
+		})
+	})
+}
```
a8m, line 1264: code was moved from integration_test to mysql_test.

rotemtam (approved):
Beautiful tests!

PR #486, opened 2022-01-25
sql/mysql: support marshal on update expressions in HCL

masseelch:
This PR adds support for marshalling inspected `ON UPDATE` clauses in MySQL / Maria to HCL. The marshaled HCL has the following form: 

```hcl
column "tTimestamp" {
  null      = false
  type      = timestamp(3)
  default   = sql("current_timestamp(3)")
  on_update = sql("current_timestamp(3)")
}
```

Adresses #471

PR #487, opened 2022-01-25
sql/mysql: support unmarshal on update expressions in HCL

masseelch:
This is the "reverse" part to #486, unmarshaling HCL like shown into Go Types.

### DIFF
--- a/sql/mysql/sqlspec.go
+++ b/sql/mysql/sqlspec.go
@@ -116,7 +116,7 @@ func convertColumn(spec *sqlspec.Column, _ *schema.Table) (*schema.Column, error
 	if attr, ok := spec.Attr("on_update"); ok {
 		exp, ok := attr.V.(*schemaspec.RawExpr)
 		if !ok {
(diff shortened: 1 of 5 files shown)


sql/mysql/sqlspec.go
```diff
@@ -114,11 +114,11 @@ func convertColumn(spec *sqlspec.Column, _ *schema.Table) (*schema.Column, error
 		return nil, err
 	}
 	if attr, ok := spec.Attr("on_update"); ok {
-		s, err := attr.String()
-		if err != nil {
-			return nil, err
+		exp, ok := attr.V.(*schemaspec.RawExpr)
+		if !ok {
+			return nil, fmt.Errorf("unexpected type %T for atrribute \"on_update\"", attr.V)
```
a8m, line 119:
```suggestion
			return nil, fmt.Errorf(`unexpected type %T for atrribute "on_update"`, attr.V)
```

a8m (approved):

PR #488, opened 2022-01-25
specutil/convert: fix panic on missing schema ref in table block

zeevmoney:
When creating a "valid" HCL schema, and running it through Unmarshal, everything works, but if I add another schema block, and the `table` block doesn't have reference to to the schema, it panics.

For example, before the fix this is valid and works
```HCL
schema "test" {
}

table "accounts" {
  column "name" {
    type = varchar(32)
  }
  primary_key {
    columns = [table.accounts.column.name]
  }
}
```
But this panics:

```HCL
schema "test" {
}

schema "test2" {
}

table "accounts" {
  column "name" {
    type = varchar(32)
  }
  primary_key {
    columns = [table.accounts.column.name]
  }
}
```

**This fix will also enforce the usage of `schema = ` attribute in the `table` block and make it more consistent on all cases.**

### DIFF
--- a/sql/internal/specutil/convert.go
+++ b/sql/internal/specutil/convert.go
@@ -33,7 +33,7 @@ type (
 	CheckSpecFunc         func(*schema.Check) *sqlspec.Check
 )
 
-var missingRefErr = errors.New("expected ref format of $schema.name")
+var errMissingRef = errors.New("expected ref format of $schema.name")
 
 // Realm converts the schemas and tables into a schema.Realm.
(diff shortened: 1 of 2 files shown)


sql/internal/specutil/convert.go
```diff
@@ -466,9 +468,12 @@ func FromCheck(s *schema.Check) *sqlspec.Check {
 
 // SchemaName returns the name from a ref to a schema.
 func SchemaName(ref *schemaspec.Ref) (string, error) {
+	if ref == nil {
+		return "", errMissingRef
```
a8m, line 472:
```suggestion
		return "", errors.New("unexpected nil reference")
```

PR #497, opened 2022-01-26
schema/schemaspec: add custom printer for string representation override

masseelch:
This PR adds another property to the `TypeSpec` that us used by the drivers to register supported types in the `TypeRegistry?. The `Printer` is used to override the default string representation of a `schema.Type` to support types with modifiers like Postgres' `time with time zone`. 

In a next step the postgres driver can then register a custom `Printer` that correctly represents time types with time zone modifiers with fractions like below:

`time(4) with time zone` (currently: `time with time zone(4)`)


a8m (approved):

rotemtam:
@masseelch , thanks for writing the detailed context in the PR, its super helpful and this will compound over time as we need to go back and consider why we did different things.

PR #498, opened 2022-01-26
sql/mysql: support reverse drop fk

hedwigz:


sql/mysql/migrate.go
```diff
@@ -377,7 +377,11 @@ func (s *state) alterTable(t *schema.Table, changes []schema.Change) error {
 			reverse.Comma().P("DROP FOREIGN KEY").Ident(change.F.Symbol)
 		case *schema.DropForeignKey:
 			b.P("DROP FOREIGN KEY").Ident(change.F.Symbol)
-			reversible = false
+			reverse.Comma().P("ADD")
+			if err := s.fks(reverse, change.F); err != nil {
+				errors = append(errors, err.Error())
+			}
+			reversible = true
```
a8m, line 384:
```suggestion
```

a8m: It’s the default. See variable above

sql/mysql/migrate.go
```diff
@@ -377,7 +377,11 @@ func (s *state) alterTable(t *schema.Table, changes []schema.Change) error {
 			reverse.Comma().P("DROP FOREIGN KEY").Ident(change.F.Symbol)
 		case *schema.DropForeignKey:
 			b.P("DROP FOREIGN KEY").Ident(change.F.Symbol)
-			reversible = false
+			reverse.Comma().P("ADD")
```
a8m, line 380: A nice approach to reduce duplications (also in drop index), is to:

reverse.Comma()
b, reverse = reverse, b
defer func() { b, reverse = reverse, b }()
falltrough to add version.

sql/mysql/migrate_test.go
```diff
@@ -213,6 +213,48 @@ func TestPlanChanges(t *testing.T) {
...
+							{Name: "id", Type: &schema.ColumnType{Type: &schema.IntegerType{T: "bigint"}}},
+							{Name: "user_id",
+								Type: &schema.ColumnType{
+									Type: &schema.IntegerType{T: "bigint"},
+								},
+							},
+						},
+					}
+					fk := &schema.ForeignKey{Symbol: "user_id", Table: pets, OnUpdate: schema.NoAction, OnDelete: schema.Cascade, RefTable: users}
+					pets.ForeignKeys = []*schema.ForeignKey{fk}
```
a8m, line 237: Please add columns and ref-columns to this examp

sql/mysql/migrate.go
hedwigz: `fallthrough` does not work in type switch

a8m: right :| 

kk, I have another idea, but let's merge this one atm

a8m (approved):

PR #502, opened 2022-01-26
README.md: add Go Reference (pkg.go.dev) badge

edigaryev:
To help package users navigate to the documentation faster.

Generated with https://pkg.go.dev/badge/.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 
 # Atlas -  A Database Toolkit
 
+[![Go Reference](https://pkg.go.dev/badge/ariga.io/atlas.svg)](https://pkg.go.dev/ariga.io/atlas)
+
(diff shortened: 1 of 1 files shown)


a8m:
Thanks for adding this @edigaryev. tbh, I'm not sure I want it, because I prefer users to go to atlasgo.io in most cases for documentation (I plan to add there the Go API as well), and there's a link to the godoc in case they want it. I didn't add it to ent README for the reasons.

@rotemtam, your thoughts?

rotemtam:
agree @a8m . as atlas is currently mostly used as a CLI, i prefer the first link to be to atlasgo.io. 
as you mentioned there's a link to the godoc there. 
thanks for the suggestion @edigaryev ! closing for the reason mentioned 🙏

PR #504, opened 2022-01-26
sql/sqlspec: support marshaling index key parts

a8m:
Also, added support for scanning and printing `string` and `bool` pointers. We can consider adding more types in the future if we'll see a need for this.  

testscript for PostgreSQL will be added in future PRs.


schema/schemaspec/schemahcl/hcl.go
```diff
@@ -278,6 +278,11 @@ func (s *state) writeAttr(attr *schemaspec.Attr, body *hclwrite.Body) error {
 		fnc := fmt.Sprintf("sql(%q)", v.X)
 		body.SetAttributeRaw(attr.K, hclRawTokens(fnc))
 	case *schemaspec.ListValue:
+		// Skip scanning nil slices ([]T(nil)) by default. Users that
+		// want to print empty lists, should use make([]T, 0) instead.
+		if v.V == nil {
```
rotemtam, line 283: 👍

rotemtam (approved):

PR #507, opened 2022-01-27
sql/postgres: minor fixes for postgres before release

a8m:
1. index-spec does not support custom attributes atm.
2. return more meaningful error in case of un supported postgres features in HCL (IDENTITY attribute for columns).

### DIFF
--- a/sql/internal/specutil/convert.go
+++ b/sql/internal/specutil/convert.go
@@ -465,7 +465,7 @@ func FromIndex(idx *schema.Index) (*sqlspec.Index, error) {
 func columnsOnly(idx *schema.Index) ([]*schemaspec.Ref, bool) {
 	parts := make([]*schemaspec.Ref, len(idx.Parts))
 	for i, p := range idx.Parts {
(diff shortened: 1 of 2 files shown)


rotemtam (approved):

DeedleFake:
> return more meaningful error in case of un supported postgres features in HCL (IDENTITY attribute for columns).

I just ran into this exact issue trying a dry run of my existing schema against my existing database with no changes. Is there a way around it?

a8m:
Hey @DeedleFake 👋, it's great to see you here.

I plan to address this later this week, and you're more than welcome to suggest a new syntax for the `IDENTITY` option in #517.
My current design is to allow inlining an SQL clause, but also provide a structured option for the [Identity struct](https://github.com/ariga/atlas/blob/master/sql/postgres/inspect.go#L661-L665). 

```hcl
table "users" {
    column "id" {
        type = int
        identity {
            start = 100
            increment = 2
        }
    }
}
```

PR #522, opened 2022-01-28
internal/integration/testdata/mysql: add time precision test

### LINKED ISSUE #472: Support schema apply for Mysql table with column "Timestamp/Datetime default with on update"
The `schema apply` doesn't seem to support datetime/timestamp columns that have default and on update with SQL expression:

For example, with the following schema hcl:

```hcl
schema "foo" {}

table "bar" {
  schema    = schema.foo
  column "id" {
    null = false
    type = int
  }
  column "create_time" {
    null    = false
    type    = datetime
    default = "CURRENT_TIMESTAMP(6)"
  }
  column "update_time" {
    null    = false
    type    = datetime
    default = "CURRENT_TIMESTAMP(6)"
    on_update = "CURRENT_TIMESTAMP(6)"
  }
  primary_key {
    columns = [table.bar.column.id, ]
  }
}
```

```bash
❯ atlas schema apply -d "mysql://root:secret@tcp(localhost:55237)/foo" -f ./foo.hcl
-- Planned Changes:
-- Create "bar" table
CREATE TABLE `foo`.`bar` (`id` int NOT NULL, `create_time` datetime NOT NULL DEFAULT "CURRENT_TIMESTAMP(6)", `update_time` datetime NOT NULL DEFAULT "CURRENT_TIMESTAMP(6)", PRIMARY KEY (`id`))
✔ Apply
Error: create "bar" table: Error 1067: Invalid default value for 'create_time'
```

There are two problems:
1. `CURRENT_TIMESTAMP(6)` shouldn't be quoted as a string.
2. OnUpdate part shouldn't be omitted

If this is actually supported, what's the correct way to define it in hcl? The document doesn't seem to mention this use case.

Mysql version: 5.7.35

Atlas version: 
```
atlas CLI version v0.3.1
https://github.com/ariga/atlas/releases/tag/v0.3.1
```

masseelch:
Fix #472

### DIFF
--- a/internal/integration/testdata/mysql/time-type-precision-maria.txt
+++ b/internal/integration/testdata/mysql/time-type-precision-maria.txt
@@ -1,0 +1,45 @@
+# Each test runs on a clean database.
+only maria
+
+# Apply schema "1.hcl" on fresh database.
+apply 1.hcl
+
+# Compare the result of "SHOW TABLE users" with the content of a file named '1.sql'.
+# The "cmpshow" command searches first a file named '<version>/1.sql' (version, defines
+# the database version), and in case it was found, it will use it instead.
+cmpshow foo 1.sql
+
+# Below files represent HCL and SQL. File names defined their index in
+# execution order. 1.hcl is executed first, 2.hcl executed second, etc.
+-- 1.hcl --
+schema "$db" {}
+
+table "foo" {
+  schema    = schema.$db
+  column "id" {
+    null = false
+    type = char(36)
+  }
+  column "create_time" {
+    null    = false
+    type    = timestamp(6)
+    default = sql("CURRENT_TIMESTAMP(6)")
+  }
+  column "update_time" {
+    null    = false
+    type    = datetime(6)
+    default = sql("CURRENT_TIMESTAMP(6)")
+    on_update = sql("CURRENT_TIMESTAMP(6)")
+  }
+  primary_key {
+    columns = [table.foo.column.id, ]
+  }
+}
+
+-- 1.sql --
+CREATE TABLE `foo` (
+  `id` char(36) NOT NULL,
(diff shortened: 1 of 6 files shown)


internal/integration/testdata/mysql/time-type-precision-maria.txt
```diff
@@ -0,0 +1,45 @@
+# Each test runs on a clean database.
+only maria
+
+# Apply schema "1.hcl" on fresh database.
+apply 1.hcl
+
+# Compare the result of "SHOW TABLE users" with the content of a file named '1.sql'.
+# The "cmpshow" command searches first a file named '<version>/1.sql' (version, defines
+# the database version), and in case it was found, it will use it instead.
```
a8m, line 9: I don't think this comment is needed here. I created one documented example test, but the rest can be without it.

internal/integration/testdata/mysql/time-type-precision-mysql.txt
```diff
@@ -0,0 +1,45 @@
+# Each test runs on a clean database.
+only mysql
```
a8m, line 2: you should write it as: `only 8 56 57`. Or, let's supported the negate option for `only` (I can add it), and it can be written as:
```
! only maria
```

sql/mysql/convert.go
```diff
@@ -84,7 +84,7 @@ func FormatType(t schema.Type) (string, error) {
 	case *schema.SpatialType:
 		f = strings.ToLower(t.T)
 	case *schema.TimeType:
-		f = strings.ToLower(t.T)
+		f = fmt.Sprintf("%s(%d)", strings.ToLower(t.T), t.Precision)
```
a8m, line 87: What if the precision was not defined by the user (i.e. 0)? Do you prefer to format it as `timestamp(0)` instead of `timestamp`?

a8m (approved):
Looks good!! 😃

PR #523, opened 2022-01-28
internal/integration: remove doubled whitespace from strings in tests…

masseelch:
…cript before comparing text

This change will remove all doubled whitespace except newlines from the strings to compare.

### DIFF
--- a/internal/integration/mysql_test.go
+++ b/internal/integration/mysql_test.go
@@ -1390,6 +1390,8 @@ func (t *myTest) cmdOnly(ts *testscript.TestScript, neg bool, args []string) {
 	ts.Value(keyT).(testscript.T).Skip("skip version", t.version)
 }
 
+var reSpace = regexp.MustCompile(`[^\S\r\n]+`)
+
 func (t *myTest) cmdCmpShow(ts *testscript.TestScript, _ bool, args []string) {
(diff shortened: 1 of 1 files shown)


internal/integration/mysql_test.go
```diff
@@ -1406,12 +1408,12 @@ func (t *myTest) cmdCmpShow(ts *testscript.TestScript, _ bool, args []string) {
 		// Trim the "table_options" if it was not requested explicitly.
 		stmts = append(stmts, create[:strings.LastIndexByte(create, ')')+1])
 	}
-
 	// Check if there is a file prefixed by database version (1.sql and <version>/1.sql).
 	if _, err := os.Stat(ts.MkAbs(filepath.Join(t.version, fname))); err == nil {
 		fname = filepath.Join(t.version, fname)
 	}
 	t1, t2 := strings.Join(stmts, "\n"), ts.ReadFile(fname)
+	t1, t2 = reSpace.ReplaceAllString(t1, " "), reSpace.ReplaceAllString(t2, " ")
```
a8m, line 1416: Call you elaborate why it's needed, and why `TrimSpace` is not enough?

a8m: Using `Replace` will edit also the table content (default values or expressions).

PR #524, opened 2022-01-28
Fix/cli tests

masseelch:


cmd/action/mux.go
```diff
@@ -111,7 +111,7 @@ func postgresSchema(dsn string) (string, error) {
 	if schema := url.Query().Get("search_path"); schema != "" {
 		return schema, nil
 	}
-	return "", nil
+	return "public", nil
```
yonidavidson, line 114: if you return public, there will never be a way to inspect/apply to a realm, the reason is that realm inspection expects that you pass nil schemas to as options.

When setting a default of public the CLI will always try to inspect only public in a case that no schemas were requested by user.  conclusion -> return empty string

yonidavidson: https://github.com/ariga/atlas/blob/ac7be316c108b9cecd50d3d35cefd1c5fa67722d/cmd/action/inspect.go#L73

yonidavidson: By not defining we inspect all realm (including public)

masseelch: Ah, it was meant to be public. But with multi schema this no longer is the case, right?

yonidavidson: Yes @masseelch  ! :)

yonidavidson (approved):

PR #531, opened 2022-01-30
internal/integration: initial testscript testing for pg

a8m:
Moved parts are existing code that moved from integration_test to script_test.


rotemtam (approved):
Well done!

PR #533, opened 2022-01-30
internal/integration: initial testscript testing for sqlite

a8m:


internal/integration/script_test.go
```diff
@@ -197,6 +248,16 @@ func (t *pgTest) cmdExist(ts *testscript.TestScript, neg bool, args []string) {
 	})
 }
 
+func (t *liteTest) cmdExist(ts *testscript.TestScript, neg bool, args []string) {
+	cmdExist(ts, neg, args, func(_, name string) (bool, error) {
+		var b bool
+		if err := t.db.QueryRow("SELECT COUNT(*) FROM sqlite_master WHERE `type`='table' AND `name` = ?", name).Scan(&b); err != nil {
```
yonidavidson, line 254: are you scanning a number into a bool ?
why not scan into a number and return
```go
// query result scanned into var n int
...
...
}
return n > 0 ,nil
```

yonidavidson (approved):

internal/integration/script_test.go
masseelch: I guess, because in SQL booleans are numbers. `>0 -> true`.

internal/integration/testdata/sqlite/index-desc.txt
```diff
@@ -0,0 +1,22 @@
...
+    column "rank" {
+        type = int
+    }
+    index "rank_idx" {
+        on {
+            desc   = true
+            column = table.users.column.rank
+        }
+    }
+}
```
masseelch, line 18: Shouldn't you add a test for `columns` too? (Check, there are no changes proposed by the planner).

masseelch (approved):

internal/integration/script_test.go
a8m: exactly.

internal/integration/testdata/sqlite/index-desc.txt
a8m: Next pr. I tried to make it an "initial infrastructure" for SQLite testing.

PR #539, opened 2022-01-31
Postgres driver: don't require varchar length

edigaryev:
Problem: when using Postgres driver, the `character_varying()` does not require a length attribute:

https://github.com/ariga/atlas/blob/ce25dd8f5ae3f594644e6bfb3414b1532bfa3d0b/sql/postgres/sqlspec.go#L338

...yet `varchar()` does:

https://github.com/ariga/atlas/blob/ce25dd8f5ae3f594644e6bfb3414b1532bfa3d0b/sql/postgres/sqlspec.go#L337

The [official PostgreSQL documentation](https://www.postgresql.org/docs/9.3/datatype-character.html) treats these two identically:

>The notations `varchar(n)` and `char(n)` are aliases for character `varying(n)` and `character(n)`, respectively.

Solution: allow `varchar()` to be specified without the length, similarly to `character_varying()`.

### DIFF
--- a/sql/postgres/sqlspec_test.go
+++ b/sql/postgres/sqlspec_test.go
@@ -29,6 +29,12 @@ table "table" {
 	column "account_name" {
 		type = varchar(32)
 	}
+	column "varchar_length_is_not_required" {
+		type = varchar()
+	}
+	column "character_varying_length_is_not_required" {
+		type = character_varying()
+	}
 	column "tags" {
 		type = hstore
 	}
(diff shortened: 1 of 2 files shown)


sql/postgres/sqlspec_test.go
```diff
@@ -29,6 +29,12 @@ table "table" {
 	column "account_name" {
 		type = varchar(32)
 	}
+	column "varchar_length_is_not_required" {
+		type = varchar()
```
rotemtam, line 33:
```suggestion
		type = varchar
```

this will work but looks nicer as it will be written in the database

sql/postgres/sqlspec_test.go
```diff
@@ -29,6 +29,12 @@ table "table" {
...
+	}
+	column "character_varying_length_is_not_required" {
+		type = character_varying()
```
rotemtam, line 36:
```suggestion
		type = character_varying
```
same

rotemtam (changes requested):
hey @edigaryev !
thanks for working on this, see small comment.

rotemtam (approved):
Thanks for your contribution @edigaryev !

a8m:
Thanks for the help @edigaryev ❤️

PR #541, opened 2022-02-01
doc/md: add dsn and ssl mode documentation

### LINKED ISSUE #518: Add option to connect to database without SSL (fix "SSL is not enabled on the server" error)
**Type**

Bug

**System**

OS: Mac OSX 11.6.3
DB: Postgres 12.8

**Details**

Installation went smoothly, but when I tried to inspect my existing local DB schema, I got an error

```
> atlas schema inspect --dsn "postgres://eric@localhost:5432/my_db"
Error: postgres: scanning system variables: pq: SSL is not enabled on the server
```

A quick search led me to [this SO answer](https://stackoverflow.com/a/30787336/3991555).

It appears that Atlas is attempting to connect with the raw DSN, without specifying an SSL mode (it appears the SSL is enabled by default):

https://github.com/ariga/atlas/blob/ee1ec7abca136edea8428a7588b0d33b104868c7/cmd/action/provider.go#L35

Perhaps this is necessary in some cases but for local DB inspection, probably not. The tests appear to get around this by explicitly disabling SSL:

https://github.com/ariga/atlas/blob/ee1ec7abca136edea8428a7588b0d33b104868c7/internal/integration/postgres_test.go#L41

I verified this fix by running Atlas with the `?sslmode=disable` query param added to my DB url, and it worked

```
> atlas schema inspect --dsn "postgres://eric@localhost:5432/my_db?sslmode=disable"

🎉 prints a schema to terminal
```

I am not well versed in Go to propose a specific solution but I think the most flexible option would be to add flag option for SSL mode (and perhaps default to `false` for inspection, since I would imagine inspecting a local DB is the most common initial step)

ericyd:
Closes #518 

### Updates

- Add `?sslmode=disable` query param to example Postgres URLs
- Add links to Postgres docs for SSL configuration

I am 100% open to changing any of this language, or examples. I tried to keep it fairly minimal but still as easy as possible for new adopters using Postgres


doc/md/getting-started/02-inspecting.md
```diff
@@ -161,3 +161,7 @@ read the [SQL Syntax](../ddl/sql.md) documentation.
 
 In the next section, we will see how we can modify our database's schema by 
 applying a modified Atlas HCL file using the Atlas CLI.
+
+## SSL Mode
```
rotemtam, line 165: The getting started tutorial uses MySQL for the demo. I think it's best to add a new page in the CLI section that documents DSNs and explains a bit about them and add this in a section about postgres. 

Then we can put a [Docusaurus Admonition](https://docusaurus.io/docs/next/markdown-features/admonitions) in this page that references there (e.g. "If you're using PostgreSQL, here's documentation about building the connection string" or something). 

Can be done in a separate PR.

rotemtam (changes requested):
see comment, i'd move paragraph to a separate page in a new PR.

doc/md/getting-started/02-inspecting.md
ericyd: This makes sense to me. 

Could you clarify where you imagine this living? I am interpreting your instructions as
1. `git mv doc/md/cli.md doc/md/cli/reference.md`
2. `touch doc/md/cli/dsn.md`
3. Move this blurb into `doc/md/cli/dsn.md` and add some details about DSNs in more detail

before I make changes want to make sure I'm on track

rotemtam: sounds good, notice that doc/md/cli.md is a generated file that is generated by [this script](https://github.com/ariga/atlas/blob/master/internal/typedoc/main.go) so when you move that be sure to update that. 

I added instructions run to test the doc site locally [here](https://atlasgo.io/contributing#contributing-documentation) , that would be the best way to see that everything looks good before committing

.github/workflows/ci.yml
```diff
@@ -30,6 +30,9 @@ jobs:
             ${{ runner.os }}-go-
       - name: run "go generate ./..."
         run: go generate ./...
+      - name: run "go generate internal/typedoc"
+        working-directory: internal/typedoc 
+        run: go generate ./...
```
ericyd, line 35: I stumbled upon this by mistake but it appeared to generate a change which wasn't generated by default with `go generate ./...` (see failing CI run [here](https://github.com/ariga/atlas/runs/5046361816?check_suite_focus=true)). I am happy to remove this if it's not necessary

ericyd:
I'm definitely not a DSN expert 😄 so please request any/all changes to the phrasing I used

rotemtam (approved):
@ericyd thanks for your contribution!
Looks great, I'll make some additions to this on a separate PR.

PR #544, opened 2022-02-02
fix/typos

int-2147483647:
Fixed typos found by running: https://goreportcard.com/report/ariga.io/atlas


rotemtam (approved):
thanks for the contribution @int-2147483647 !

PR #549, opened 2022-02-03
sql/mysql: add support for float and decimal unsigned attribute

### LINKED ISSUE #545: MySQL - Atlas fails to inpect schema when a column is specified as an unsigned float
It seems that when you create a column in mysql with an unsigned float, Atlas cannot parse its precision in sql/mysql/convert.go and throws a lovely error message like below:

Error: mysql: parse precision "unsigned"

masseelch:
Fix #545


a8m (approved):
Beast! 🥳

sql/mysql/convert.go
```diff
@@ -60,6 +63,9 @@ func FormatType(t schema.Type) (string, error) {
 		if f == TypeFloat && t.Precision > 24 || f == TypeReal {
 			f = TypeDouble
 		}
+		if t.Unsigned {
+			f += " unsigned"
```
yonidavidson, line 67: don't you have another way to add words without using "space" as part of the string?

PR #551, opened 2022-02-07
sql/postgres: same inspect, but faster

a8m:


doc/md/ddl/sql_types.md
```diff
@@ -187,6 +187,7 @@ slug: /ddl/sql-types
             <td>decimal</td>
             <td>
                 <ul>
+                        <li>unsigned (bool)</li>
```
a8m, line 190: Not part of my PR, but CI is falling.

masseelch: Shouldn't we fix it first then and rebase this PR onto that?

a8m: I don't mind landing this PR with this failure. It's not a test issue.

a8m: Fixed.

masseelch (approved):

PR #554, opened 2022-02-08
sql/internal/specutil: support local column referencing

a8m:


doc/md/ddl/sql.md
```diff
@@ -39,18 +37,18 @@ table "users" {
...
   primary_key {
     columns = [
-        table.users.column.id
+      column.id
     ]
   }
   index "idx_name" {
     columns = [
-      table.users.column.name
+      column.name
```
rotemtam, line 45: 😍

sql/internal/specutil/convert_test.go
```diff
@@ -67,10 +67,10 @@ func TestFromForeignKey(t *testing.T) {
 	require.EqualValues(t, &sqlspec.ForeignKey{
 		Symbol: "fk",
 		Columns: []*schemaspec.Ref{
-			{V: "$table.users.$column.parent_id"},
+			{V: "$column.parent_id"},
 		},
 		RefColumns: []*schemaspec.Ref{
-			{V: "$table.users.$column.id"},
+			{V: "$column.id"},
```
rotemtam, line 73: can we keep the existing tests and add new ones for the new syntax?

sql/mysql/sqlspec_test.go
```diff
@@ -566,7 +566,7 @@ func TestMarshalSpec_IndexParts(t *testing.T) {
   index "idx" {
     on {
       desc   = true
-      column = table.users.column.name
+      column = column.name
```
rotemtam, line 569: ditto

rotemtam (approved):
i love this change, only request is keep the existing test so that `table.x.column.y` to verify we keep support for it?

a8m:
> i love this change, only request is keep the existing test so that `table.x.column.y` to verify we keep support for it?

Integration tests still use the old form

PR #556, opened 2022-02-09
sql/migrate: add migration directory

masseelch:
First PR in a series to support versioned migration files for atlas.

The `FS` interface allows for different storage solutions to be used with this feature. It can range from AWS over GitHub repos or HTTP APIs. Exmaple for a local (dev) implementation is given.


rotemtam (commented):
Why not use fs.FS?

https://pkg.go.dev/io/fs

masseelch:
> Why not use fs.FS?
> 
> https://pkg.go.dev/io/fs

fs.FS has no methods for manipulation.

sql/migrate/migrate.go
```diff
@@ -127,303 +124,144 @@ func Schema(s *schema.Schema) StateReader {
...
-	dirFS struct {
-		fs.FS
-		dir string
+	// Printer wraps the methods for naming and dumping a migration plan to one or more files.
+	Printer interface {
+		// Print prints the given Plan.
+		// The first return argument contains a slice of filenames.
+		// The second return argument is meant to hold the contents for each filename.
+		// The length of the filenames-slice and contents-slice must be equal.
+		Print(*Plan) ([]string, [][]byte, error)
```
rotemtam, line 148: Any particular reason not to use a map?

sql/migrate/migrate.go
```diff
@@ -127,303 +124,144 @@ func Schema(s *schema.Schema) StateReader {
...
-		}
-		if !fi.IsDir() {
-			return fmt.Errorf("sql/migrate: %q not a dir", path)
-		}
-		d.fs = &dirFS{FS: os.DirFS(path), dir: path}
-		return nil
-	}
+// Read reads the contents of a file by name.
+func (fs *LocalFS) Read(name string) ([]byte, error) {
+	return ioutil.ReadFile(filepath.Join(fs.dir, name))
```
rotemtam, line 182: I think since 1.16 can use os package instead ?

sql/migrate/migrate.go
```diff
@@ -127,303 +124,144 @@ func Schema(s *schema.Schema) StateReader {
...
-//			"{{timestamp}}{{.Name}}.up.sql", "{{range $c := .Changes}}{{println $c.Cmd}}{{end}}",
-//			"{{timestamp}}{{.Name}}.down.sql", "{{range $c := .Changes}}{{println $c.Reverse}}{{end}}",
-//		),
-//	)
-//
-func DirTemplates(nameFileTmpl ...string) DirOption {
-	return func(d *Dir) error {
-		if n := len(nameFileTmpl); n == 0 || n%2 == 1 {
-			return fmt.Errorf("odd or zero argument count")
+// GoMigratePrinter implements Printer for a golang-migrate/migrate compatible migration files.
```
rotemtam, line 207: Drop the "a" in the comment

rotemtam (commented):
Looks good, see comments

rotemtam (approved):
LGTM but i'd wait for @a8m to green light since he worked on this area.

cmd/action/apply.go
```diff
@@ -18,12 +18,12 @@ import (
...
-		Web    bool
-		Addr   string
-		DryRun bool
-		Schema []string
+		DSN         string
+		File        string
+		Web         bool
+		Addr        string
+		DryRun      bool
+		Schema      []string
```
yonidavidson, line 26: is our formatting different?

masseelch: No. One of the users added the auto-approve flag and apparently did not run gofmt. My IDE does this on every commit. And this is what it spit out.

masseelch: Weird though, that the linter does not catch it :thinking:

sql/migrate/migrate.go
```diff
@@ -127,303 +123,140 @@ func Schema(s *schema.Schema) StateReader {
...
+	FS interface {
+		// List all present migration files. Usually *.sql files.
+		List() ([]string, error)
 
-	// DirOption allows configuring the Dir
-	// using functional options.
-	DirOption func(*Dir) error
-)
+		// Read the contents of a file by name.
+		Read(name string) ([]byte, error)
```
a8m, line 132: Can we use the standard [io.ReasDirFS](https://pkg.go.dev/io/fs#ReadDirFS) here?

sql/migrate/migrate.go
```diff
@@ -127,303 +123,140 @@ func Schema(s *schema.Schema) StateReader {
...
-	// dirFS wraps the os.DirFS with additional writing capabilities.
-	dirFS struct {
-		fs.FS
-		dir string
+	// Printer wraps the methods for naming and dumping a migration plan to one or more files.
+	Printer interface {
+		// Print prints the given Plan.
+		// The returned map contains the contents for each file to dump.
+		Print(*Plan) (map[string][]byte, error)
 	}
```
a8m, line 146: I would expect a Printer to print, Writer to write, but it looks like this one "formats" the output. Maybe use `Formater` instead?

Also, can we use `[]File` instead of a map? Iterating over a map can be annoying if you care about the order, and the `File` can be easily extended in the future.

sql/migrate/migrate.go
```diff
@@ -127,303 +123,140 @@ func Schema(s *schema.Schema) StateReader {
...
+		fs:  fs,
+		pr:  pr,
+	}
 }
 
-// RemoveFile implements the FileWriter interface.
-func (d *dirFS) RemoveFile(name string) error {
-	return os.Remove(filepath.Join(d.dir, name))
+// LocalFS implements FS for a local path.
+type LocalFS struct {
```
a8m, line 167: Why not a Dir?

sql/migrate/migrate.go
```diff
@@ -127,303 +123,140 @@ func Schema(s *schema.Schema) StateReader {
...
-{{- end }}`)),
+	if !fi.IsDir() {
+		return nil, fmt.Errorf("sql/migrate: %q is not a dir", path)
 	}
-)
+	return &LocalFS{dir: path, glob: glob}, nil
+}
+
+// GoMigratePrinter implements Printer for golang-migrate/migrate compatible migration files.
+type GoMigratePrinter struct{}
```
a8m, line 205: tbh, I'm not sure if having it exported is a good idea (let's discuss it f2f).

a8m (commented):
Let's discuss this f2f.

sql/migrate/migrate.go
```diff
@@ -127,303 +122,103 @@ func Schema(s *schema.Schema) StateReader {
...
-	// Dir represents a versioned migration directory.
-	Dir struct {
-		fs        fs.FS
-		conn      Driver
-		pattern   string
-		templates []struct{ N, T *template.Template }
+	// Dir describes the methods needed to work with the Planner.
+	Dir interface {
+		// Open opens the named file.
+		Open(name string) (io.ReadWriteCloser, error)
```
a8m, line 128: List/DirFiles is missing.

a8m (approved):

PR #562, opened 2022-02-10
cmd/action: fix grammar error in apply description

masseelch:


yonidavidson (approved):

rotemtam (approved):
probably need to regen docs?

PR #563, opened 2022-02-10
sql/mysql: support index types

a8m:
Also, renamed `index.using` to `index.type`.

### DIFF
--- a/sql/mysql/sqlspec.go
+++ b/sql/mysql/sqlspec.go
@@ -66,7 +66,7 @@ func MarshalSpec(v interface{}, marshaler schemaspec.Marshaler) ([]byte, error) 
 var (
 	hclState = schemahcl.New(
 		schemahcl.WithTypes(TypeRegistry.Specs()),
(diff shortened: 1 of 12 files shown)


rotemtam (approved):

sql/mysql/sqlspec.go
```diff
@@ -165,6 +185,17 @@ func tableSpec(t *schema.Table) (*sqlspec.Table, error) {
 	return ts, nil
 }
 
+func indexSpec(idx *schema.Index) (*sqlspec.Index, error) {
+	spec, err := specutil.FromIndex(idx)
+	if err != nil {
+		return nil, err
+	}
+	if i := (IndexType{}); sqlx.Has(idx.Attrs, &i) {
```
yonidavidson, line 193: do you need the parentheses?

yonidavidson (approved):

sql/mysql/sqlspec.go
a8m: Yes.

PR #566, opened 2022-02-11
sql/migrate: add template formatter

masseelch:
This PR adds `TemplateFormatter` which implements `Formatter` and uses `text/template.Template`s to define how migrations files are written.

A default formatter is given, that is compatible with [golang-migrate/migrate](https://github.com/golang-migrate/migrate).

### DIFF
--- a/sql/migrate/migrate.go
+++ b/sql/migrate/migrate.go
@@ -5,12 +5,16 @@
 package migrate
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
+	"text/template"
+	"time"
 
 	"ariga.io/atlas/sql/schema"
 )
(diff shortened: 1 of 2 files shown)


sql/migrate/migrate.go
```diff
@@ -222,3 +203,122 @@ func (p *Planner) WritePlan(plan *Plan) error {
...
+			},
+		},
+	}
+)
+
+// NewTemplateFormatter creates a new Formatter working with the given templates.
+//
+//	migrate.NewTemplateFormatter(
+//		template.Must(template.New("").Parse("{{now.Unix}}{{.Name}}.sql")),					// name template
+//		template.Must(template.New("").Parse("{{range .Changes}}{{println .Cmd}}{{end}}")),	// content template
```
rotemtam, line 268: comment seems too indented on L267

rotemtam (approved):

sql/migrate/migrate.go
```diff
@@ -222,3 +203,122 @@ func (p *Planner) WritePlan(plan *Plan) error {
...
 	return nil
 }
+
+// LocalDir implements Dir for a local path.
+type LocalDir struct {
+	dir string
+}
+
+// Open implements Dir.Open.
+func (dir *LocalDir) Open(name string) (File, error) {
```
a8m, line 213: use short var.
```suggestion
func (d *LocalDir) Open(name string) (File, error) {
```

sql/migrate/migrate.go
```diff
@@ -222,3 +203,122 @@ func (p *Planner) WritePlan(plan *Plan) error {
...
+	t := new(TemplateFormatter)
+	for i := 0; i < len(templates); i += 2 {
+		t.templates = append(t.templates, struct{ N, C *template.Template }{templates[i], templates[i+1]})
+	}
+	return t, nil
+}
+
+// Format implements the Formatter interface.
+func (t *TemplateFormatter) Format(plan *Plan) ([]File, error) {
+	var fs []File
```
a8m, line 284:
```suggestion
	fs := make([]File, 0, len(t.templates))
```

a8m (approved):
Looks good. The name `TemplateFormatter` feels a big long for me, but I still like it 😃

PR #587, opened 2022-02-15
sql/mysql: TiDB inspector for MySQL driver

hedwigz:
In this PR I begin the work of integrating TiDB with Atlas.  
I began with solving the issues of the first integration test (`TestMySQL_AddDropTable`) which are well described [here](https://github.com/pingcap/tidb/issues/32411) and [here](https://github.com/pingcap/tidb/issues/28918).  
Some other tests were also solved.  
in `tidb.go` I implement a decorator for `mysql` inspector which applies patches after MySQL inspect.

### DIFF
--- a/internal/integration/docker-compose.yaml
+++ b/internal/integration/docker-compose.yaml
@@ -123,8 +123,14 @@ services:
       - 4308:3306
 
 # Default DB test, No Password
-  tidb:
+  tidb5:
     platform: linux/amd64
     image: pingcap/tidb:v5.4.0
     ports:
       - 4309:4000
+
+  tidb_latest:
+    platform: linux/amd64
+    image: pingcap/tidb:latest
+    ports:
+      - 4310:4000
--- a/internal/integration/docker-compose.yaml
+++ b/internal/integration/docker-compose.yaml
@@ -121,3 +121,12 @@ services:
       test: mysqladmin ping -ppass
     ports:
       - 4308:3306
+
+# Default DB test, No Password
+  tidb:
+    platform: linux/amd64
+    image: pingcap/tidb:latest
+    healthcheck:
+      test: mysqladmin ping -ppass
+    ports:
+      - 4309:4000
--- a/internal/integration/docker-compose.yaml
+++ b/internal/integration/docker-compose.yaml
@@ -125,7 +125,7 @@ services:
 # Default DB test, No Password
   tidb:
     platform: linux/amd64
-    image: pingcap/tidb:latest
+    image: pingcap/tidb:v5.4.0
     healthcheck:
       test: mysqladmin ping -ppass
     ports:
--- a/internal/integration/docker-compose.yaml
+++ b/internal/integration/docker-compose.yaml
@@ -126,7 +126,5 @@ services:
   tidb:
     platform: linux/amd64
     image: pingcap/tidb:v5.4.0
-    healthcheck:
-      test: mysqladmin ping -ppass
     ports:
       - 4309:4000
--- a/internal/integration/mysql_test.go
+++ b/internal/integration/mysql_test.go
@@ -36,7 +36,7 @@ var myTests struct {
 func myRun(t *testing.T, fn func(*myTest)) {
 	myTests.Do(func() {
 		myTests.drivers = make(map[string]*myTest)
-		for version, port := range map[string]int{"56": 3306, "57": 3307, "8": 3308, "Maria107": 4306, "Maria102": 4307, "Maria103": 4308} {
+		for version, port := range map[string]int{"TiDB": 4309} {
 			password := ":pass"
 			if version == "TiDB" {
 				password = ""
--- a/internal/integration/mysql_test.go
+++ b/internal/integration/mysql_test.go
@@ -1279,6 +1279,9 @@ func (t *myTest) defaultAttrs() []schema.Attr {
 		collation = "latin1_swedish_ci"
 	)
 	switch {
+	case t.version == "TiDB":
+		charset = "utf8mb4"
+		collation = "utf8mb4_bin"
 	case t.version == "8":
 		charset = "utf8mb4"
 		collation = "utf8mb4_0900_ai_ci"
--- a/internal/integration/mysql_test.go
+++ b/internal/integration/mysql_test.go
@@ -36,7 +36,7 @@ var myTests struct {
 func myRun(t *testing.T, fn func(*myTest)) {
 	myTests.Do(func() {
 		myTests.drivers = make(map[string]*myTest)
-		for version, port := range map[string]int{"TiDB": 4309} {
+		for version, port := range map[string]int{"56": 3306, "57": 3307, "8": 3308, "Maria107": 4306, "Maria102": 4307, "Maria103": 4308} {
 			password := ":pass"
 			if version == "TiDB" {
 				password = ""
--- a/internal/integration/mysql_test.go
+++ b/internal/integration/mysql_test.go
@@ -36,9 +36,9 @@ var myTests struct {
 func myRun(t *testing.T, fn func(*myTest)) {
 	myTests.Do(func() {
 		myTests.drivers = make(map[string]*myTest)
-		for version, port := range map[string]int{"56": 3306, "57": 3307, "8": 3308, "Maria107": 4306, "Maria102": 4307, "Maria103": 4308, "tiDB": 4309} {
+		for version, port := range map[string]int{"56": 3306, "57": 3307, "8": 3308, "Maria107": 4306, "Maria102": 4307, "Maria103": 4308} {
 			password := ":pass"
-			if version == "tiDB" {
+			if version == "TiDB" {
 				password = ""
 			}
 			db, err := sql.Open("mysql", fmt.Sprintf("root%s@tcp(localhost:%d)/test?parseTime=True", password, port))
--- a/internal/integration/mysql_test.go
+++ b/internal/integration/mysql_test.go
@@ -36,8 +36,12 @@ var myTests struct {
 func myRun(t *testing.T, fn func(*myTest)) {
 	myTests.Do(func() {
 		myTests.drivers = make(map[string]*myTest)
-		for version, port := range map[string]int{"56": 3306, "57": 3307, "8": 3308, "Maria107": 4306, "Maria102": 4307, "Maria103": 4308} {
-			db, err := sql.Open("mysql", fmt.Sprintf("root:pass@tcp(localhost:%d)/test?parseTime=True", port))
+		for version, port := range map[string]int{"56": 3306, "57": 3307, "8": 3308, "Maria107": 4306, "Maria102": 4307, "Maria103": 4308, "tiDB": 4309} {
+			password := ":pass"
+			if version == "tiDB" {
+				password = ""
+			}
+			db, err := sql.Open("mysql", fmt.Sprintf("root%s@tcp(localhost:%d)/test?parseTime=True", password, port))
 			require.NoError(t, err)
 			drv, err := mysql.Open(db)
 			require.NoError(t, err)
--- a/sql/mysql/driver.go
+++ b/sql/mysql/driver.go
@@ -99,6 +99,11 @@ func (d *conn) mariadb() bool {
 	return strings.Index(d.version, "MariaDB") > 0
 }
 
+// tidb reports if the Driver is connected to TiDC database.
+func (d *conn) tidb() bool {
+	return strings.Index(d.version, "TiDB") > 0
+}
+
 // compareV returns an integer comparing two versions according to
 // semantic version precedence.
 func (d *conn) compareV(w string) int {
--- a/sql/mysql/driver.go
+++ b/sql/mysql/driver.go
@@ -192,7 +192,8 @@ const (
 	IndexTypeHash     = "HASH"
 	IndexTypeFullText = "FULLTEXT"
 	IndexTypeSpatial  = "SPATIAL"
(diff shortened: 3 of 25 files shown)


internal/integration/docker-compose.yaml
```diff
@@ -120,4 +120,13 @@ services:
...
       test: mysqladmin ping -ppass
     ports:
-      - 4308:3306
\ No newline at end of file
+      - 4308:3306
+
+# Default DB test, No Password
+  tidb:
+    platform: linux/amd64
+    image: pingcap/tidb:v5.4.0
```
a8m, line 128: can't we use latest?

internal/integration/docker-compose.yaml
```diff
@@ -120,4 +120,13 @@ services:
...
+    healthcheck:
+      test: mysqladmin ping -ppass
```
a8m, line 130: are you sure the image contains `mysqladmin`?

sql/mysql/driver.go
```diff
@@ -46,6 +46,14 @@ func Open(db schema.ExecQuerier) (*Driver, error) {
...
 		return nil, fmt.Errorf("mysql: scan system variables: %w", err)
 	}
+	if c.tidb() {
+		return &Driver{
+			conn:        c,
+			Differ:      &sqlx.Diff{DiffDriver: &diff{c}},
+			Inspector:   &tinspect{&inspect{c}},
+			PlanApplier: &planApply{c},
+		}, nil
+	}
```
a8m, line 56: Override only the inspector.

sql/mysql/driver.go
```diff
@@ -99,6 +107,11 @@ func (d *conn) mariadb() bool {
 	return strings.Index(d.version, "MariaDB") > 0
 }
 
+// tidb reports if the Driver is connected to TiDC database.
```
a8m, line 110:
```suggestion
// tidb reports if the Driver is connected to a TiDC database.
```

sql/mysql/inspect.go
```diff
@@ -465,14 +465,14 @@ func (i *inspect) extraAttr(t *schema.Table, c *schema.Column, extra string) err
 // the 'SHOW CREATE' command.
 func (i *inspect) showCreate(ctx context.Context, s *schema.Schema) error {
 	for _, t := range s.Tables {
-		s, ok := popShow(t)
+		st, ok := popShow(t)
```
a8m, line 468: ?

sql/mysql/inspect.go
```diff
@@ -465,14 +465,14 @@ func (i *inspect) extraAttr(t *schema.Table, c *schema.Column, extra string) err
...
 		if !ok {
 			continue
 		}
 		if err := i.createStmt(ctx, t); err != nil {
 			return err
 		}
-		if err := i.setAutoInc(s, t); err != nil {
+		if err := i.setAutoInc(st, t); err != nil {
```
a8m, line 475: ?

sql/mysql/tidb.go
```diff
@@ -0,0 +1,123 @@
+package mysql
```
a8m, line 1: license header, please.

sql/mysql/tidb.go
```diff
@@ -0,0 +1,123 @@
...
+	"fmt"
+	"regexp"
+	"strings"
+
+	"ariga.io/atlas/sql/internal/sqlx"
+	"ariga.io/atlas/sql/schema"
+)
+
+type tinspect struct {
+	*inspect
```
a8m, line 14: You can also embed non-pointer.

sql/mysql/tidb.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (i *tinspect) InspectRealm(ctx context.Context, opts *schema.InspectRealmOption) (*schema.Realm, error) {
+	r, err := i.inspect.InspectRealm(ctx, opts)
+	if err != nil {
+		return nil, err
+	}
+	for _, s := range r.Schemas {
+		_, err := i.patchSchema(ctx, s)
+		if err != nil {
```
a8m, line 32:
```suggestion
		if _, err := i.patchSchema(ctx, s); err != nil {
```

sql/mysql/tidb.go
```diff
@@ -0,0 +1,123 @@
...
+			return nil, err
+		}
+	}
+	return r, nil
+}
+
+func (i *tinspect) patchSchema(ctx context.Context, s *schema.Schema) (*schema.Schema, error) {
+	for _, t := range s.Tables {
+		if err := i.createStmt(ctx, t); err != nil {
```
a8m, line 41: Maybe check before if the `CreateStmt` exists in the attributes?

sql/mysql/tidb.go
```diff
@@ -0,0 +1,123 @@
...
+			return nil, err
+		}
+		if err := i.setCollation(t); err != nil {
+			return nil, err
+		}
+		if err := i.setFKs(s, t); err != nil {
+			return nil, err
+		}
```
a8m, line 49: Maybe pass here the `CreateStmt` and avoid the checks `setCollation` and `setFKs`.

sql/mysql/tidb.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	return s, nil
+}
+
+// e.g CONSTRAINT "" FOREIGN KEY ("foo_id") REFERENCES "foo" ("id")
+var reFK = regexp.MustCompile("(?i)CONSTRAINT\\s+[\"`]*(\\w+)[\"`]*\\s+FOREIGN\\s+KEY\\s*\\(([,\"` \\w]+)\\)\\s+REFERENCES\\s+[\"`]*(\\w+)[\"`]*\\s*\\(([,\"` \\w]+)\\)")
```
a8m, line 55: This is the regexp from SQLite? I think it can be simplified a lot because TiDB/MySQL "SHOW CREATE .." statement returns a formatted (and consistent) output.

sql/mysql/tidb.go
```diff
@@ -0,0 +1,123 @@
...
+func columns(schema *schema.Schema, s string) []string {
+	names := strings.Split(s, ",")
+	for i := range names {
+		names[i] = strings.Trim(strings.TrimSpace(names[i]), "`\"")
+	}
+	return names
+}
+
+// e.g CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+var reColl = regexp.MustCompile(`(?i)CHARSET\s*=\s*(\w+)\s*COLLATE\s*=\s*(\w+)`)
```
a8m, line 108: ditto

sql/mysql/tidb.go
```diff
@@ -0,0 +1,123 @@
...
+	var c CreateStmt
+	if !sqlx.Has(t.Attrs, &c) {
+		return fmt.Errorf("missing CREATE TABLE statment in attribuets for %q", t.Name)
+	}
+	matches := reColl.FindStringSubmatch(c.S)
+	if len(matches) != 3 {
+		return fmt.Errorf("missing COLLATE and/or CHARSET information on CREATE TABLE statment for %q", t.Name)
+	}
+	schema.ReplaceOrAppend(&t.Attrs, &schema.Charset{V: matches[1]})
+	schema.ReplaceOrAppend(&t.Attrs, &schema.Collation{V: matches[2]})
```
a8m, line 121:
```suggestion
	t.SetCharset(matches[1])
	t.SetCollation(matches[2])
```

sql/schema/dsl.go
```diff
@@ -728,6 +728,12 @@ func replaceOrAppend(attrs *[]Attr, v Attr) {
 	*attrs = append(*attrs, v)
 }
 
+// ReplaceOrAppend searches an attribute of the same type as v in
+// the list and replaces it. Otherwise, v is appended to the list.
+func ReplaceOrAppend(attrs *[]Attr, v Attr) {
+	replaceOrAppend(attrs, v)
+}
```
a8m, line 735: not really needed. see comment above.

sql/mysql/inspect.go
masseelch: I don't like shadowing too

sql/mysql/tidb.go
masseelch: So use backtics to mark the regex string. Gets rid of all the escaped quotes

internal/integration/docker-compose.yaml
hedwigz: We can but then if they change something in a later version (say v6.0.0) and our tests will break that's an unwanted behaviour.  
Maybe let's do both `tidb5:v5.4.0` and also a "rolling" `tidb_latest:latest` and when a new version breaks our tests we can add fixes to that version and then add `tidb<fail_version>:<fail_version>`

internal/integration/docker-compose.yaml
hedwigz: there isn't - removed

sql/mysql/driver.go
hedwigz: will do the others as needed in the next PRs

sql/mysql/tidb.go
```diff
@@ -0,0 +1,122 @@
...
+		names[i] = strings.Trim(strings.TrimSpace(names[i]), "`\"")
+	}
+	return names
+}
+
+// e.g CHARSET=utf8mb4 COLLATE=utf8mb4_bin
+var reColl = regexp.MustCompile(`(?i)CHARSET\s*=\s*(\w+)\s*COLLATE\s*=\s*(\w+)`)
+
+// setCollation extracts the updated Collation from CREATE TABLE statement.
+func (i *tinspect) setCollation(t *schema.Table) error {
```
masseelch, line 110: Rename to `Collate`

sql/mysql/tidb.go
hedwigz: simplified

sql/mysql/tidb.go
hedwigz: suggestions? seems pretty short and basic IMO

sql/mysql/tidb.go
hedwigz: :+1:

internal/integration/mysql_test.go
```diff
@@ -36,8 +36,12 @@ var myTests struct {
 func myRun(t *testing.T, fn func(*myTest)) {
 	myTests.Do(func() {
 		myTests.drivers = make(map[string]*myTest)
-		for version, port := range map[string]int{"56": 3306, "57": 3307, "8": 3308, "Maria107": 4306, "Maria102": 4307, "Maria103": 4308} {
-			db, err := sql.Open("mysql", fmt.Sprintf("root:pass@tcp(localhost:%d)/test?parseTime=True", port))
+		for version, port := range map[string]int{"TiDB": 4309} {
```
masseelch, line 39: Don't forget to include the other drivers again

sql/mysql/driver.go
```diff
@@ -99,6 +107,11 @@ func (d *conn) mariadb() bool {
 	return strings.Index(d.version, "MariaDB") > 0
 }
 
+// tidb reports if the Driver is connected to a TiDC database.
```
masseelch, line 110:
```suggestion
// tidb reports if the Driver is connected to a TiDB database.
```

sql/mysql/inspect.go
```diff
@@ -648,7 +648,7 @@ SELECT
 	t1.CREATE_OPTIONS
 FROM
 	INFORMATION_SCHEMA.TABLES AS t1
-	JOIN INFORMATION_SCHEMA.COLLATIONS AS t2
+	LEFT JOIN INFORMATION_SCHEMA.COLLATIONS AS t2
```
masseelch, line 651: Why is a left join needed?

internal/integration/mysql_test.go
hedwigz: Whoops, thx

sql/mysql/inspect.go
hedwigz: Because in TiDB you might not have collate information (see point 3 in my issue [here](https://github.com/pingcap/tidb/issues/32411))

internal/integration/docker-compose.yaml
a8m: >We can but then if they change something in a later version (say v6.0.0) and our tests will break that's an unwanted behaviour.

As I see it, that's a wanted behavior and helped me catch bugs in new versions of MySQL, PostgreSQL, and actually SQLite. Does pingcap maintain multiple versions or recommend working with latest?

a8m: I see now that you added latest. nvm

sql/mysql/tidb.go
```diff
@@ -0,0 +1,129 @@
...
+		if len(m) != 5 {
+			return fmt.Errorf("unexpected number of matches for a table constraint: %q", m)
+		}
+		ctName, clmns, refTableName, refClmns := m[1], m[2], m[3], m[4]
+		fk := &schema.ForeignKey{
+			Symbol: ctName,
+			Table:  t,
+			// There is no support in TiDB for FKs so inherently there are no actions on update/delete.
+			OnUpdate: schema.NoAction,
+			OnDelete: schema.NoAction,
```
a8m, line 78: Are you sure it won't create a diff in case the referential action is not "NO ACTION"?

a8m (approved):
Looks good, minor comment/question.

sql/mysql/tidb.go
hedwigz: no, but if it does I'll find out soon enough with the integration tests.  
Why would there be any action other than that? they don't support this constraint anyway :)

PR #589, opened 2022-02-16
sql/mysql: rename collation attribute to collate

a8m:
`collate` is shorter and also used by MySQL/PostgreSQL

### DIFF
--- a/sql/mysql/sqlspec.go
+++ b/sql/mysql/sqlspec.go
@@ -176,7 +176,7 @@ func schemaSpec(s *schema.Schema) (*sqlspec.Schema, []*sqlspec.Table, error) {
 		sc.Extra.Attrs = append(sc.Extra.Attrs, specutil.StrAttr("charset", c))
 	}
 	if c, ok := hasCollate(s.Attrs, nil); ok {
-		sc.Extra.Attrs = append(sc.Extra.Attrs, specutil.StrAttr("collation", c))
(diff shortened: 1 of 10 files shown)


sql/mysql/sqlspec.go
```diff
@@ -270,7 +270,12 @@ func convertCharset(spec specutil.Attrer, attrs *[]schema.Attr) error {
 		}
 		*attrs = append(*attrs, &schema.Charset{V: s})
 	}
-	if attr, ok := spec.Attr("collation"); ok {
+	// For backwards compatibility, accepts both "collate" and "collation".
+	attr, ok := spec.Attr("collate")
+	if !ok {
+		attr, ok = spec.Attr("collation")
```
masseelch, line 276: Meh. Clutters code. We are pre-release (at least not v1) and can work with API changes, can't we?

masseelch (approved):

sql/mysql/sqlspec.go
rotemtam: We actually don't know the impact this will have on users. I think it's fine to mention it in the release notes and for the next minor version (bump to 0.4) we can announce its drop. WDYT?

masseelch: I agree to that. We should given users time to adjust. But we cannot slug every change with us for ever :-)

a8m: Agree, it's temporary and will be removed in a few weeks from now.

PR #590, opened 2022-02-17
sql: use enums for referential actions

a8m:


doc/md/ddl/sql.mdx
```diff
@@ -190,8 +190,8 @@ as references to columns in other tables.
 foreign_key "manager_fk" {
   columns = [column.manager_id]
   ref_columns = [column.id]
-  on_delete = "CASCADE"
-  on_update = "NO ACTION"
+  on_delete = CASCADE
+  on_update = NO_ACTION
```
masseelch, line 194: Can we leave one test with a string? They are still working, don't they? I don't want to "accidently" break that behaviour in the future, so having a test for it is a great idea :-)

a8m: I actually have one in `internal/integration/hclsqlspec` 😃

PR #591, opened 2022-02-18
fixed sqlite escaping on migrate insert

devnode:


sql/sqlite/migrate.go
```diff
@@ -344,7 +344,32 @@ func (s *state) copyRows(from *schema.Table, to *schema.Table, changes []schema.
...
+			toCols += cc
+		} else {
+			toCols += "`" + cc + "`"
+		}
+		if i+1 < len(toC) {
+			toCols += ", "
+		}
+	}
+
+	var fromCols string
```
a8m, line 360: The code for `toC` was correct in the previous commit, but for `fromC` is actually more tricky because it can contain expression and a placeholder (see, L332). 

Maybe add the backticks when appending the columns? L330, L339, L343-4.

a8m: Re-reading it now, the code is correct. Another approach is to move it to a private function and reuse it here:

```go
func identComma(c []string) string {
	b := &sqlx.Builder{QuoteChar: '`'}
	b.MapComma(c, func(i int, b *sqlx.Builder) {
		if strings.ContainsRune(c[i], '`') {
			b.Ident(c[i])
		} else {
			b.WriteString(c[i])
		}
	})
	return b.String()
}
```

devnode: yea I didn't have time to deep dive into the code so I went with the least invasive option. let me know if anything else needs to be adjusted.

sql/sqlite/migrate.go
```diff
@@ -344,7 +344,8 @@ func (s *state) copyRows(from *schema.Table, to *schema.Table, changes []schema.
 			fromC = append(fromC, column.Name)
 		}
 	}
-	stmt := fmt.Sprintf("INSERT INTO %s (%s) SELECT %s FROM %s", to.Name, strings.Join(toC, ", "), strings.Join(fromC, ", "), from.Name)
+
```
a8m, line 347:
```suggestion
```

a8m (approved):
Looks good @devnode! Thanks a lot for the contribution. It's appreciated.

PR #595, opened 2022-02-22
sql/mysql: use default attributes when from realm is nil

yonidavidson:


sql/mysql/diff.go
```diff
@@ -21,12 +21,16 @@ type diff struct{ conn }
...
 	// Charset change.
-	if change := d.charsetChange(from.Attrs, from.Realm.Attrs, to.Attrs); change != noChange {
+	if change := d.charsetChange(from.Attrs, ra, to.Attrs); change != noChange {
 		changes = append(changes, change)
 	}
 	// Collation change.
-	if change := d.collationChange(from.Attrs, from.Realm.Attrs, to.Attrs); change != noChange {
+	if change := d.collationChange(from.Attrs, ra, to.Attrs); change != noChange {
 		changes = append(changes, change)
 	}
```
a8m, line 35: I guess the code is panic here (@zeevmoney also showed me that), but why do you have a schema without a Realm?

sql/mysql/diff.go
```diff
@@ -21,12 +21,16 @@ type diff struct{ conn }
 // SchemaAttrDiff returns a changeset for migrating schema attributes from one state to the other.
 func (d *diff) SchemaAttrDiff(from, to *schema.Schema) []schema.Change {
 	var changes []schema.Change
+	var ra []schema.Attr
+	if from.Realm != nil {
+		ra = from.Realm.Attrs
+	}
```
a8m, line 27:
```suggestion
	var (
		topAttrs []schema.Attr
		changes []schema.Change
	)
	if from.Realm != nil {
		topAttrs = from.Realm.Attrs
	}
```

sql/mysql/diff.go
yonidavidson: because sometimes we diff with schemas that are stored in text mode (and the realm is not marshaled )

a8m (approved):
💯

PR #596, opened 2022-02-23
sql/TiDB: break down alter statements

hedwigz:
TiDB does not support multischema changes (see [issue](https://github.com/pingcap/tidb/issues/14766)) - therefore we break down each change to its own `alter` statement.  
It is therefore important that index/foreign key drop occur before their associated columns.  
Also added support for the syntax of FK reference options (e.g "ON DELETE CASCADE")


hedwigz:
For benchmark - running the `integration/mysql_test` I get 99 passes and 33 fail

sql/mysql/migrate.go
```diff
@@ -266,13 +266,15 @@ func (s *state) dropTable(drop *schema.DropTable) {
...
-	for _, change := range skipAutoChanges(modify.Changes) {
+	for _, change := range s.skipAutoChanges(modify.Changes) {
 		switch change := change.(type) {
 		// Constraints should be dropped before dropping columns, because if a column
 		// is a part of multi-column constraints (like, unique index), ALTER TABLE
 		// might fail if the intermediate state violates the constraints.
 		case *schema.DropIndex:
 			changes[0] = append(changes[0], change)
+		case *schema.DropForeignKey:
+			changes[0] = append(changes[0], change)
```
a8m, line 277: Why it's here? Try this:

```
mysql> show create table users;
+-------+-------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                        |
+-------+-------------------------------------------------------------------------------------------------------------------------------------+
| users | CREATE TABLE `users` (
  `id` int NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci |
+-------+-------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> show create table pets;
+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table | Create Table                                                                                                                                                                                                                                                        |
+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| pets  | CREATE TABLE `pets` (
  `id` int NOT NULL,
  `owner` int DEFAULT NULL,
  PRIMARY KEY (`id`),
  KEY `owner` (`owner`),
  CONSTRAINT `pets_ibfk_1` FOREIGN KEY (`owner`) REFERENCES `users` (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci |
+-------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> alter table pets drop column owner;
ERROR 1828 (HY000): Cannot drop column 'owner': needed in a foreign key constraint 'pets_ibfk_1'
```

sql/mysql/migrate.go
```diff
@@ -307,8 +309,17 @@ func (s *state) modifyTable(modify *schema.ModifyTable) error {
...
+			if strings.Contains(s.version, "TiDB") {
+				for _, c := range changes[i] {
+					if err := s.alterTable(modify.T, []schema.Change{c}); err != nil {
+						return err
+					}
+				}
+			} else {
+				if err := s.alterTable(modify.T, changes[i]); err != nil {
+					return err
+				}
```
a8m, line 322: This code can be deleted by wrapping the differ and breaking one ModifyTable command into n ModifyTable commands.

internal/integration/tidb_test.go
```diff
@@ -0,0 +1,1074 @@
...
+	"ariga.io/atlas/sql/schema"
+
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/stretchr/testify/require"
+)
+
+func tidbRun(t *testing.T, fn func(*myTest)) {
+	myTests.Do(func() {
+		myTests.drivers = make(map[string]*myTest)
+		for version, port := range map[string]int{"TiDB": 4309} {
```
a8m, line 23: Only one? You have 2 in the docker-compose. 

If it's only one remove the loop.

internal/integration/tidb_test.go
```diff
@@ -0,0 +1,1074 @@
...
+			Parts:  []*schema.IndexPart{{C: parts[0]}, {C: parts[1]}, {C: parts[2]}},
+		})
+		changes := t.diff(t.loadUsers(), usersT)
+		require.NotEmpty(t, changes, "usersT contains 2 new columns and 1 new index")
+		t.migrate(&schema.ModifyTable{T: usersT, Changes: changes})
+		ensureNoChange(t, usersT)
+
+		// In TiDB, dropping a column should remove it from the key.
+		// However, on MariaDB an explicit DROP/ADD INDEX is required.
+		if t.mariadb() || t.tidb() {
```
a8m, line 85: Why mariadb is here?

internal/integration/tidb_test.go
```diff
@@ -0,0 +1,1074 @@
...
+			usersT.ForeignKeys = usersT.ForeignKeys[:len(usersT.ForeignKeys)-1]
+			changes = t.diff(t.loadUsers(), usersT)
+			require.Len(t, changes, 2)
+			t.migrate(&schema.ModifyTable{T: usersT, Changes: changes})
+			ensureNoChange(t, usersT)
+		})
+	})
+}
+
+// func TestTiDB_Ent(t *testing.T) {
```
a8m, line 453: remove all these comments.

sql/mysql/tidb.go
```diff
@@ -8,14 +8,158 @@ import (
...
+	case *schema.AddColumn:
+		return 7
+	case *schema.ModifyTable:
+		// each modifyTable should have a single change since we apply `flat` before we sort.
+		return priority(c.Changes[0])
+	case *schema.ModifySchema:
+		// each modifyTable should have a single change since we apply `flat` before we sort.
+		return priority(c.Changes[0])
+	default:
+		return 8
```
a8m, line 57: Why you it can be something like:
```suggestion
	case *schema.ModifyTable:
		// each modifyTable should have a single change since we apply `flat` before we sort.
		return priority(c.Changes[0])
	case *schema.ModifySchema:
		// each modifyTable should have a single change since we apply `flat` before we sort.
		return priority(c.Changes[0])
		case *schema.DropIndex, *schema.DropForeignKey, *schema.ModifyIndex, *schema.ModifyForeignKey, ...:
		return 0
	default:
		1
```

sql/mysql/tidb.go
```diff
@@ -8,14 +8,158 @@ import (
...
+				flat = append(flat, &schema.ModifySchema{
+					S:       m.S,
+					Changes: []schema.Change{c},
+				})
+			}
+		default:
+			flat = append(flat, change)
+		}
+	}
+	return flat
```
a8m, line 85: nice

sql/mysql/tidb.go
```diff
@@ -8,14 +8,158 @@ import (
...
+		conn: p.conn,
+		Plan: migrate.Plan{
+			Name: name,
+			// A plan is reversible, if all
+			// its changes are reversible.
+			Reversible:    true,
+			Transactional: false,
+		},
+	}
+
```
a8m, line 106:
```suggestion
```

sql/mysql/tidb.go
```diff
@@ -8,14 +8,158 @@ import (
...
+	case *schema.ModifySchema:
+		// each modifyTable should have a single change since we apply `flat` before we sort.
+		return priority(c.Changes[0])
+	default:
+		return 8
+	}
+}
+
+// flat takes a list of changes and breaks them down to single atomic changes (e.g: no modify table with multiple AddColumn inside it).
+// note that the only "changes" that include sub-changes are `ModifyTable` and `ModifySchema`.
```
a8m, line 62:
```suggestion
// flat takes a list of changes and breaks them down to single atomic changes (e.g: no ModifyTable
// with multiple AddColumn inside it). Note that, the only "changes" that include sub-changes are
// `ModifyTable` and `ModifySchema`.
```

sql/mysql/tidb.go
```diff
@@ -8,14 +8,158 @@ import (
...
+		default:
+			flat = append(flat, change)
+		}
+	}
+	return flat
+}
+
+// PlanChanges returns a migration plan for the given schema changes.
+func (p *tplanApply) PlanChanges(ctx context.Context, name string, changes []schema.Change) (*migrate.Plan, error) {
+	// break down changes to atomic operations (tidb does not support multiple changes in each alter)
```
a8m, line 90:
```suggestion
```
It's already written in this file.

sql/mysql/tidb.go
```diff
@@ -8,14 +8,158 @@ import (
...
+	fc := flat(changes)
+	// sort the changes according to the right order of execution.
```
a8m, line 92:
```suggestion
```
Obvious comment.

sql/mysql/tidb.go
```diff
@@ -8,14 +8,158 @@ import (
...
+			Name: name,
+			// A plan is reversible, if all
+			// its changes are reversible.
+			Reversible:    true,
+			Transactional: false,
+		},
+	}
+
+	for _, c := range fc {
+		// use the planner of MySQL with each "atomic" chanage.
```
a8m, line 108:
```suggestion
		// Use the planner of MySQL with each "atomic" chanage.
```

sql/mysql/tidb.go
```diff
@@ -8,14 +8,158 @@ import (
...
+	return t.diff.IndexPartAttrChanged(from, to)
+}
+
+func (t tdiff) IsGeneratedIndexName(table *schema.Table, s *schema.Index) bool {
+	return t.diff.IsGeneratedIndexName(table, s)
+}
+
+func (t tdiff) ReferenceChanged(from, to schema.ReferenceOption) bool {
+	return t.diff.ReferenceChanged(from, to)
 }
```
a8m, line 163: This is not needed. It embeds the `diff` struct.

sql/mysql/tidb.go
```diff
@@ -8,14 +8,158 @@ import (
...
+	// tdiff decorates MySQL diff.
+	tdiff struct{ diff }
+	// tinspect decorates MySQL inspect.
+	tinspect struct{ inspect }
+)
+
+// priority computes the priority of each change.
+// TiDB does not support multischema alters (no multiple changes in a single alter statement) and therefore
+// we have to break down each alter. priority helps order the alters so they work.
+// for example - we can't drop a column if it is a foreign key - we have to drop the fk first. this is why DropForeignKey gets a different priority
```
a8m, line 31: Arrange it better.
```suggestion
// priority computes the priority of each change.
//
// TiDB does not support multischema ALTERs (i.e. multiple changes in a single ALTER statement).
// Therefore, we have to break down each alter. This function helps order the ALTERs so they work.
// e.g. priority gives precedence to DropForeignKey over DropColumn, because a column cannot be
// dropped if its foreign key was not dropped before.  
```

sql/mysql/tidb.go
```diff
@@ -69,13 +214,22 @@ func (i *tinspect) setFKs(s *schema.Schema, t *schema.Table) error {
 		if len(m) != 5 {
 			return fmt.Errorf("unexpected number of matches for a table constraint: %q", m)
 		}
-		ctName, clmns, refTableName, refClmns := m[1], m[2], m[3], m[4]
+		statement, ctName, clmns, refTableName, refClmns := m[0], m[1], m[2], m[3], m[4]
```
a8m, line 217:
```suggestion
		stmt, ctName, clmns, refTableName, refClmns := m[0], m[1], m[2], m[3], m[4]
```

a8m (commented):
Amazing!

Looks good, please address these minor comments and I'll approve this.

a8m (approved):
💯

PR #598, opened 2022-02-23
sql/mysql: mysql implementation for schema.Normalize

a8m:
You know the context. I'll add more docs to it after I'll push the rest of the implementation for the other drivers.


rotemtam (approved):
Twins!!

PR #600, opened 2022-02-23
sql/postgres: support capital tables

kinggor1010:
…cause "to_regclass" will always return null if a table has capital letters


masseelch:
@kinggor1010 Wow! Amazing, that you found this :-) 

Can I ask you to add a test for this? If you need help with that, don't hesitate to ask.

a8m:
I actually promised @kinggor1010 that I'll handle the test cases for it.
See discussion on our Discord.

PR #603, opened 2022-02-27
sql/tidb: add more integration tests

hedwigz:
Not many integration tests are left. namely:
TestTiDB_Ent  
TestTiDB_HCL  
TestTiDB_CLI  
TestTiDB_CLI_MultiSchema

### DIFF
--- a/internal/integration/tidb_test.go
+++ b/internal/integration/tidb_test.go
@@ -5,6 +5,7 @@
 package integration
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"testing"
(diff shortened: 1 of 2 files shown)


internal/integration/tidb_test.go
```diff
@@ -429,3 +430,391 @@ func TestTiDB_ForeignKey(t *testing.T) {
...
+	})
+}
+
+func TestTiDB_Sanity(t *testing.T) {
+	n := "atlas_types_sanity"
+	t.Run("Common", func(t *testing.T) {
+		ddl := `
+create table atlas_types_sanity
+(
+    tBit                        bit(10)              default b'1000000001'                                       null,
```
a8m, line 488: fix ident.

sql/mysql/tidb.go
```diff
@@ -157,10 +158,41 @@ func (i *tinspect) patchSchema(ctx context.Context, s *schema.Schema) (*schema.S
...
+			i.patchColumn(ctx, c)
+		}
 	}
 	return s, nil
 }
 
+func (i *tinspect) patchColumn(ctx context.Context, c *schema.Column) {
+	switch c.Type.Type.(type) {
+	case *BitType:
+		// TiDB has a bug where it does not format bit default value correctly. see: https://github.com/pingcap/tidb/issues/32655.
```
a8m, line 171:
```suggestion
		// TiDB has a bug where it does not format bit default value correctly.
		// See: https://github.com/pingcap/tidb/issues/32655.
```

sql/mysql/tidb.go
```diff
@@ -157,10 +158,41 @@ func (i *tinspect) patchSchema(ctx context.Context, s *schema.Schema) (*schema.S
...
+		if lit, ok := c.Default.(*schema.Literal); ok {
+			lit.V = bufferToBitLiteral([]byte(lit.V))
+		}
+	}
+}
+
+// bufferToBitLiteral converts a buffer to MySQL bit literal (e.g: []byte{4} -> b'100')
+func bufferToBitLiteral(b []byte) string {
```
a8m, line 179: why buffer? use `bytes` instead.

sql/mysql/tidb.go
```diff
@@ -157,10 +158,41 @@ func (i *tinspect) patchSchema(ctx context.Context, s *schema.Schema) (*schema.S
...
+	builder.WriteString("b'")
+	for i, digit := range b {
+		if i != 0 {
+			bits := strconv.FormatInt(int64(digit), 2)
+			pads := 8 - len(bits)
+			builder.WriteString(strings.Repeat("0", pads))
+			builder.WriteString(bits)
+		} else {
+			builder.WriteString(strconv.FormatInt(int64(digit), 2))
+		}
```
a8m, line 190: I'm not sure the arithmetic here is correct. Use `binary.Uvarint`/`binary.Varint` with `Fprintf(w, "%b", n)` instead.

internal/integration/tidb_test.go
hedwigz: image
LGTM

sql/mysql/tidb.go
hedwigz: `binary.Uvarint` returns `2` for `[]byte{2,1}` in which case I expect `513`

sql/mysql/tidb.go
```diff
@@ -157,11 +158,36 @@ func (i *tinspect) patchSchema(ctx context.Context, s *schema.Schema) (*schema.S
...
+			lit.V = bytesToBitLiteral([]byte(lit.V))
+		}
+	}
+}
+
+// bytesToBitLiteral converts a bytes to MySQL bit literal.
+// e.g. []byte{4} -> b'100', []byte{2,1} -> b'1000000001'.
+// See: https://github.com/pingcap/tidb/issues/32655.
+func bytesToBitLiteral(b []byte) string {
+	bytes := []byte{0, 0, 0, 0, 0, 0, 0, 0}
```
a8m, line 182:
```suggestion
	bytes := make([]byte, 8)
```
or, [8]byte

a8m (approved):

sql/mysql/tidb.go
```diff
@@ -157,11 +158,36 @@ func (i *tinspect) patchSchema(ctx context.Context, s *schema.Schema) (*schema.S
...
 
-// e.g CONSTRAINT "" FOREIGN KEY ("foo_id") REFERENCES "foo" ("id")
+func (i *tinspect) patchColumn(ctx context.Context, c *schema.Column) {
+	switch c.Type.Type.(type) {
+	case *BitType:
+		// TiDB has a bug where it does not format bit default value correctly.
+		if lit, ok := c.Default.(*schema.Literal); ok {
+			lit.V = bytesToBitLiteral([]byte(lit.V))
+		}
+	}
```
a8m, line 175: An alternative:
```suggestion
	b, ok := c.Type.Type.(*BitType)
	if !ok {
		return
	}
	// TiDB has a bug where it does not format bit default value correctly.
	if lit, ok := c.Default.(*schema.Literal); ok {
		lit.V = bytesToBitLiteral([]byte(lit.V))
	}
```

PR #608, opened 2022-03-01
cmd/atlas: add support for simple mysql url format

a8m:


cmd/action/mux_test.go
```diff
@@ -153,3 +157,41 @@ func Test_PostgresSchemaDSN(t *testing.T) {
...
+			calls, l := mockServer(t)
+			require.NoError(t, mysqld.SetLogger(log.New(ioutil.Discard, "", 1)))
+			_, err := action.DefaultMux.OpenAtlas(fmt.Sprintf(u, l.Addr()))
+			require.Error(t, err, "mock server rejects all incoming connections")
+			require.NotZero(t, *calls)
+		}
+	})
+}
+
+func mockServer(t *testing.T) (*int, net.Listener) {
```
a8m, line 180: Didn't really like it, but wanted to add tests to the OpenAtlas part.

There's no way to register a "mock driver" with the same name, because the mysql package already registers it, and `database/sql.Register` will panic in case of multiple calls with the same name.

Another alternative was to call `sql.unregisterAllDrivers`, but this requires using the `unsafe` package which I prefer to avoid.

cmd/action/mux.go
```diff
@@ -43,8 +43,9 @@ func NewMux() *Mux {
 }
 
 var (
-	defaultMux = NewMux()
-	inMemory   = regexp.MustCompile("^file:.*:memory:$|:memory:|^file:.*mode=memory.*")
+	// DefaultMux is the default Mux that is used by the different commands.
+	DefaultMux = NewMux()
```
masseelch, line 47: I don't like the fact that this is exported only to do the test. Can we do this different (by renaming the test package or rewrite the test)?

masseelch: e.g. calling `NewMux()` is the same as `defaultMux` (as of now).

cmd/action/mux.go
```diff
@@ -43,8 +43,9 @@ func NewMux() *Mux {
...
+	reMemMode  = regexp.MustCompile("^file:.*:memory:$|:memory:|^file:.*mode=memory.*")
```
masseelch, line 48:
```suggestion
	reMemMode  = regexp.MustCompile(":memory:|^file:.*mode=memory.*")
```

That does the same, doesn't it?

a8m: Yes, it was there. I didn't write it.

cmd/action/mux.go
a8m: > I don't like the fact that this is exported only to do the test. Can we do this different (by renaming the test package or rewrite the test)?

Currently, the different commands depend on it and use it, but there are no tests for it. I'd structure it differently, but I can't add a piece of code without testing it. 

> e.g. calling NewMux() is the same as defaultMux (as of now).

Not really, because in the provider.go file, the `Register` is called on init.

PR #611, opened 2022-03-02
cmd/atlas: introduce the dev-url to apply command

a8m:


cmd/action/apply.go
```diff
@@ -31,16 +32,18 @@ var (
 	ApplyCmd = &cobra.Command{
 		Use:   "apply",
 		Short: "Apply an atlas schema to a target database.",
-		Long: "`atlas schema apply`" + ` plans and executes a database migration to bring a given database
-to the state described in the Atlas schema file. Before running the migration, Atlas will print the migration
-plan and prompt the user for approval.
+		// Use 80-columns as max width.
```
masseelch, line 35: :-1: I don't like 80 cols. We are no longer working on SD resolutions and no-one is coding on a smartphone

masseelch (approved):

cmd/action/apply.go
a8m: What do you don't like about this output? When the screen size is smaller (my case), the line breaks make it ugly. 

image

masseelch: Your screen supports 120 too by the looks of it.

masseelch: image

PR #616, opened 2022-03-03
sql/migrate: (BC) read from state from migration directory instead of an extra state reader

masseelch:
This is the first PR of several to add the `atlas migrate` commands.

### DIFF
--- a/sql/migrate/migrate.go
+++ b/sql/migrate/migrate.go
@@ -158,8 +158,8 @@ type (
 	}
 )
 
-// New creates a new Planner.
-func New(drv Driver, dir Dir, fmt Formatter) *Planner {
+// NewPlanner creates a new Planner.
+func NewPlanner(drv Driver, dir Dir, fmt Formatter) *Planner {
 	return &Planner{
 		drv: drv,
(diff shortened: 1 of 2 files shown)


sql/migrate/migrate.go
```diff
@@ -214,7 +238,11 @@ type LocalDir struct {
...
 func NewLocalDir(path string) (*LocalDir, error) {
 	fi, err := os.Stat(path)
-	if err != nil {
+	if err == os.ErrNotExist {
+		if err := os.MkdirAll(path, 0755); err != nil {
+			return nil, err
+		}
+	} else if err != nil {
 		return nil, err
 	}
```
a8m, line 247: I'm not sure it's better in some way

```suggestion
	if err == os.ErrNotExist {
		err = os.MkdirAll(path, 0755)
	}
	if err != nil {
		return nil, err
	}
```

a8m (approved):
💯

sql/migrate/migrate.go
masseelch: Less explicit so I vote in favor of mine 😎

PR #624, opened 2022-03-04
sql/migrate: add hash integrity file

masseelch:
Suppose you have multiple teams develop a feature in parallel and both of them need a migration. If Team A and Team B do not check in with each other we might end up with a broken migration directory since new files do not raise a merge conflict for example.

With this PR every time a new migration file is created programatically, a hash is computed. This is done by walking over the whole migration dir in lexicographic order and hashing the contents. That file is then checked into version control. In the above case the second merging team will have to resolve a merge conflict on the hashfile. Atlas itself will get a `Validate` method to check if the hash and the migration directory matches.

For now this only works with programmtically created migration files, but I will add a way to re-compute the hash for the migration directory since this is needed when editing a file, adding one by hand or resolving the above mentioned conflict.

The `Hash` interface is needed, since one might not want to save the integrity file in the migration directory, for example when the migration directory does live in another repo / s3 bucket / whatever.


a8m:
> plus the contents of the new migration files.

Maybe, migration filename? because content can be changed by the user. e.g. `CREATE INDEX` => `CREATE INDEX CONCURRENTLY`, I would expect atlas generate the hash / update lock file, after running `atlas diff`.

Update: we talked about it over the phone, but I feel like the process will be simpler if we'll use the filenames in the lock file. 

A few thoughts on the file format - I think we should keep a list of filenames (with hashes) if we want to fail in vcs merge (due to conflicts). i.e. a file with one line is not good enough. Why the hash? because we don't want the user to resolve the git conflict manully, it should be generated by atlas.


Example:
(We can use any format. HCL or simple text are both fine imo).
```text
1_add_t1: 0OppXMNlmlGpAvSC7aaebg==
2_add_t2: iu0r71q7BjEagxdWWomwjA==
<<<<<<< HEAD
3_add_t3: OV+o/g6txWQyZ/kKh+etmw==
=======
3_edit_t2: WhXwXVbR4koUTEMj3N0JVQ==
>>>>>>> ...
```
In this case, I expect atlas to provide a way to clean current (local) migration file, let's say with a command like `reset` or `rm`, but it can also be done manaully. After that, the user can run again `atlas migrate diff 3_edit_t2` and get its new updated migration file.

But what if the user check the latest remote migration and see that there are not conflicts with its local (in development) changes? I think in this case, we should provide a way to `resolve`/`merge` this state. i.e. calcualte the hash and update the lock file.

> For now this only works with programmtically created migration files, but I will add a way to re-compute the hash for the migration directory since this is needed when editing a file, adding one by hand or resolving the above mentioned conflict.

Maybe we can introduce a new command/flag, `atlas migrate new ...`/`atlas migrate diff ... --new`

masseelch:
> ```
1_add_t1: 0OppXMNlmlGpAvSC7aaebg==
2_add_t2: iu0r71q7BjEagxdWWomwjA==
<<<<<<< HEAD
3_add_t3: OV+o/g6txWQyZ/kKh+etmw==
=======
3_edit_t2: WhXwXVbR4koUTEMj3N0JVQ==
>>>>>>> ...
```

This format won't raise a merge conflict if there have been two lines added cause git handles that correctly, if I am not mistaken. So this won't work if there isn't an extra line with a complete hash somewhere, correct?

An idea to keep the file human readable is to take your format but add a global has as first line. 

> In this case, I expect atlas to provide a way to clean current (local) migration file, let's say [...]

We don't need that if we just add a command to (force) compute the integrity hash file. Whenever the user did a manual edit to a file (edit, create, delete), he needs to run the command to create a valid hash file. 

We can then add a `atlas migrate validate` command to check files and hash against each other. This can also be run in CI similar to the `go generate` checks.

> Maybe we can introduce a new command/flag, `atlas migrate new ...`/`atlas migrate diff ... --new`/`atlas migrate diff ... --empty`

With my idea we didn't need it. I think we should not add it unless the users request it. 

Thoughts? Also @rotemtam, @yonidavidson

**Comment by a8m** [comment]
> This format won't raise a merge conflict if there have been two lines added cause git handles that correctly, if I am not mistaken. So this won't work if there isn't an extra line with a complete hash somewhere, correct?

You're right, ignore the example. We can use a JSON format, or edit a line at the head of the file to cause the conflict. Having a `validate` command is useful, but we need an extra check here.

> An idea to keep the file human readable is to take your format but add a global hash as first line.

Exactly. But having a list of migration files is not only for making the file human-readable, it's necessary for keeping the history of the migrations. If users want to change the history (e.g. compaction), it can be done using `atlas`.

> With my idea we didn't need it. I think we should not add it unless the users request it.

Having just one hash of the migration directory, make it possible to change the history without noticing it. We can still provide a way to re-generate the lock file (recompute all hashes), and edit the history, but this should be done with an explicit command IMO.

**Comment by yonidavidson** [comment]
Trying to get things in order (for me to understand).
1. Migration files: each file will have timestamp_user_specific_name.sql
2. Lock file (like go.mod?) with each file and it's hash ordered by creation time.
3. Atlas will have helper commands to handle this. "atlas migration tidy"? for example" The lock file will have to be by "DB/Schema/Realm"? maybe it should be external since a few projects might change same DB and they need to share lock file no ?

**Comment by a8m** [comment]
> Atlas will have helper commands to handle this. "atlas migration tidy"? for example" The lock file will have to be by "DB/Schema/Realm"? maybe it should be external since a few projects might change same DB and they need to share lock file no ?

For a migration directory. 

@yonidavidson, I think it's best to jump on a quick meeting tomorrow, discuss this and unblock you @masseelch. Let me know what do you think.

File: sql/migrate/migrate.go [hunk 1]
```diff
@@ -191,6 +195,43 @@ func WithStateReader(dsr StateReader) PlannerOption {
 	}
 }
 
+// DisableHashSum disables the hash-sum functionality for the migration directory.
+func DisableHashSum() PlannerOption {
```
a8m, line 199: Maybe `DisableSumFile`?

```suggestion
func DisableSumFile() PlannerOption {
```

sql/migrate/migrate.go
```diff
@@ -227,6 +268,18 @@ func (p *Planner) WritePlan(plan *Plan) error {
...
+		if err != nil {
+			return err
+		}
+		b, err := sum.MarshalText()
+		if err != nil {
+			return err
+		}
+		return p.dir.WriteFile(hashFile, b)
+	}
+
```
a8m, line 282:
```suggestion
```

sql/migrate/migrate.go
```diff
@@ -5,8 +5,11 @@
 package migrate
 
 import (
+	"bufio"
 	"bytes"
 	"context"
+	"crypto/sha256"
```
a8m, line 11: wonder why `hash/fnv` is not enough.

sql/migrate/migrate.go
```diff
@@ -373,6 +397,117 @@ type templateFile struct {
 // Name implements the File interface.
 func (f *templateFile) Name() string { return f.n }
 
+// Filename of the migration directory integrity sum file.
+const hashFile = "atlas.sum"
+
+// HashFile represents the integrity sum file of the migration dir.
+type HashFile []struct{ N, H string }
```
a8m, line 404: love this 😍

sql/migrate/migrate.go
```diff
@@ -373,6 +397,117 @@ type templateFile struct {
...
+
+// HashSum reads the whole dir, sorts the files by name and creates a HashSum from its contents.
+func HashSum(dir Dir) (HashFile, error) {
+	var hs HashFile
+	h := sha256.New()
```
a8m, line 409:
```suggestion
	var (
		hs HashFile
		h = sha256.New()
	)
```
Just a suggestion, but I don't like to mix different declaration types in the same block

sql/migrate/migrate.go
```diff
@@ -373,6 +397,117 @@ type templateFile struct {
...
+	}
+	return nil
+}
+
+// MarshalText implements encoding.TextMarshaler.
+func (s HashFile) MarshalText() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	sha := sha256.New()
+	for _, f := range s {
+		buf.WriteString(fmt.Sprintf("%s h:%s\n", f.N, f.H))
```
a8m, line 472:
```suggestion
		fmt.Fprintf(buf, "%s h:%s\n", f.N, f.H)
```

`h` for hash? what is `h1` in go?

sql/migrate/migrate.go
```diff
@@ -373,6 +397,117 @@ type templateFile struct {
...
+		sha.Write([]byte(f.H))
+	}
+	ret := new(bytes.Buffer)
+	ret.WriteString(fmt.Sprintf("sum h:%s\n", base64.StdEncoding.EncodeToString(sha.Sum(nil))))
+	ret.Write(buf.Bytes())
+	return ret.Bytes(), nil
```
a8m, line 478:
```suggestion
	return fmt.Sprintf("sum h:%s\n%s", base64.StdEncoding.EncodeToString(sha.Sum(nil)), buf)
```

sql/migrate/migrate.go
```diff
@@ -373,6 +397,117 @@ type templateFile struct {
...
+}
+
+// UnmarshalText implements encoding.TextUnmarshaler.
+func (s *HashFile) UnmarshalText(b []byte) error {
+	sc := bufio.NewScanner(bytes.NewReader(b))
+	sc.Scan() // skip first line
```
a8m, line 484: We don't care about the sum when reading it?

sql/migrate/migrate.go
```diff
@@ -373,6 +397,117 @@ type templateFile struct {
...
+		li := strings.SplitN(sc.Text(), "h:", 2)
+		if len(li) != 2 {
+			return errors.New("invalid sum file")
+		}
+		*s = append(*s, struct{ N, H string }{strings.TrimSpace(li[0]), li[1]})
+	}
+	if err := sc.Err(); err != nil {
+		return err
+	}
+	return nil
```
a8m, line 495:
```suggestion
	return sc.Err()
```

sql/migrate/migrate.go
masseelch: https://go.dev/ref/mod#go-sum-files

> The hash column consists of an algorithm name (like h1) and a base64-encoded cryptographic hash, separated by a colon (:). Currently, SHA-256 (h1) is the only supported hash algorithm. If a vulnerability in SHA-256 is discovered in the future, support will be added for another algorithm (named h2 and so on).

We might wanna go with `h1` too then :-)

sql/migrate/migrate.go
masseelch: I chose SHA-256 because the `go.sum` does as well. https://go.dev/ref/mod#go-sum-files

sql/migrate/migrate.go
masseelch: Inspired by Ariel ;-)

sql/migrate/migrate.go
```diff
@@ -373,6 +397,117 @@ type templateFile struct {
...
+	if len(s) != len(o) {
+		return false
+	}
+	for i := range s {
+		if s[i].N != o[i].N || s[i].H != o[i].H {
+			return false
+		}
+	}
+	return true
+}
```
a8m, line 509: Don't you need to compare also the `sum` value?

sql/migrate/migrate.go
```diff
@@ -373,6 +397,117 @@ type templateFile struct {
...
+	}
+	return hs, nil
+}
+
+// ErrChecksumMismatch is returned from Validate if the hash sums don't match.
+var ErrChecksumMismatch = errors.New("checksum mismatch")
+
+// Validate checks if the migration dir is in sync with its sum file.
+// If they don't match ErrChecksumMismatch is returned.
+func Validate(dir Dir) error {
```
a8m, line 443: What about validation that all lines match the `sum` in the header?

sql/migrate/migrate.go
masseelch: No. The whole purpose of `sum` is to generate a merge conflict in VCS.

sql/migrate/migrate.go
masseelch: No, it is only used for VCS merge conflict and not for the validity of the file.

sql/migrate/migrate.go
```diff
@@ -373,6 +397,117 @@ type templateFile struct {
...
+			f, err := dir.Open(path)
+			if err != nil {
+				return err
+			}
+			defer f.Close()
+			if _, err = io.Copy(h, f); err != nil {
+				return err
+			}
+			hs = append(hs, struct{ N, H string }{path, base64.StdEncoding.EncodeToString(h.Sum(nil))})
+			h.Reset()
```
a8m, line 428: Why do you `Reset` it for each file? Didn't we say that each hash depends on its previous one?

sql/migrate/migrate.go
masseelch: I don't think it is needed. In what case should we check it?

sql/migrate/migrate.go
a8m: Good to know.

sql/migrate/migrate.go
a8m: 👍

sql/migrate/migrate.go
a8m: Asking, because the order is important and can't be edited by a human. Same thing for the `sum` attribute.

a8m (approved):
💯 

It's so beautiful! ❤️

PR #627, opened 2022-03-06
cmd/action: change defualt listen address to work on all cases

zeevmoney:


rotemtam (approved):
CI broken. Also, I think there were tests for this? 
Merge after fixing

PR #648, opened 2022-03-11
sql/postgres: implement schema.Locker

a8m:
Please, don't review yet. Working on integration tests.

### DIFF
--- a/sql/postgres/driver_test.go
+++ b/sql/postgres/driver_test.go
@@ -1,0 +1,103 @@
+// Copyright 2021-present The Atlas Authors. All rights reserved.
+// This source code is licensed under the Apache 2.0 license found
+// in the LICENSE file in the root directory of this source tree.
+
+package postgres
+
+import (
+	"context"
+	"io"
+	"testing"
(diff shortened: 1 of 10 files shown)


sql/postgres/driver_test.go
```diff
@@ -0,0 +1,103 @@
...
+		require.Nil(t, unlock)
+	})
+}
+
+func TestDriver_UnlockError(t *testing.T) {
+	db, m, err := sqlmock.New()
+	require.NoError(t, err)
+	d := &Driver{}
+	d.ExecQuerier = db
+	name, hash := "up", 1551306158
```
yonidavidson, line 74: These hash numbers are just random, right? it's not calculated or something

yonidavidson (approved):
LGTM

sql/postgres/driver_test.go
a8m: it's calculated (fnv), but it's here for testing.

hedwigz (approved):
LGTM

PR #651, opened 2022-03-15
internal/docker: docker client to spin up containers on the fly

masseelch:
This will be useful for spinning up an ephemeral dev db when replaying a migration directory or for normalizing. This is a first draft.


internal/docker/client.go
```diff
@@ -0,0 +1,230 @@
+package docker
```
a8m, line 1: header license. 
Why it's not under cmd/?

go.mod
```diff
@@ -5,34 +5,51 @@ go 1.17
...
 	github.com/stretchr/testify v1.7.1-0.20210427113832-6241f9ab9942
 	github.com/zclconf/go-cty v1.8.0
 	golang.org/x/mod v0.5.1
-	golang.org/x/sys v0.0.0-20211205182925-97ca703d548d // indirect
-	golang.org/x/text v0.3.7 // indirect
 )
 
 require (
+	github.com/docker/docker v20.10.13+incompatible
+	github.com/docker/go-connections v0.4.0
```
a8m, line 19: Wonder if it's not simpler to just use os.Exec (depends on the lib size).

internal/docker/client.go
```diff
@@ -0,0 +1,230 @@
...
+	"time"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
+	"github.com/docker/go-connections/nat"
+	mysqld "github.com/go-sql-driver/mysql"
+)
+
+const defaultPass = "pass"
```
a8m, line 24: Can't we generate it on on creation?

internal/docker/client.go
```diff
@@ -0,0 +1,230 @@
...
+		return nil, err
+	}
+	return &Container{
+		c:        c,
+		cfg:      *cfg,
+		ID:       con.ID,
+		Password: defaultPass,
+		Port:     p,
+	}, nil
+}
```
a8m, line 160: Code is beautiful, but isn't that simpler to use `docker run --rm`?

internal/docker/client.go
```diff
@@ -0,0 +1,230 @@
...
+
+// Driver returns the database/sql driver name.
+func (c *Container) Driver() string {
+	return strings.SplitN(c.cfg.Image, ":", 2)[0]
+}
+
+// DSN returns a DSN to connect to the Container.
+func (c *Container) DSN() string {
+	return fmt.Sprintf("root:%s@tcp(localhost:%s)/", c.Password, c.Port)
+}
```
a8m, line 209: Use `URL` instead. Also, this is a mysql driver format. Use the formats that exist here: https://atlasgo.io/cli/url

go.mod
masseelch: I cannot reliably shut down / destroy containers and have to do string parsing to get things like ID etc. 

Which eventually will make me rewriting the lib...

internal/docker/client.go
masseelch: Was not sure where to put it and started here. We can move the package, no problem

internal/docker/client.go
masseelch: I don't see the need to. It is just temprorary. Password does not really matter. It is needed for the various databases.

internal/docker/client.go
masseelch: It is not a URL in this case. It is meant to be used by `database/sql` which requires DSN, not URL. Atlas does.

cmd/action/docker.go
```diff
@@ -0,0 +1,232 @@
...
+	)...,
+	)
+}
+
+// Image sets the docker image to use.
+//
+// Example:
+//   - Image("mysql")
+//   - Image("postgres:13")
+func Image(i string) ConfigOption {
```
a8m, line 79:
```suggestion
// Image sets the docker image to use. For example:
//
//	Image("mysql")
//	Image("postgres:13")
//
func Image(i string) ConfigOption {
```

cmd/action/docker.go
```diff
@@ -0,0 +1,232 @@
...
+		c.Port = p
+		return nil
+	}
+}
+
+// Env sets the environment variables to pass to the container.
+//
+// Example:
+//   - mysql: Env("MYSQL_ROOT_PASSWORD=password")
+//   - pg:    Env("POSTGRES_PASSWORD=password")
```
a8m, line 102: Write examples in Go syntax.
```suggestion
// Env sets the environment variables to pass to the container. For example:
//
// 	Config(Image("mysql"), Env("MYSQL_ROOT_PASSWORD=password"))
// 	Config(Image("postgres"), Env("MYSQL_ROOT_PASSWORD=password"))
//
```

cmd/action/docker.go
```diff
@@ -0,0 +1,232 @@
...
+func freePort() (string, error) {
+	a, err := net.ResolveTCPAddr("tcp", ":0")
+	if err != nil {
+		return "", err
+	}
+	l, err := net.ListenTCP("tcp", a)
+	if err != nil {
+		return "", err
+	}
+	defer l.Close()
```
a8m, line 230: why defer?

a8m (approved):
It's beautiful 😍

PR #652, opened 2022-03-15
internal/integration: add tests for pg enums with default values

a8m:


masseelch:
I am confused, I thought this was a bug, how do the tests run correclty?

masseelch (approved):
:100:

PR #656, opened 2022-03-16
internal/integration: speed up test execution

masseelch:
Integration tests are now run each in its own job.

I had to do some kind of "hacky" solution, so it working correctly in the CI by using an env var when running the test. This does not affect local running.

### DIFF
--- a/internal/ci/main.go
+++ b/internal/ci/main.go
@@ -49,7 +49,7 @@ var (
 		{
 			Version: "mysql56",
 			Image:   "mysql:5.6.35",
-			Regex:   "MySQL.*/56",
+			Regex:   "MySQL.*/mysql56",
 			Env:     mysqlEnv,
 			Ports:   []string{"3306:3306"},
 			Options: mysqlOptions,
@@ -57,7 +57,7 @@ var (
 		{
 			Version: "mysql57",
 			Image:   "mysql:5.7.26",
-			Regex:   "MySQL.*/57",
+			Regex:   "MySQL.*/mysql57",
 			Env:     mysqlEnv,
 			Ports:   []string{"3307:3306"},
 			Options: mysqlOptions,
@@ -65,7 +65,7 @@ var (
 		{
 			Version: "mysql8",
 			Image:   "mysql:8",
-			Regex:   "MySQL.*/8",
+			Regex:   "MySQL.*/mysql8",
 			Env:     mysqlEnv,
 			Ports:   []string{"3308:3306"},
 			Options: mysqlOptions,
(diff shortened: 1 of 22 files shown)


internal/ci/main.go
```diff
@@ -0,0 +1,158 @@
...
+	"log"
+	"os"
+	"text/template"
+)
+
+//go:generate go run main.go
+
+// Job defines an integration job to run.
+type Job struct {
+	Version string   // version to test (used as GO_TEST_ONLY_VERSION env)
```
a8m, line 15: Do you still need this env?

internal/integration/integration_test.go
```diff
@@ -21,6 +23,22 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestMain(m *testing.M) {
+	var service string
+	flag.StringVar(&service, "service", "", "[mysql56, postgres10, tidb5, ...] what version to test")
```
a8m, line 28: `dialect` or `database` feel like a better name imo.

internal/integration/mysql_test.go
```diff
@@ -28,22 +29,21 @@ type myTest struct {
...
 
-var myTests struct {
-	sync.Once
-	drivers map[string]*myTest
-}
+var (
+	myTests struct {
+		drivers map[string]*myTest
+	}
+	myPorts = map[string]int{
```
a8m, line 36: you can keep it under the `myTests` variable.

internal/integration/script_test.go
```diff
@@ -174,11 +174,15 @@ func (t *myTest) cmdCmpShow(ts *testscript.TestScript, _ bool, args []string) {
 
 func (t *pgTest) cmdCmpShow(ts *testscript.TestScript, _ bool, args []string) {
 	cmdCmpShow(ts, args, func(schema, name string) (string, error) {
-		buf, err := exec.Command("docker", "ps", "-qa", "-f", fmt.Sprintf("name=postgres%s", t.version)).CombinedOutput()
+		buf, err := exec.Command("docker", "ps", "-qa", "-f", fmt.Sprintf("publish=%d", t.port)).CombinedOutput()
```
a8m, line 177: Can you elaborate on why filter by port and not by name?

a8m (approved):
Looks good, but please see questions before landing.

internal/integration/script_test.go
a8m: Got the answer on our chat.

internal/ci/main.go
masseelch: No. I'll change the docs

PR #658, opened 2022-03-18
cmd/action: add user explanation when checksum mismatch occur

masseelch:
More clear message if there is an checksum mismatch on the CLI.

image

### DIFF
--- a/cmd/action/migrate.go
+++ b/cmd/action/migrate.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"strings"
 	"text/template"
 	"time"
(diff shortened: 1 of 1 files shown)


cmd/action/migrate.go
```diff
@@ -49,7 +50,17 @@ var (
 				if err != nil {
 					return err
 				}
-				return migrate.Validate(dir) // TODO(masseelch): tell the user what's wrong
+				if err := migrate.Validate(dir); err != nil {
+					fmt.Fprintf(
+						os.Stderr,
+						"Error: %s\n\nYou have a checksum error in your migration directory.\n"+
+							"This usually happens if you manually create or edit a migration file.\n"+
```
a8m, line 57: usually or can?

cmd/action/migrate.go
```diff
@@ -49,7 +50,17 @@ var (
...
+							"Please check your migration files and run\n\n"+
+							"'atlas migrate hash --force'\n\nto re-hash the contents and resolve the error.\n\n",
```
a8m, line 59: Next action items are to provide additional guidance and explanation on this.

a8m (approved):
💯

cmd/action/migrate.go
masseelch: It will always happen if you manually edit any file

PR #663, opened 2022-03-21
sql/migrate: GlobStateReader will now try to get a lock before replay…

masseelch:
…ing the migration directory, check for the driver to be in an empty state and clean up after the replay is done


a8m (approved):
💯

PR #665, opened 2022-03-22
cmd/action: add docker provider

masseelch:


cmd/action/docker.go
```diff
@@ -186,6 +187,7 @@ func (c *Container) Wait(ctx context.Context, timeout time.Duration) error {
 			if err := db.PingContext(ctx); err != nil {
 				continue
 			}
+			fmt.Println("Service is ready to connect!")
```
a8m, line 190: Can we remove these printing from the CLI, or at least enable them in verbose mode? It doesn't feel UNIX-ish and will also add noise when redirecting stdout (e.g. atlas ... > out.sql).

cmd/action/docker.go
```diff
@@ -203,7 +205,7 @@ func (c *Container) Driver() string {
 // DSN returns a DSN to connect to the Container.
 func (c *Container) DSN() (string, error) {
 	switch drv := c.Driver(); drv {
-	case "mysql", "maria", "mariadb":
+	case "mysql", "mariadb":
```
a8m, line 208: 👍

cmd/action/provider.go
```diff
@@ -18,9 +22,10 @@ func init() {
 	DefaultMux.RegisterProvider("mariadb", mysqlProvider)
 	DefaultMux.RegisterProvider("postgres", postgresProvider)
 	DefaultMux.RegisterProvider("sqlite", sqliteProvider)
+	DefaultMux.RegisterProvider("docker", dockerProvider)
```
a8m, line 25: 😍

cmd/action/provider.go
```diff
@@ -75,3 +80,63 @@ func sqliteProvider(dsn string) (*Driver, error) {
 		Closer:      db,
 	}, nil
 }
+
+var reDockerConfig = regexp.MustCompile("(mysql|mariadb)(?::([0-9a-zA-Z.-]+)?(\\?.*)?)?")
```
a8m, line 84: Can we be URL compliant and use`url.Parse`?

cmd/action/provider_test.go
```diff
@@ -0,0 +1,24 @@
+package action
```
a8m, line 1: license

cmd/action/provider.go
```diff
@@ -75,3 +80,63 @@ func sqliteProvider(dsn string) (*Driver, error) {
...
+	if cerr != nil {
+		err = fmt.Errorf("%w: %v", derr, cerr)
+	}
+	return
+}
+
+func dockerProvider(ctx context.Context, dsn string) (*Driver, error) {
+	// The DSN has the driver part (docker:// removed already.
+	// Get rid of the query arguments, and we have the image name.
+	m := reDockerConfig.FindStringSubmatch(dsn)
```
a8m, line 106: See my comment above about url.Parse.

cmd/action/docker.go
masseelch: Currently the exec commands to spit out to os.Stdout as well, we would enable that as well. 

Another thing: we can print to stderr instead, in that case redirect would work as expected.

cmd/action/provider.go
masseelch: I'll check

cmd/action/docker.go
```diff
@@ -36,6 +36,7 @@ type (
 	// A Container is an instance of a created container.
 	Container struct {
 		cfg DockerConfig // DockerConfig used to create this container
+		Out io.Writer    // custom write to log status messages to
```
a8m, line 39: If it's exported, as a proper doc.
```suggestion
		// Out is a custom writer to log status messages to.
		Out io.Writer
```

btw, I'd replace this with a custom `Log` function.

cmd/action/docker.go
```diff
@@ -220,6 +222,8 @@ func (c *DockerConfig) validate() error {
 	return nil
 }
 
+func (DockerConfigOption) providerOption() {}
```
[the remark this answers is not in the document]
masseelch: damn, forgot to remove

cmd/action/docker.go
masseelch: It is used to pipe the exec docker commands output.

a8m (approved):

Issue #667, opened 2022-03-23
Postgres Reverse change missing comma 


tprebs:
I noticed as part of using the new ent versioned migrations that all of the down scripts with column modifications were not adding commas in correctly

Up generates correctly e.g. 
```sql
ALTER TABLE "subscriptions" ALTER COLUMN "created_at" SET DEFAULT CURRENT_TIMESTAMP, ALTER COLUMN "updated_at" SET DEFAULT CURRENT_TIMESTAMP;
```

Down has missing commas between each column change
```sql
ALTER TABLE "subscriptions" ALTER COLUMN "created_at" DROP DEFAULT ALTER COLUMN "updated_at" DROP DEFAULT;
```

a8m:
Closed with #668 

Thanks for the contribution 🙏

PR #668, opened 2022-03-23
sql/postgres: add missing comma in column changes for reverse changes in Postgres

tprebs:
Adds the missing comma when generating reverse changes on columns.   #667

This change just adds the comma the same way it has been done for other changes and in the  mysql migration but it doesn't feel right to have to manually add the comma in when generating the reverse sql when it is done automatically by `MapComma` for alter table cmd. 

Would it make more sense to either split the logic for reverse out into a second iteration via `reversible.MapComma` create a wrapper struct which exposes both builders to the existing code e.g.  `rb.MapComma(changes, func(i int, b *sqlx.Builder, rb *sqlx.Builder))` This would mean in the future the logic for applying commas is standard between the generation of the cmd and reverse


a8m (approved):
Thanks for the contribution @tprebs.

Regarding your suggestion, I agree it can be improved. I'll try to think of a solution for this.

PR #672, opened 2022-03-30
schema/schemaspec: adding qualifier label to blocks

rotemtam:
adding a "Qualifier" attribute to `schemaspec.Resource`. In HCL can be defined as:

```hcl
resource "qualifier" "name" {

}
```

This is done as infrastructure for HCL document supporting two tables with the same name (but in different schemas).

### DIFF
--- a/schema/schemaspec/spec.go
+++ b/schema/schemaspec/spec.go
@@ -9,10 +9,11 @@ import (
 type (
 	// Resource is a generic container for resources described in configurations.
 	Resource struct {
-		Name     string
-		Type     string
-		Attrs    []*Attr
-		Children []*Resource
+		Name      string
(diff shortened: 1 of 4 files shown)


schema/schemaspec/spec.go
```diff
@@ -9,10 +9,11 @@ import (
 type (
 	// Resource is a generic container for resources described in configurations.
 	Resource struct {
-		Name     string
-		Type     string
-		Attrs    []*Attr
-		Children []*Resource
+		Name      string
+		Qualifier string
```
a8m, line 13: I'm thinking, maybe the qualifiers should be a type `[]string` instead? 
https://mariadb.com/kb/en/identifier-qualifiers/

PR #676, opened 2022-03-31
cmd/action: add MariaDB and PostgreSQL support for docker driver

masseelch:


cmd/action/docker.go
```diff
@@ -124,6 +147,17 @@ func Out(w io.Writer) DockerConfigOption {
 	}
 }
 
+// Setup adds statements to execute once the serviceis ready. For example:
+//
+//  Setup("DROP SCHEMA IF EXISTS public CASCADE;")
+//
+func Setup(s ...string) DockerConfigOption {
+	return func(c *DockerConfig) error {
+		c.Setup = s
```
a8m, line 156: How's consuming this configuration? Or, where/when it's executed?

masseelch: [Here](https://github.com/ariga/atlas/blob/3a6b3bc4e73f4bb0c86356f9d76bcf3098b79f3f/cmd/action/provider.go#L155-L159).

a8m: It feels like the (private) logic of one package resides and is executed by another package. Do you think it's cleaner to drop the schema in the `docker` package?

masseelch: Yeah, I wasn't really sure where I liked it more, in the docker part after the connections can be established safely, or in the provider. But since you raise the same concerns I had, I will move it there :-)

a8m: > [Here](https://github.com/ariga/atlas/blob/3a6b3bc4e73f4bb0c86356f9d76bcf3098b79f3f/cmd/action/provider.go#L155-L159). Is is used to have the database in a "clean" state, which means having no schemas. But the docker image adds the public schema to the default database on startup, rendering the docker "useless" as dev-db.

Yes, understood this, but it feels like the (private) logic of one part is executed by a different part of the code. Don't you think it's cleaner to drop the schema on the container initialization?

a8m (approved):
No code comments besides my design concern.

PR #679, opened 2022-04-03
sql: qualify tables with duplicate names when marshaling

rotemtam:


sql/postgres/sqlspec.go
```diff
@@ -96,6 +96,23 @@ func MarshalSpec(v interface{}, marshaler schemaspec.Marshaler) ([]byte, error)
...
+			}
+			s.Qualifier = schemaName
+			schemaName, err = specutil.SchemaName(tbl.Schema)
+			if err != nil {
+				return nil, err
+			}
+			tbl.Qualifier = schemaName
+		}
+		seen[tbl.Name] = tbl
+	}
```
a8m, line 115: duplicate of th above.

rotemtam: yes sorry, pushed it now

a8m (approved):

PR #681, opened 2022-04-04
sql/migrate: harden atlas.sum file integrity

masseelch:
This PR fixes one bug, where a raised error from `UnmarshalText()` was not propagated up and hardens the integrity of the file by taking the filename of the migrations into the hash as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,11 +12,12 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
+      steps:
+      - uses: actions/setup-go@v2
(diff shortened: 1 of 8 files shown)


a8m:
Ignore the lint, I think upgrading golangci-lint will solve the issue. https://github.com/golangci/golangci-lint/releases/tag/v1.45.2

masseelch:
> Ignore the lint, I think upgrading golangci-lint will solve the issue. https://github.com/golangci/golangci-lint/releases/tag/v1.45.2

It does, but then other issues pop up ...

a8m (approved):

PR #686, opened 2022-04-06
sql/postgres: use native pg time types

a8m:


sql/postgres/sqlspec.go
```diff
@@ -488,3 +452,33 @@ func attr(typ *schemaspec.Type, key string) (*schemaspec.Attr, bool) {
...
+	}
+	s, err := FormatType(t)
+	if err != nil {
+		return nil, err
+	}
+	return &schemaspec.Type{T: s}, nil
+}
+
+// formatTime overrides the default printing logic done by schemahcl.hclType.
+func formatTime() specutil.TypeSpecOption {
```
rotemtam, line 472: Should the driver know about hcl?

sql/postgres/convert.go
```diff
@@ -162,13 +151,13 @@ func ParseType(typ string) (schema.Type, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Normalize PostgreSQL array data types from "CREATE TABLE" format to
+	// Normalize PostgresSQL array data types from "CREATE TABLE" format to
```
masseelch, line 154:
```suggestion
	// Normalize PostgreSQL array data types from "CREATE TABLE" format to
```

a8m: Thanks

sql/postgres/sqlspec.go
masseelch: This is why I added the `Printer` prop you removed in this PR @a8m

a8m: The driver knows nothing about hcl. It's the same schemaspec API that exists today. However, since there is some reflection/parsing logic going on there, I added this comment to make the life easier for future readers. 

I don't think the Printer is needed. From its name, I assumed the printer was used for printing, but that's was not the case (print to driver presentation and then call Parse). Instead, I'd add something like `WithTypeParser` to allow overriding the parsing logic of one typespec if we'll need something like this to avoid one layer of abstraction.

PR #693, opened 2022-04-11
sql/migrate: golang-migrate and goose support

masseelch:
@a8m @rotemtam I am not sure we want to have the tooling support be part of the Core or move it out to a `cli` package, that can be consumed by package users (and our `cmd` package).

### DIFF
--- a/sql/migrate/migrate.go
+++ b/sql/migrate/migrate.go
@@ -389,27 +389,18 @@ var _ Dir = (*LocalDir)(nil)
 var (
 	// templateFuncs contains the template.FuncMap for the DefaultFormatter.
 	templateFuncs = template.FuncMap{
-		"now": time.Now,
-		"sem": ensureSemicolonSuffix,
+		"now": func() string { return time.Now().Format("20060102150405") },
 		"rev": reverse,
 	}
-	// DefaultFormatter is a default implementation for Formatter. Compatible with golang-migrate/migrate.
+	// DefaultFormatter is a default implementation for Formatter.
 	DefaultFormatter = &TemplateFormatter{
 		templates: []struct{ N, C *template.Template }{
 			{
 				N: template.Must(template.New("").Funcs(templateFuncs).Parse(
-					"{{ now.Unix }}{{ with .Name }}_{{ . }}{{ end }}.up.sql",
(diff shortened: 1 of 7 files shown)


sql/migrate/migrate.go
```diff
@@ -389,27 +389,18 @@ var _ Dir = (*LocalDir)(nil)
 var (
 	// templateFuncs contains the template.FuncMap for the DefaultFormatter.
 	templateFuncs = template.FuncMap{
-		"now": time.Now,
-		"sem": ensureSemicolonSuffix,
+		"now": func() string { return time.Now().Format("20060102150405") },
```
hedwigz, line 392: Maybe just introduce a const for this format? so it is more clear and add a small comment

sql/migrate/tooling.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2021-present The Atlas Authors. All rights reserved.
```
a8m, line 1: I think `tools.go` is a better name for this file.

sql/migrate/migrate.go
masseelch: Since it is only used in one place I don't think adding a const is necessary. I added a comment explaining this choice though.

sql/migrate/tooling.go
```diff
@@ -0,0 +1,51 @@
...
+// NewGooseFormatter returns a migrate.Formatter computable with pressly/goose.
+func NewGooseFormatter() (Formatter, error) {
+	return templateFormatter(
+		"{{ now }}{{ with .Name }}_{{ . }}{{ end }}.sql",
+		`-- +goose Up
+{{ range .Changes }}{{ with .Comment }}-- {{ println . }}{{ end }}{{ printf "%s;\n" .Cmd }}{{ end }}
+-- +goose Down
+{{ range rev .Changes }}{{ if .Reverse }}{{ with .Comment }}-- reverse: {{ println . }}{{ end }}{{ printf "%s;\n" .Reverse }}{{ end }}{{ end }}`,
+	)
+}
```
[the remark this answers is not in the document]
masseelch: Moved them to its own package.

a8m (approved):
💯

PR #700, opened 2022-04-13
sql/internal/specutil: support input variables

rotemtam:


sql/internal/lang/vars.go
```diff
@@ -0,0 +1,43 @@
+package lang
```
masseelch, line 1: license

sql/internal/lang/vars.go
```diff
@@ -0,0 +1,43 @@
...
+				specutil.TypeSpec("int"),
+				specutil.TypeSpec("bool"),
+				specutil.TypeSpec("float"),
+			),
+		).Specs(),
+	))
+)
+
+func init() {
+	schemaspec.Register("variable", &Var{})
```
masseelch, line 42: Why not `var`?

masseelch (approved):
:blush:

sql/internal/lang/vars.go
a8m, line 1: We agree to move it `sql/internal/specutil`.

sql/internal/lang/vars.go
a8m: Because `terraform` uses `variables`. Maybe we should accept the 2 forms.

rotemtam: trying to keep things familiar for terraform users:
https://www.terraform.io/language/values/variables#declaring-an-input-variable

a8m (approved):

PR #701, opened 2022-04-13
sql/mysql: support marshaling generated columns

a8m:


sql/mysql/driver.go
```diff
@@ -282,7 +282,7 @@ const (
 	defaultGen    = "default_generated"
 	autoIncrement = "auto_increment"
 
-	genVirtual    = "virtual"
-	genStored     = "stored"
-	genPersistent = "persistent"
+	genVirtual    = "VIRTUAL"
+	genStored     = "STORED"
+	genPersistent = "PERSISTENT"
```
masseelch, line 287: Why `gen` prefix?

sql/mysql/sqlspec.go
```diff
@@ -256,9 +256,28 @@ func columnSpec(c *schema.Column, t *schema.Table) (*sqlspec.Column, error) {
...
+	}
+	spec.Extra.Children = append(spec.Extra.Children, &schemaspec.Resource{
+		Type: "as",
+		Attrs: []*schemaspec.Attr{
+			specutil.VarAttr("type", t),
+			specutil.StrAttr("expr", x.Expr),
+		},
+	})
+}
+
```
masseelch, line 280: Shouldn't the internal representation be the same, regardless of what was defined in HCL? And I think being explicit is cleaner, so if we marshal it wen can always do the `Resource` way.

masseelch (approved):

sql/mysql/sqlspec.go
a8m: Talked to it @rotemtam. I'll move it to the `specutil` after I'll need it in PostgreSQL.

a8m: >  And I think being explicit is cleaner, so if we marshal it wen can always do the Resource way.

Agree.

sql/mysql/driver.go
a8m: No special reason. I love grouping/namespacing constants that are related to each other. I can remove the prefix.

PR #702, opened 2022-04-13
sql/migrate: local files can be created with read permissions for eve…

masseelch:
…ryone, since they are meant to be read by other tools.

### DIFF
--- a/sql/migrate/migrate.go
+++ b/sql/migrate/migrate.go
@@ -381,7 +381,7 @@ func (d *LocalDir) Open(name string) (fs.File, error) {
 
 // WriteFile implements Dir.WriteFile.
 func (d *LocalDir) WriteFile(name string, b []byte) error {
-	return os.WriteFile(filepath.Join(d.dir, name), b, 0600)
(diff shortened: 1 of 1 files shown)


sql/migrate/migrate.go
```diff
@@ -381,7 +381,7 @@ func (d *LocalDir) Open(name string) (fs.File, error) {
 
 // WriteFile implements Dir.WriteFile.
 func (d *LocalDir) WriteFile(name string, b []byte) error {
-	return os.WriteFile(filepath.Join(d.dir, name), b, 0600)
+	return os.WriteFile(filepath.Join(d.dir, name), b, 0644) // nolint: gosec
```
a8m, line 384: move this config to golangci.yml

a8m: Maybe there's a cleaner configuration for it, but here's how I did it in ent https://github.com/ent/ent/blob/master/.golangci.yml#L60-L62

a8m (approved):

PR #704, opened 2022-04-13
internal/types: move typeregistry outside sql

rotemtam:
need to use the type registry more widely (specifically moving the hcl variables code to the hcl package). still don't want to export it so moving to `internal/types` instead.

reducing some stuttering this way as well


internal/types/types.go
```diff
@@ -1,4 +1,4 @@
-package specutil
+package types
```
a8m, line 1: Package `types` is too vague, and I'm not sure it belongs to this namescape. It should reside in the HCL namespace in my opinion, because other languages (e.g. TypeScript) already have a type system. I'd also love to see how we can merge this logic into one of the existing packages because the logic of `schemaspec` spreads across too many atm.

masseelch (approved):

a8m (approved):

PR #706, opened 2022-04-14
internal/integration: initial testscripts for mysql generated columns

a8m:


internal/integration/script_test.go
```diff
@@ -137,14 +137,11 @@ func (t *liteTest) setupScript(env *testscript.Env) error {
...
 func cmdOnly(ts *testscript.TestScript, neg bool, args []string) {
-	if neg {
-		ts.Fatalf("unsupported: ! only")
-	}
 	ver := ts.Getenv("version")
 	for i := range args {
 		re, rerr := regexp.Compile(`(?mi)` + args[i])
 		ts.Check(rerr)
-		if re.MatchString(ver) {
+		if !neg == re.MatchString(ver) {
```
masseelch, line 144: Isn't this the wrong way round :thinking:

masseelch: Got confused. Why not `neg != ...`?

a8m: `only x` means runs only version `x`. `! only x` means runs all, but skip `x`.

yonidavidson (approved):

PR #707, opened 2022-04-14
schema/schemaspec/schemahcl: handle input variables

rotemtam:


schema/schemaspec/schemahcl/context.go
```diff
@@ -40,6 +46,60 @@ func evalCtx(ctx *hcl.EvalContext, f *hcl.File) (*hcl.EvalContext, error) {
...
+//
+// When unmarshaling a document, input values may be provided by using the WithInputValues option
+// upon building the State:
+// state := New(
+// 	 WithInputValues(map[string]interface{}{
+// 		"name": "rotemtam",
+// 		"int":  42,
+// 		"bool": true,
+// 	 }),
+// )
```
a8m, line 64:
```suggestion
// setInputVals sets the input values into the evaluation context. HCL documents can define
// input variables in the document body by defining "variable" blocks:
//
// 	variable "name" {
// 	  type = string // also supported: int, bool
// 	  default = "rotemtam"
// 	}
//
// When unmarshaling a document, input values may be provided by using the WithInputValues
// option upon building the State:
//
// 	state := New(
// 		WithInputValues(map[string]interface{}{
// 			"name": "rotemtam",
// 			"int":  42,
// 			"bool": true,
// 		}),
// 	)
```

PR #714, opened 2022-04-17
sql/postgres: fixed invalid postgres DSN

svstanev:
DSN for Postgres with underscore (_) in the username causes an error: first path segment in URL cannot contain colon. 

Reason: the provider (postgres://) got stripped from the DSN which is then parsed as an URL. If the DSN contains username:password url.Parse either misinterpret the url (if the username contains only letters) or otherwise produces an error "first path segment in URL cannot contain colon".

### DIFF
--- a/cmd/action/mux.go
+++ b/cmd/action/mux.go
@@ -109,9 +109,6 @@ func SchemaNameFromURL(ctx context.Context, url string) (string, error) {
 		}
 		return cfg.DBName, err
 	case "postgres":
-		// Use `url` instead of `dsn` as dsn lacks the scheme part (postgres://).
-		// url.Parse pareses such URLs incorrectly or produce an
-		// error: first path segment in URL cannot contain colon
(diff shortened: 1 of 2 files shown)


cmd/action/mux.go
```diff
@@ -109,7 +109,10 @@ func SchemaNameFromURL(ctx context.Context, url string) (string, error) {
 		}
 		return cfg.DBName, err
 	case "postgres":
-		return postgresSchema(dsn)
+		// Use `url` instead of `dsn` as dsn lacks the scheme part (postgres://).
+		// url.Parse pareses such URLs incorrectly or produce an
+		// error: first path segment in URL cannot contain colon
```
a8m, line 114:
```suggestion
```

No need for this comment. Can reside in the commit message.

a8m (approved):
Thanks for the contribution @svstanev! Please, fix the minor comment and I'll merge this.

PR #715, opened 2022-04-17
sql/postgres: fixed inspection of multicolumn indexes

svstanev:
I have an unique index like this:

```sql
create unique index if not exists u_entity_name on entity (
  entity_type,
  coalesce(parent_id,0),
  name
);
```

which when inspected become like this:

```hcl
index "u_entity_name" {
    unique = true
    type   = BTREE
    on {
      expr = "COALESCE(parent_id, (0)::bigint)"
    }
    on {
      expr = "COALESCE(parent_id, (0)::bigint)"
    }
    on {
      expr = "COALESCE(parent_id, (0)::bigint)"
    }
  }
```

Expected:
```hcl
index "u_entity_name" {
    unique = true
    type   = BTREE
    on {
      column = column.entity_type
    }
    on {
      expr = "COALESCE(parent_id, (0)::bigint)"
    }
    on {
      column = column.name
    }
  }
```

Reason: incorrect index metadata query.

### DIFF
--- a/sql/postgres/inspect.go
+++ b/sql/postgres/inspect.go
@@ -387,16 +387,16 @@ func (i *inspect) addIndexes(s *schema.Schema, rows *sql.Rows) error {
 			})
 		}
 		switch {
-		case sqlx.ValidString(expr):
-			part.X = &schema.RawExpr{
-				X: expr.String,
-			}
 		case sqlx.ValidString(column):
 			part.C, ok = t.Column(column.String)
 			if !ok {
 				return fmt.Errorf("postgres: column %q was not found for index %q", column.String, idx.Name)
 			}
 			part.C.Indexes = append(part.C.Indexes, idx)
+		case sqlx.ValidString(expr):
+			part.X = &schema.RawExpr{
+				X: expr.String,
+			}
 		default:
 			return fmt.Errorf("postgres: invalid part for index %q", idx.Name)
 		}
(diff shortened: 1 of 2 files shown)


sql/postgres/inspect.go
```diff
@@ -387,18 +387,17 @@ func (i *inspect) addIndexes(s *schema.Schema, rows *sql.Rows) error {
 				NullsLast:  nullslast.Bool,
 			})
 		}
-		switch {
-		case sqlx.ValidString(expr):
-			part.X = &schema.RawExpr{
-				X: expr.String,
-			}
-		case sqlx.ValidString(column):
```
a8m, line 395: Why this change is needed?

sql/postgres/inspect.go
```diff
@@ -777,25 +776,31 @@ SELECT
...
 	obj_description(to_regclass($1 || i.relname)::oid) AS comment
 FROM
-	pg_index idx
+	(
+		select
+			*,
+			generate_series(1,array_length(i.indkey,1)) as ord,
+			unnest(i.indkey) AS key
+		from pg_index i
+	) idx
```
a8m, line 791: 💯

sql/postgres/inspect.go
svstanev: Each row of the indexesQuery returns an index part/column in a multicolumn index. As per the suggested query each part has a valid expression but column only if it is created with a column name:

```sql
LEFT JOIN pg_attribute a ON (a.attrelid, a.attnum) = (idx.indrelid, idx.key)
```

a8m: Thanks for the explanation. If you don't mind, please use revert to the switch statement and I'll merge it.

a8m (approved):
💯

a8m:
Thanks for the contribution @svstanev 🙏🏻

PR #718, opened 2022-04-17
sql/postgres: marshal spec duplicate enum entries

svstanev:
When marshalling 2 or more columns that reference the same enum type the generated Atlas DDL document contains duplicate entries for this enum (one entry per column).

### DIFF
--- a/sql/postgres/sqlspec.go
+++ b/sql/postgres/sqlspec.go
@@ -290,14 +290,20 @@ func schemaSpec(schem *schema.Schema) (*doc, error) {
 	}
 	d.Schemas = []*sqlspec.Schema{s}
 	d.Tables = tbls
+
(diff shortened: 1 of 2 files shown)


sql/postgres/sqlspec.go
```diff
@@ -290,14 +290,20 @@ func schemaSpec(schem *schema.Schema) (*doc, error) {
 	}
 	d.Schemas = []*sqlspec.Schema{s}
 	d.Tables = tbls
+
+	added := make(map[string]struct{})
+
```
a8m, line 295:
```suggestion
	enums := make(map[string]struct{})
```

No need for extra new lines.

a8m (approved):
Looks good, besides one small style comment. 


Thanks for the contribution! 🙏

PR #720, opened 2022-04-18
schema/schemaspec: support defining resource names as attr

rotemtam:
Needed in order to support input variables for schema names: 

```hcl
variable "tenant" {
   type = string
}

schema "tenant" {
  name = var.tenant
}

table "users" {
  schema = schema.tenant
}
```

### DIFF
--- a/schema/schemaspec/extension_test.go
+++ b/schema/schemaspec/extension_test.go
@@ -194,3 +194,20 @@ func TestListRef(t *testing.T) {
 	require.NoError(t, err)
 	require.EqualValues(t, resource, scan)
 }
+
+func TestNameAttr(t *testing.T) {
+	type Named struct {
+		Name string `spec:"name,name"`
+	}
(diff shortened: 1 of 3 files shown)


schema/schemaspec/extension_test.go
```diff
@@ -194,3 +194,20 @@ func TestListRef(t *testing.T) {
...
+		Name string `spec:"name,name"`
+	}
+	schemaspec.Register("named", &Named{})
+	resource := &schemaspec.Resource{
+		Name: "id",
+		Attrs: []*schemaspec.Attr{
+			schemautil.StrLitAttr("name", "atlas"),
+		},
+	}
+	tgt := Named{}
```
a8m, line 209:
```suggestion
	var tgt Named
```

a8m (approved):

masseelch (approved):

PR #721, opened 2022-04-18
sql/postgres: apply generate columns for new columns

a8m:


sql/postgres/diff.go
```diff
@@ -171,7 +171,15 @@ func (d *diff) typeChanged(from, to *schema.Column) (bool, error) {
 	case *schema.BinaryType, *schema.BoolType, *schema.DecimalType, *schema.FloatType,
 		*schema.IntegerType, *schema.JSONType, *schema.SpatialType, *schema.StringType,
 		*schema.TimeType, *BitType, *NetworkType, *UserDefinedType:
-		changed = mustFormat(toT) != mustFormat(fromT)
+		t1, err1 := FormatType(toT)
```
zeevmoney, line 174: Why do you need err1 & err2? can't you use just `err`?

zeevmoney (approved):
lgtm, see comment.

rotemtam (approved):

sql/postgres/diff.go
a8m: I can. I wrote it differently and then changed it.

Will fix, thanks

PR #725, opened 2022-04-19
cmd/actions: refactor actions api

hedwigz:
* Moved actions to public API.
* Moved mux+provider to SQL core
* Moved docker to internal/docker


sql/sqltool/tool_test.go
```diff
@@ -12,6 +12,7 @@ import (
 
 	"ariga.io/atlas/sql/migrate"
 	"ariga.io/atlas/sql/sqltool"
+
```
masseelch, line 15: ?

cmd/action/apply.go
```diff
@@ -10,6 +10,8 @@ import (
 	"io/ioutil"
 	"strings"
 
+	"ariga.io/atlas/sql"
+
```
masseelch, line 14: ?

internal/docker/docker.go
```diff
@@ -40,22 +40,22 @@ type (
 	}
 	// A Container is an instance of a created container.
 	Container struct {
-		cfg DockerConfig // DockerConfig used to create this container
-		out io.Writer    // custom write to log status messages to
+		cfg Config    // Config used to create this container
```
masseelch, line 43: :+1:

api.go
```diff
@@ -0,0 +1,89 @@
+package atlas
```
masseelch, line 1: Should be moved to SQL as well I think

masseelch (commented):
Good start. I think we can get rid of the `sql.Driver` (previouyl `action.Driver`) struct as well, can't we?

PR #730, opened 2022-04-22
sql/migrate: interfaces for migration execution

masseelch:
This PR introduces interfaces and first implementation for executing migration files on a database.

Extensive tests will come, when the `Scanner` interface has a default implementation. `StateReadWriter` implementations will be provided by the different Drivers.


sql/migrate/migrate.go
```diff
@@ -159,6 +159,23 @@ type (
...
+	// DirUnraveler wraps the UnravelDir method.
+	DirUnraveler interface {
+		// UnravelDir returns a set of files from the given Dir to be executed on a database.
+		UnravelDir(Dir) []File
+	}
+
+	// FileUnraveler wraps the UnravelFile method.
+	FileUnraveler interface {
+		UnravelFile(File) []string
+	}
```
a8m, line 171: Do you mind explaining why these 2 interfaces are needed? Is it here because you have a WIP code or does it help organizing your thoughts for now? 

In general, it's hard (for me) to see the benefit of interfaces without seeing how they are used (consumed or implemented). 

Last comment about this - Why did you choose the term/word Unraveler? it feels less common and not intuitive for usage (and verbal discussions), but again, it's hard for me to judge without seeing implementations or usage of it.

sql/migrate/migrate.go
```diff
@@ -171,6 +188,29 @@ type (
...
+		// Hash is the check-sum of this migration as stated by the migration directories HashFile.
+		Hash string
+		// OperatorVersion holds a string representation of the Atlas operator managing this database migration.
+		OperatorVersion string
+		// Meta holds additional custom meta-data given for this migration.
+		Meta map[string]string
+	}
+
+	// State is an ordered set of Migration structs.
+	State []Migration
```
a8m, line 213: 3 comments about the naming here. It feels better for me to name the `migrate.Migration` struct above as `migrate.Revision`, the `State` below as something like `History` or `Log`, and the container above as follows:

```go
type HistoryReadWriter interface {
	ReadHisotry() (History, error)
	WriteRevision(Revision) error
}
```

a8m: Rethinking about this, the `History` (revision list) feels redundant. You can also write the above interface as follows:

```go
type RevisionReadWriter interface {
	ReadRevisions() ([]Revision, error)
	WriteRevision(Revision) error
}
```

masseelch: I like `Revision` :-)

sql/migrate/migrate.go
```diff
@@ -171,6 +190,40 @@ type (
...
+	// Revisions is an ordered set of Revision structs.
+	Revisions []*Revision
+
+	// Executor is responsible to manage and execute a set of migration files against a database.
+	Executor struct {
+		drv Driver // the Driver to access and manage the database
+		dir Dir    // the Dir with migration files to use
+	}
+
+	// ExecutorOption allows managing an  Executor using functional arguments.
```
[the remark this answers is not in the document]
a8m: Is it needed? I don't see any option.

sql/migrate/migrate.go
```diff
@@ -209,11 +262,156 @@ func DisableChecksum() PlannerOption {
...
+		if err := opt(p); err != nil {
+			return nil, err
+		}
+	}
+	return p, nil
+}
+
+// Execute executes n missing migration files on the database
+func (e *Executor) Execute(ctx context.Context, n int) error {
+	// Don't operate with a broken migration directory. TODO(maseeelch): do not check here but let the caller check it before? Or let caller decide if to skip this flag by using some flags.
```
a8m, line 324:
```suggestion
	// Don't operate with a broken migration directory.
	// TODO(maseeelch): do not check here but let the caller check it before? Or let the caller decide if to skip this flag by using some flags.
```

sql/migrate/migrate.go
```diff
@@ -209,11 +262,156 @@ func DisableChecksum() PlannerOption {
...
+	if err := Validate(e.dir); err != nil {
+		return fmt.Errorf("sql/migrate: execute: validate migration directory: %w", err)
+	}
+	// Check if the Driver implements RevisionReadWriter interface.
+	rrw, ok := e.drv.(RevisionReadWriter)
+	if !ok {
+		return fmt.Errorf("sql/migrate: execute: no revisions reader available")
```
a8m, line 331:
```suggestion
		return errors.New("sql/migrate: execute: no revisions reader available")
```

sql/migrate/migrate.go
```diff
@@ -209,11 +262,156 @@ func DisableChecksum() PlannerOption {
...
+	}
+	// Check if the Dir implements Scanner interface.
+	sc, ok := e.dir.(Scanner)
+	if !ok {
+		return fmt.Errorf("sql/migrate: execute: no scanner available")
```
a8m, line 336:
```suggestion
		return errors.New("sql/migrate: execute: no scanner available")
```

sql/migrate/migrate.go
```diff
@@ -209,11 +262,156 @@ func DisableChecksum() PlannerOption {
...
+		return fmt.Errorf("sql/migrate: execute: select migration files: %w", err)
+	}
+	// Check if the existing revisions did come from the migration directory and not from somewhere else.
+	hf, err := readHashFile(e.dir)
+	if err != nil {
+		return fmt.Errorf("sql/migrate: execute: read atlas.sum file: %w", err)
+	}
+	// For up to len(revisions) revisions the migration files must both match in order and content.
+	if len(revisions) > len(migrations) {
+		return fmt.Errorf("sql/migrate: execute: revisions and migrations mismatch: more revisions than migrations")
```
a8m, line 355: ditto

a8m: Take in mind we may provide compaction support in the future.

sql/migrate/migrate.go
```diff
@@ -209,11 +262,156 @@ func DisableChecksum() PlannerOption {
...
+		if err != nil {
+			return fmt.Errorf("sql/migrate: execute: scan description from %q: %w", m.Name(), err)
+		}
+		r := revisions[i]
+		if v != r.Version || d != r.Description || hf[i].H != r.Hash { // TODO(masseelch): version / desc check necessary?
+			return fmt.Errorf("sql/migrate: execute: revisions and migrations mismatch: rev %q <> file %q", v, r.Version)
+		}
+	}
+	if len(migrations) == len(revisions) {
+		return fmt.Errorf("sql/migrate: execute: nothing to do")
```
a8m, line 373: I think `migrate apply` command should not return an error in case there's nothing to apply.

sql/migrate/migrate.go
```diff
@@ -209,11 +262,156 @@ func DisableChecksum() PlannerOption {
...
+	for _, opt := range opts {
+		if err := opt(p); err != nil {
+			return nil, err
+		}
+	}
+	return p, nil
+}
+
+// Execute executes n missing migration files on the database
+func (e *Executor) Execute(ctx context.Context, n int) error {
```
a8m, line 323: In the case of `migrate apply`, I'd expect that Atlas will apply all migrations. Do we want a custom method for this, or do you want to accept -1 as a parameter to indicate "all"?

sql/migrate/migrate.go
```diff
@@ -209,11 +262,156 @@ func DisableChecksum() PlannerOption {
...
+			return fmt.Errorf("sql/migrate: execute: revisions and migrations mismatch: rev %q <> file %q", v, r.Version)
+		}
+	}
+	if len(migrations) == len(revisions) {
+		return fmt.Errorf("sql/migrate: execute: nothing to do")
+	}
+	defer rrw.WriteRevisions(ctx, revisions) // TODO:(masseelch): handle error
+	// TODO(masseelch): run in a transaction
+	for i := len(revisions); i < len(revisions)+n; i++ {
+		var m = migrations[i]
```
a8m, line 378:
```suggestion
		m := migrations[i]
```

Why var?

a8m (approved):

sql/migrate/migrate.go
```diff
@@ -171,6 +190,40 @@ type (
...
+		// Description of this migration.
+		Description string
+		// ExecutionState of this migration. One of ["ongoing", "ok", "error"].
+		ExecutionState string
+		// ExecutedAt denotes when this migration was started to be executed.
+		ExecutedAt time.Time
+		// ExecutionTime denotes the time it took for this migration to be applied on the database.
+		ExecutionTime time.Duration
+		// Error holds information about a migration error (if occurred).
+		Error string
```
a8m, line 207: Can you document here the error formats you use in the code (e.g. "Go", "Statement").

sql/migrate/migrate.go
masseelch: My plan is to accept `-1` as this is common in other implementations (e.g. `strings.SplitN()`

PR #732, opened 2022-04-24
sql/schema: add helper methods for working with changesets

a8m:
I'm not super enthusiastic about this API, but we should make users' life easier when working with changes.


rotemtam (approved):
💯

PR #745, opened 2022-04-28
sql/sqlclient: use schemahacl.Eval

rotemtam:


cmd/action/apply.go
```diff
@@ -98,7 +98,7 @@ func applyRun(ctx context.Context, client *sqlclient.Client, url string, file st
 	f, err := ioutil.ReadFile(file)
 	cobra.CheckErr(err)
 	desired := &schema.Realm{}
-	cobra.CheckErr(client.UnmarshalSpec(f, desired))
+	cobra.CheckErr(client.Eval(f, desired, nil))
```
rotemtam, line 101: will change the signature to make this optional

a8m (approved):

PR #749, opened 2022-04-28
cmd/internal: move docker client to internal package

a8m:


cmd/internal/docker/docker.go
```diff
@@ -213,20 +211,20 @@ func (c *Container) Wait(ctx context.Context, timeout time.Duration) error {
...
 	for {
 		select {
 		case <-time.After(100 * time.Millisecond):
-			db, err := sql.Open(drv, dsn)
+			client, err := sqlclient.Open(ctx, u)
 			if err != nil {
-				return err
+				continue
 			}
+			db := client.DB
```
masseelch, line 227: Why did you change this from the `database/sql` to `sqlclient`? It is only meant to check for availablitlity. Also, you need to close the `sqlclient`, correct?

cmd/internal/docker/docker.go
```diff
@@ -325,6 +299,11 @@ func dockerClient(ctx context.Context, u *url.URL) (client *sqlclient.Client, er
 	if err != nil {
 		return nil, err
 	}
+	if u.Query().Has("v") || u.Query().Has("verbose") {
+		if err := Out(os.Stdout)(cfg); err != nil {
+			return nil, err
+		}
+	}
```
masseelch, line 306: :heart_eyes:

masseelch (approved):
:100:

cmd/internal/docker/docker.go
a8m: Because I deleted the DSN function, as it's implemented by the drivers.

I'm closing the DB, but will switch to closing the client.

a8m: Rethinking it. No need to close the client, closing the underlying DB is fine.

cmd/internal/docker/docker.go
```diff
@@ -307,13 +281,13 @@ func freePort() (string, error) {
...
+	sqlclient.Register("docker", sqlclient.OpenerFunc(client))
 }
 
-func dockerClient(ctx context.Context, u *url.URL) (client *sqlclient.Client, err error) {
-	var cfg *DockerConfig
+func client(ctx context.Context, u *url.URL) (client *sqlclient.Client, err error) {
+	var cfg *Config
 	switch img, tag := u.Host, strings.TrimPrefix(u.Path, "/"); img {
-	case "mysql", "tidb":
+	case "mysql":
```
rotemtam, line 290: @a8m i suppose the drivers should later register this ?

PR #750, opened 2022-04-28
sql/mysql: Ent based RevisionReadWriter implementation for mysql

masseelch:


sql/migrate/migrate.go
```diff
@@ -379,17 +380,42 @@ func (e *Executor) Execute(ctx context.Context, n int) error {
...
 	}
-	defer rrw.WriteRevisions(ctx, revisions) // TODO:(masseelch): handle error
+	defer func(rrw RevisionReadWriter, ctx context.Context, revisions *Revisions) {
+		if dErr := rrw.WriteRevisions(ctx, *revisions); dErr != nil {
+			if err != nil {
+				err = fmt.Errorf("sql/migrate: execute: write revisions: %w: %v", dErr, err)
+				return
+			}
+			err = dErr
+		}
```
a8m, line 390:
```suggestion
		if werr := rrw.WriteRevisions(ctx, *revisions); werr != nil {
			if err != nil {
				werr = fmt.Errorf("sql/migrate: execute: write revisions: %w: %v", err, werr)
			}
			err = werr
		}
```

sql/internal/migrate/client.go
```diff
@@ -0,0 +1,212 @@
+// Copyright 2021-present The Atlas Authors. All rights reserved.
+// This source code is licensed under the Apache 2.0 license found
+// in the LICENSE file in the root directory of this source tree.
+
+// Code generated by entc, DO NOT EDIT.
+
+package migrate
```
a8m, line 7: Can you please use the standard ent naming? You can place it also under the `sql/migrate` package.

sql/internal/migrate/entc.go
```diff
@@ -0,0 +1,70 @@
...
+func (c *Client) init() {
+	{{- range $n := $.Nodes }}
+    	c.{{ $n.Name }} = New{{ $n.Name }}Client(c.config)
+	{{- end }}
+}`
+	testTpl = `{{ with extend $ "Package" "enttest" -}}
+	{{ template "header" . }}
+{{ end }}
+
+// Overridden by a custom template because it is unused in this project.`
```
a8m, line 50: What do you think about creating a feature-flag for enttest (that is enabled by default) and disabling it here? I can do it.

sql/mysql/runtime/runtime.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// InitEntSchemaMigrator stitches in the Ent migration engine to the mysql.Driver at runtime. This is necessary
+// because the Ent migration engine imports atlas and therefore would introduce a cyclic dependency.
+func InitEntSchemaMigrator(drv *mysql.Driver, db *sql.DB, dialect string) {
+	mgr := migrate.NewSchema(entsql.NewDriver(entsql.Conn{ExecQuerier: db}, dialect))
+	drv.InitSchemaMigrator(func(ctx context.Context) error {
+		return mgr.Create(ctx, entschema.WithAtlas(true))
+	})
+}
```
a8m, line 20: I'd put it under `ent/runtime` with all others dialects and register it in init.

sql/mysql/migrate.go
```diff
@@ -52,6 +54,69 @@ func (p *planApply) ApplyChanges(ctx context.Context, changes []schema.Change) e
...
+	if err != nil {
+		return nil, err
+	}
+	if !rs.Next() {
+		// If there are no results the table does not exist and there are no revisions yet.
+		return migrate.Revisions{}, nil
+	}
+	if err := rs.Close(); err != nil {
+		return nil, err
+	}
```
a8m, line 77: Not sure why this behavior is needed and acceptable. Can you please elaborate?

sql/mysql/migrate.go
```diff
@@ -52,6 +54,69 @@ func (p *planApply) ApplyChanges(ctx context.Context, changes []schema.Change) e
...
+	revs, err := r.c.Revision.Query().All(ctx)
```
a8m, line 78: I think we need to add some ORDERing here. There's no default ordering in the database.

a8m:
💯 

Code looks really good but added a few questions/comments.

sql/mysql/migrate.go
masseelch: For a fresh DB e.g.

sql/mysql/migrate.go
masseelch: Well, there is: order of insertion. But to be safe we can add one.

sql/mysql/runtime/runtime.go
masseelch: If you know a way to register it inside init on a `migrate.Driver`, yes. 

It can't be a global var since we have multiple instances of drivers for different versions at runtime (e.g. in tests)

sql/mysql/migrate.go
a8m: There’s not guarantee that rows will return some order unless the ORDER BY clause is used.

a8m: There’s no guarantee that rows will return in some order unless the ORDER BY clause is used.

a8m (approved):

PR #752, opened 2022-04-29
sql/postgres: ent based revisions for postgres

masseelch:


sql/internal/sqlx/migrate.go
```diff
@@ -0,0 +1,84 @@
...
+			SetExecutedAt(rev.ExecutedAt).
+			SetExecutionTime(rev.ExecutionTime).
+			SetHash(rev.Hash).
+			SetOperatorVersion(rev.OperatorVersion).
+			SetMeta(rev.Meta)
+	}
+	return r.c.Revision.CreateBulk(bulk...).
+		OnConflict(
+			sql.ConflictColumns(revision.FieldID),
+		).
```
a8m, line 79: Why do you expect conflict? In case of retrying?

sql/internal/sqlx/migrate.go
```diff
@@ -0,0 +1,84 @@
...
+	c *ent.Client
+	// sc is the function signature of the Ent migration engine.
+	// Due to cyclic dependencies between Ent and Atlas, this value is stitched in at runtime.
+	sc func(context.Context) error
+}
+
+// NewRevisionStorage creates a new RevisionStorage with the given ent.Client.
+func NewRevisionStorage(c *ent.Client) *RevisionStorage {
+	return &RevisionStorage{c: c}
+}
```
a8m, line 27:
```suggestion
func NewRevisionStorage(dialect string, db schema.ExecQuerier) *RevisionStorage {
	return &RevisionStorage{
		c: ent.NewClient(ent.Driver(sql.NewDriver(sql.Conn{ExecQuerier: db}, dialect))),
	}
}
```

Then, in drivers, call:

```go
sqlx.NewRevisionStorage(dialect.Postgres, db)
sqlx.NewRevisionStorage(dialect.MySQL, db)
```

sql/internal/sqlx/migrate.go
```diff
@@ -0,0 +1,84 @@
...
+	"context"
+
+	"ariga.io/atlas/sql/migrate"
+	"ariga.io/atlas/sql/migrate/ent"
+	"ariga.io/atlas/sql/migrate/ent/revision"
+	"entgo.io/ent/dialect/sql"
+)
+
+// A RevisionStorage provides implementation for the migrate.RevisionReadWriter interface.
+type RevisionStorage struct {
```
a8m, line 17:
```suggestion
type EntRevision struct {
```

sounds better IMO for an ent-based implementation.

sql/internal/sqlx/migrate.go
masseelch: I always do an insert and when there already is data I replace it.

a8m: Something I don't understand about the product flow. In which case one execution overrides the previous, on retrying? Don't you think it makes sense to store the previous execution information as well? I'd expect to see this in the UI as well.

@rotemtam, thoughts?

a8m (approved):

PR #754, opened 2022-04-30
schema/schemaspec/schemahcl: patch references to blocks with variable names

rotemtam:
Fixes a bug in cases like:
```hcl
variable "tenant" {
	type = string
	default = "test"
}
schema "tenant" {
	name = var.tenant
}
table "users" {
	schema = schema.tenant
	column "id" {
		type = int
	}
}
```
When provided an input var such as `tenant=rotemtam`, the value of the reference to `schema.tenant` was `$schema.tenant` instead of `$schema.rotemtam`

### DIFF
--- a/schema/schemaspec/extension.go
+++ b/schema/schemaspec/extension.go
@@ -200,6 +200,30 @@ func (r *Resource) As(target interface{}) error {
 	return nil
 }
 
+// FinalName returns the final name for the resource by examining the struct tags for
+// the extension of the Resource's type. If no such extension is registered or the
+// extension struct does not have a name field, an error is returned.
+func (r *Resource) FinalName() (string, error) {
+	extensionsMu.RLock()
+	defer extensionsMu.RUnlock()
+	t, ok := extensions[r.Type]
+	if !ok {
+		return "", fmt.Errorf("no extension registered for %q", r.Type)
+	}
+	for _, fd := range specFields(t) {
+		if fd.isName() {
+			if fd.tag != "" {
+				name, ok := r.Attr(fd.tag)
+				if ok {
+					return name.String()
+				}
+			}
+			return r.Name, nil
+		}
+	}
+	return "", fmt.Errorf("extension %q has no name field", r.Type)
+}
+
 func validateStructPtr(target interface{}) error {
 	typeOf := reflect.TypeOf(target)
 	if typeOf.Kind() != reflect.Ptr {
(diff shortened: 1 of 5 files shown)


schema/schemaspec/extension.go
```diff
@@ -200,6 +200,30 @@ func (r *Resource) As(target interface{}) error {
 	return nil
 }
 
+// FinalName returns the final name for the resource by examining the struct tags for
+// the extension of the Resource's type. If no such extension is registered or the
+// extension struct does not have a name field, an error is returned.
+func (r *Resource) FinalName() (string, error) {
+	extensionsMu.RLock()
+	defer extensionsMu.RUnlock()
+	t, ok := extensions[r.Type]
```
a8m, line 209: Feels not right to me to synchronize codec flow. Don't you have this info on the Resource itself?

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -72,12 +72,71 @@ func (s *State) Eval(data []byte, v interface{}, input map[string]string) error
 	if err != nil {
 		return fmt.Errorf("schemahcl: failed decoding: %w", err)
 	}
+	if err := patchRefs(spec); err != nil {
+		return err
+	}
```
a8m, line 77: I don't see any test case for this. Is it intentional?

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -72,12 +72,71 @@ func (s *State) Eval(data []byte, v interface{}, input map[string]string) error
...
 	if err := spec.As(v); err != nil {
 		return fmt.Errorf("schemahcl: failed reading spec as %T: %w", v, err)
 	}
 	return nil
 }
 
+// refMap maps addresses to their referenced resource.
+type refMap map[string]*schemaspec.Resource
```
a8m, line 85: Don't add the type to the name, use something like `addrRef` instead.

schema/schemaspec/extension.go
rotemtam: The lock protects concurrent reading and writing extensions (schemaspec.Register).
I did something similar to:
https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/database/sql/sql.go;l=817

schema/schemaspec/schemahcl/hcl.go
rotemtam: it's tested in the drivers, but I will add a specific case in the `schemahcl` package, you're right.

schema/schemaspec/extension.go
rotemtam: discussed offlien, will remove this in a later pr

a8m (approved):

PR #761, opened 2022-05-03
cmd/atlascmd: add support for input variables in apply

rotemtam:
adds support for input variables in `atlas schema apply`

Given:
```hcl
variable "tenant" {
	type = string
}
schema "tenant" {
	name = var.tenant
}
table "users" {
	schema = schema.tenant
	column "id" {
		type = int
	}
}
```

Running:
```shell
atlas schema apply -u ... -f ... --var tenant=rotemtam
```
Will replace the variable `tenant` with the value `rotemtam`

### DIFF
--- a/schema/schemaspec/extension.go
+++ b/schema/schemaspec/extension.go
@@ -200,6 +200,30 @@ func (r *Resource) As(target interface{}) error {
 	return nil
 }
 
+// FinalName returns the final name for the resource by examining the struct tags for
+// the extension of the Resource's type. If no such extension is registered or the
+// extension struct does not have a name field, an error is returned.
+func (r *Resource) FinalName() (string, error) {
+	extensionsMu.RLock()
+	defer extensionsMu.RUnlock()
+	t, ok := extensions[r.Type]
+	if !ok {
+		return "", fmt.Errorf("no extension registered for %q", r.Type)
+	}
+	for _, fd := range specFields(t) {
+		if fd.isName() {
+			if fd.tag != "" {
+				name, ok := r.Attr(fd.tag)
+				if ok {
+					return name.String()
+				}
+			}
+			return r.Name, nil
+		}
+	}
+	return "", fmt.Errorf("extension %q has no name field", r.Type)
+}
+
 func validateStructPtr(target interface{}) error {
 	typeOf := reflect.TypeOf(target)
 	if typeOf.Kind() != reflect.Ptr {
--- a/schema/schemaspec/extension_test.go
+++ b/schema/schemaspec/extension_test.go
@@ -131,6 +131,9 @@ func TestNested(t *testing.T) {
 	err = scan.Scan(&pb)
 	require.NoError(t, err)
 	require.EqualValues(t, pet, scan)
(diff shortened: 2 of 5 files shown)


cmd/atlascmd/schema.go
```diff
@@ -334,3 +339,15 @@ func fmtFile(task fmttask) (bool, error) {
 	}
 	return false, nil
 }
+
+func argMap(s []string) (map[string]string, error) {
```
a8m, line 343:
```suggestion
func parseVars(s []string) (map[string]string, error) {
```

a8m (approved):
💯 

Excellent work! Please, consider dropping arguments in `applyRun` and access ApplyFlags directly.

cmd/atlascmd/schema.go
```diff
@@ -334,3 +339,15 @@ func fmtFile(task fmttask) (bool, error) {
 	}
 	return false, nil
 }
+
```
masseelch, line 342: Why are you not using https://pkg.go.dev/github.com/spf13/pflag?utm_source=godoc#StringToStringVarP?

cmd/atlascmd/schema.go
```diff
@@ -167,14 +170,16 @@ func CmdInspectRun(cmd *cobra.Command, _ []string) {
 
 // CmdApplyRun is the command used when running CLI.
 func CmdApplyRun(cmd *cobra.Command, _ []string) {
+	args, err := argMap(ApplyFlags.Vars)
+	cobra.CheckErr(err)
```
masseelch, line 174: See my other comment, this is obsolete then and you can pass in the flag directly to `applyRun`.

masseelch (approved):
Nicely done :-)

cmd/atlascmd/schema.go
```diff
@@ -334,3 +339,15 @@ func fmtFile(task fmttask) (bool, error) {
 	}
 	return false, nil
 }
+
+func argMap(s []string) (map[string]string, error) {
+	out := make(map[string]string, len(s))
+	for _, i := range s {
+		parts := strings.Split(i, "=")
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("expected key=value for arg, got %q", i)
```
a8m, line 348:
```suggestion
			return nil, fmt.Errorf("expected key=value for var, got %q", i)
```

cmd/atlascmd/schema.go
```diff
@@ -334,3 +339,15 @@ func fmtFile(task fmttask) (bool, error) {
 	}
 	return false, nil
 }
+
+func argMap(s []string) (map[string]string, error) {
+	out := make(map[string]string, len(s))
```
a8m, line 344:
```suggestion
	vars := make(map[string]string, len(s))
```

masseelch, line 344:
```suggestion
	vars := make(map[string]string, len(s))
```

cmd/atlascmd/schema.go
rotemtam: nice

PR #763, opened 2022-05-03
cmd/atlascmd: add support for input vars on apply

rotemtam:
adds support for input variables in `atlas schema apply`

Given:
```hcl
variable "tenant" {
	type = string
}
schema "tenant" {
	name = var.tenant
}
table "users" {
	schema = schema.tenant
	column "id" {
		type = int
	}
}
```

Running:
```shell
atlas schema apply -u ... -f ... --var tenant=rotemtam
```
Will replace the variable `tenant` with the value `rotemtam`


masseelch (approved):

doc/md/cli/reference.md
```diff
@@ -177,17 +177,18 @@ migration.
...
+                             [driver://username:password@address/dbname?param=value]
+  -s, --schema strings       Set schema names.
+      --dev-url string       URL for the dev database. Used to validate schemas and calculate diffs
+                             before running migration.
+      --dry-run              Dry-run. Print SQL plan without prompting for execution.
+      --auto-approve         Auto approve. Apply the schema changes without prompting for approval.
+  -w, --web                  Open in a local Atlas UI.
+      --addr string          used with -w, local address to bind the server to. (default ":5800")
+      --verbose              enable verbose logging
+      --var stringToString   input variables (default [])
```
a8m, line 191:
```suggestion
      -var 'foo=bar'   input variables (default [])
```

I like terraform doc example. 

image

PR #764, opened 2022-05-03
doc: document input variables usage

rotemtam:


cmd/atlascmd/schema.go
```diff
@@ -110,7 +110,7 @@ files in the directory are formatted, no input will be printed out.
 )
 
 const (
-	answerApply = "SchemaApply"
+	answerApply = "Apply"
```
rotemtam, line 113: not related, just noticed this while testing

a8m (approved):

hedwigz (approved):
:rocket:

PR #769, opened 2022-05-04
sql/postgres: support unmarshaling partitions

a8m:


sql/postgres/sqlspec.go
```diff
@@ -127,7 +128,70 @@ var (
...
+		return nil
+	}
+	var p struct {
+		Type    string            `spec:"type"`
+		Columns []*schemaspec.Ref `spec:"columns"`
+		Parts   []*struct {
+			Expr   string          `spec:"expr"`
+			Column *schemaspec.Ref `spec:"column"`
+		} `spec:"by"`
+	}
```
rotemtam, line 154: you can do it in a subsequent PR but since we will want to marshal this to HCL the type should be defined outside the package

rotemtam (approved):

sql/postgres/sqlspec.go
a8m: 👍

Issue #709, opened 2022-04-15
proposal: support workspaces/environments 


a8m:
# Atlas Workspaces _(initial version)_

Hey all! We just started to work on the initial version of Atlas workspaces and I wanted to summarize it here and hear your thoughts about it before we have major progress on it. 

## Motivation

There are a few limitation today in Atlas that we want to address:
- Support maintaining an "Atlas schema" on multiple files 
- Allow applying the same "Atlas schema" on multiple environments. e.g. multi-tenant deployment or multiple environments. 
- Support input variables

## Proposal

Make it possible to group one or more HCL files as an "Atlas schema" and allow interact with them as they were a single entity. An "Atlas schema" represents one or more database schemas (named-databases), that may be logically connected through foreign-keys and can be managed with the same connection (established with the provided `url` option).

> Note: In the future, we would like to add support to manage multiple databases in the same project. This is the first stage to get there, and we'll publish a proper design for it.

An "Atlas schema" that spans on one or more files and uses the new format can be used as a "schema template" (~similar to terraform modules). That is, it accepts input variables and allow to use them in the various resources. A rough example:

```hcl
variable "tenant_name" {
  type = string
}

schema "tenant" {
  name = var.tenant_name
}

table "users" {
  schema = schema.tenant
  ...
}

view "t" {
  schema = schema.tenant
  as     = file("t.sql")
}
```

We also introduce a new file named `atlas.hcl` that holds a **global** configuration of the project, e.g. your credentials to GCP or AWS, and a set of blocks describe the  **environments** (or workspaces) the different schemas are applied on. The structure is not final, but here's a taste of what we thought about.

1\. Allow working with the same schema on multiple environments

```hcl
aws {
  credentials = "..."
}

env "dev" {
  url     = "<dev-url>"
  dev     = "<ariga-url>" 
  schema  = "./dir"
}

env "prod" {
  url    = "<prod-url>"
  schema = "./dir"
}
```

Then, the interaction with the database will look as follows:

```console
atlas schema inspect dev
atlas migrate apply prod
```

2\. Accept input variables in atlas.hcl

```hcl
variable "tenant" {
  type = string
}

variable "user" {
  type    = string
  default = "root"
}

variable "password" {
  type = string
}

env "project1" "prod" {
  url     = "mysql://root:${var.password}..."
  schema  = "./dir1"
  
  // Inject input variables.
  tenant_name = var.tenant
}

env "project2" "prod" {
  url     = "mysql://root:${var.password}..."
[...]

PR #771, opened 2022-05-04
cmd/atlascmd: support project file in inspect

rotemtam:
This PR adds initial support for project files that define environments as proposed by @a8m on https://github.com/ariga/atlas/issues/709. 

Initially, an `atlas.hcl` file is searched for in the current working directory, it's format is: 

```hcl
env "local" {
	url = "mysql://root:pass@localhost:3306/"
	dev = "docker://mysql/8"
	src = "./app.hcl"
	schemas = ["hello", "world"]
}
```

With this PR, environments are supported with the `schema inspect` and `schema apply` commands, with more commands to follow incrementally until this feature is prepared. For the file above, this would be supported:

```text
atlas schema inspect local
# or
atlas schema apply local
```


cmd/atlascmd/project.go
```diff
@@ -0,0 +1,78 @@
...
+
+const projectFileName = "atlas.hcl"
+
+// projectFile represents an atlas.hcl file.
+type projectFile struct {
+	Envs []*Environment `spec:"env"`
+}
+
+// Environment represents an Atlas environment.
+type Environment struct {
```
a8m, line 23: Too ambiguous, let's rename Env to EnvCmd (before the release) and use `Env` here. Also, why `project.go` and not `env.go`?

cmd/atlascmd/project.go
```diff
@@ -0,0 +1,78 @@
...
+	"ariga.io/atlas/schema/schemaspec"
+	"ariga.io/atlas/schema/schemaspec/schemahcl"
+)
+
+const projectFileName = "atlas.hcl"
+
+// projectFile represents an atlas.hcl file.
+type projectFile struct {
+	Envs []*Environment `spec:"env"`
+}
```
a8m, line 20: I'd treat it as an Atlas config rather than an Atlas "project". A config file can store also credentials/tokens to Ariga, AWS, etc.

cmd/atlascmd/project.go
```diff
@@ -0,0 +1,78 @@
...
+
+// LoadEnv reads the project file in path, and loads the environment
+// with the provided name into env.
+func LoadEnv(path string, name string) (*Environment, error) {
+	b, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	project := projectFile{}
```
a8m, line 49:
```suggestion
	var project projectFile
```

cmd/atlascmd/schema.go
```diff
@@ -114,6 +116,20 @@ const (
...
+	if len(args) == 0 {
+		return nil
+	}
+	env, err := LoadEnv(projectFileName, args[0])
+	if err != nil {
+		return err
+	}
+	activeEnv = env
+	return nil
+}
```
a8m, line 131: Why do we need a global var here? You can return the value instead.

rotemtam: similar to cli flags, this value is used by other commands at runtime.

a8m: Does not feel right to me. Can we change it, please?

cmd/atlascmd/project.go
```diff
@@ -0,0 +1,76 @@
...
+	Schemas []string `spec:"schemas"`
+	schemaspec.DefaultExtension
+}
+
+// LoadEnv reads the project file in path, and loads the environment
+// with the provided name into env.
+func LoadEnv(path string, name string) (*Env, error) {
+	b, err := os.ReadFile(path)
+	if err != nil {
+		fmt.Println("@###", path)
```
a8m:
```suggestion
```

cmd/atlascmd/schema.go
```diff
@@ -256,12 +280,23 @@ func promptUser() bool {
...
+		activeEnv, err := selectEnv(args)
+		if err != nil {
+			return err
+		}
 		dsnF, urlF := cmd.Flag("dsn"), cmd.Flag("url")
 		switch {
+		case activeEnv != nil && activeEnv.URL != "":
+			urlF.Changed = true
+			if err := urlF.Value.Set(activeEnv.URL); err != nil {
+				panic(err)
```
a8m:
```suggestion
				return err
```

cmd/atlascmd/schema.go
```diff
@@ -129,8 +138,8 @@ func init() {
 	SchemaApply.Flags().BoolVarP(&ApplyFlags.Verbose, migrateDiffFlagVerbose, "", false, "enable verbose logging")
 	SchemaApply.Flags().StringToStringVarP(&ApplyFlags.Vars, "var", "", nil, "input variables")
 	cobra.CheckErr(SchemaApply.MarkFlagRequired("url"))
-	cobra.CheckErr(SchemaApply.MarkFlagRequired("file"))
-	dsn2url(SchemaApply, &ApplyFlags.URL)
+	//cobra.CheckErr(SchemaApply.MarkFlagRequired("file"))
```
a8m, line 141:
```suggestion
	//cobra.CheckErr(SchemaApply.MarkFlagRequired("file"))
```
?

a8m (approved):
Next step is to add it to migrate, right?

PR #779, opened 2022-05-08
internal/integration: testscript for cli tests

rotemtam:


internal/integration/integration_test.go
```diff
@@ -542,3 +541,11 @@ func (r *rrw) ReadRevisions(ctx context.Context) (migrate.Revisions, error) {
 func (r *rrw) clean() {
 	*r = rrw(migrate.Revisions{})
 }
+
+func compileCLI(t *testing.T) (string, error) {
```
a8m, line 545:
```suggestion
func buildCmd(t *testing.T) (string, error) {
```
A suggestion, no need to rename if you don't like it.

internal/integration/script_test.go
```diff
@@ -135,6 +137,11 @@ func (t *liteTest) setupScript(env *testscript.Env) error {
 	// environment as tests run in parallel.
 	env.Values[keyDB] = db
 	env.Values[keyDrv] = drv
+	cliPath, err := compileCLI(t.T)
```
a8m, line 140: Why not storing in the temporary directory (of testscript)?

internal/integration/script_test.go
```diff
@@ -310,6 +317,31 @@ func (t *liteTest) cmdExec(ts *testscript.TestScript, _ bool, args []string) {
 	cmdExec(ts, args, ts.Value(keyDB).(*sql.DB))
 }
 
+func (t *liteTest) cmdCLI(ts *testscript.TestScript, _ bool, args []string) {
+	for i, arg := range args {
+		args[i] = strings.ReplaceAll(arg, "URL", t.dsn())
+	}
+	l := len(args)
+	switch {
```
a8m, line 325:
```suggestion
	switch l := len(args); {
```

internal/integration/script_test.go
```diff
@@ -310,6 +317,31 @@ func (t *liteTest) cmdExec(ts *testscript.TestScript, _ bool, args []string) {
...
+			path := filepath.Join(wd, args[l-1])
+			out, err := os.Create(path)
+			if err != nil {
+				return err
+			}
+			defer out.Close()
+			cmd := exec.Command(ts.Getenv("cli"), args[0:l-2]...)
+			cmd.Stdout = out
+			return cmd.Run()
+		}())
```
a8m, line 339: why it's better than:
```suggestion
		wd := ts.Getenv("WORK")
		path := filepath.Join(wd, args[l-1])
		out, err := os.Create(path)
		ts.Check(err)
		defer out.Close()
		cmd := exec.Command(ts.Getenv("cli"), args[0:l-2]...)
		cmd.Stdout = out
		ts.Check(cmd.Run())
```

a8m (approved):
Looks good, added minor questions/suggestions

internal/integration/script_test.go
rotemtam: dont want to compile it for every test case

PR #780, opened 2022-05-09
cmd/atlascmd: replay migration dir if a dev-url is given

masseelch:
also refactor some tests


a8m:
Maybe make the "dev-url" a flag of the "migrate" command?

masseelch:
> Maybe make the "dev-url" a flag of the "migrate" command?

It will be mentioned to be valid for all commands then, but `atlas migrate hash` eg does not use it.

cmd/atlas/main.go
```diff
@@ -20,10 +20,6 @@ func main() {
 	ctx, _ := signal.NotifyContext(context.Background(), os.Interrupt, os.Kill)
 	atlascmd.Root.SetOut(os.Stdout)
 	err := atlascmd.Root.ExecuteContext(ctx)
-	// Print error from command
-	if err != nil {
-		atlascmd.Root.PrintErrln("Error:", err)
-	}
```
a8m, line 26: Can you elaborate?

masseelch: Cobra prints the error already by itself. No need to do it twice.

a8m (approved):

PR #791, opened 2022-05-12
sql/sqlite: simple file lock to implement schema.Locker for SQLite

masseelch:
This PR assumes, that SQLite files will most likely be run locally (which means never be accessed remotely) and therefore this implementation works for that cases. It aims to provide consistency with the other drivers.

### DIFF
--- a/sql/sqlite/driver.go
+++ b/sql/sqlite/driver.go
@@ -6,9 +6,15 @@ package sqlite
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io/ioutil"
 	"net/url"
+	"os"
+	"path/filepath"
+	"strconv"
 	"strings"
+	"time"
 
 	"ariga.io/atlas/sql/internal/sqlx"
 	"ariga.io/atlas/sql/migrate"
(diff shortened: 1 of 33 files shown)


sql/sqlite/driver.go
```diff
@@ -99,6 +105,43 @@ func (d *Driver) Clean(ctx context.Context) error {
 	return nil
 }
 
+// Lock implements the schema.Locker interface.
+func (d *Driver) Lock(_ context.Context, name string, timeout time.Duration) (schema.UnlockFunc, error) {
+	dir, err := os.UserCacheDir()
```
a8m, line 110: Why the elaborate why do you use CacheDir and not TempDir for example?

sql/sqlite/driver_test.go
```diff
@@ -0,0 +1,44 @@
+package sqlite
```
a8m, line 1: license

sql/sqlite/driver.go
masseelch: TempDir might get cleaned up occasionally by OS and I don't want that. Cache will live at least until reboot. And in that case it doesn't matter anymore since the DB will rebooted as well.

a8m: afaik, temporary files won’t be deleted until the process is exit. Am I wrong?

a8m (approved):
💯

PR #781, opened 2022-05-09
cmd/atlascmd: acquire a lock before executing migrations

masseelch:

PR #793, opened 2022-05-13
sql/migrate: introduce basic execution logging

masseelch:
Package consumers can then use the logs for whatever they need it to be (JSON / pretty TTY / Sentry / ...)

~Will be rebased against master once the base (#781) is merged.~


sql/migrate/migrate.go
```diff
@@ -574,6 +609,39 @@ func (r *Revision) setSQLErr(err error, stmt string) error {
...
+		Kind LogKind
+		// Pending files to be executed.
+		Pending []string
+		// Version executed.
+		Version string
+		// Desc of migration executed.
+		Desc string
+		// SQL statement executed.
+		SQL string
+	}
```
a8m, line 637: I wonder maybe define "LogEntry" as an interface and let the logger print it with type assertion. A good example is the `ast.Node` in the different `ast` packages (go, templates, graphql, etc).

sql/migrate/migrate.go
```diff
@@ -587,6 +609,37 @@ func (r *Revision) setSQLErr(err error, stmt string) error {
...
+	LogFile struct {
+		// Version executed.
+		Version string
+		// Desc of migration executed.
+		Desc string
+	}
+
+	// LogStmt is sent if a new SQL statement is executed.
+	LogStmt string
+)
```
a8m, line 637: Looks good, but I suggest using struct for all different types as it's easy to extend without introducing breaking changes.

a8m (approved):
Approved with one suggestion, you're welcome to accept or reject it :heart:

a8m (approved):

PR #796, opened 2022-05-14
sql/sqlclient: expose registered name

masseelch:
This is needed to create the Ent client for EntRevisions. Once Ent is based on the Atlas driver, it can be removed again.


a8m (approved):
Please, remember to revert it when it’s no longer needed.

PR #797, opened 2022-05-14
sql/mysql: Attempt a fix for RawExpr cases

### LINKED ISSUE #795: Inspect of default uuid() generates wrong schema.
### atlas version
```
atlas version v0.3.8-593b502ead94-canary
https://github.com/ariga/atlas/releases/latest
```

### mariadb --version
```
mariadb  Ver 15.1 Distrib 10.5.15-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper
```

### Steps to reproduce:
* Create database and table with MariaDB
  ```sql
  CREATE DATABASE `atlas_test` CHARSET utf8 COLLATE utf8_bin;
  CREATE TABLE `sample` (
    `id` char(36) COLLATE utf8_bin NOT NULL DEFAULT uuid(),
    `name` varchar(255) COLLATE utf8_bin NOT NULL DEFAULT '',
    `createdAt` timestamp NOT NULL DEFAULT current_timestamp(),
    `updatedAt` timestamp NULL DEFAULT NULL ON UPDATE current_timestamp(),
    PRIMARY KEY (`id`)
  ) DEFAULT CHARSET=utf8 COLLATE=utf8_bin COMMENT='Note id has default uuid()';
  ```
* Create and read a record
  ```sql
  INSERT INTO atlas_test.sample(name) VALUES ('Piyush');
  SELECT * FROM atlas_test.sample;
  ```
  ```
  +--------------------------------------+--------+---------------------+-----------+
  | id                                   | name   | createdAt           | updatedAt |
  +--------------------------------------+--------+---------------------+-----------+
  | 7f96e25f-d39d-11ec-b7d1-42010a800002 | Piyush | 2022-05-14 15:49:54 | NULL      |
  +--------------------------------------+--------+---------------------+-----------+
  1 row in set (0.000 sec)
  ```
* Generate HCL file with `atlas schema inspect`
  ```shell
  atlas schema inspect -u "mariadb://user:pass@localhost:3306/atlas_test" > atlas_test.hcl
  cat atlas_test.hcl
  ```
  ```hcl
  table "sample" {
    schema  = schema.atlas_test
    comment = "Note id has default uuid()"
    column "id" {
      null    = false
      type    = char(36)
      default = sql("uuid()")
    }
    column "name" {
      null    = false
      type    = varchar(255)
      default = ""
    }
    column "createdAt" {
      null    = false
      type    = timestamp
      default = sql("current_timestamp()")
    }
    column "updatedAt" {
      null      = true
      type      = timestamp
      on_update = sql("current_timestamp()")
    }
    primary_key {
      columns = [column.id]
    }
  }
  schema "atlas_test" {
    charset = "utf8"
    collate = "utf8_bin"
  }
  ```

* Note the `"uuid()"`. It'll get interpreted as a string in next step
* Drop the db to be created again
  ```sql
  DROP DATABASE atlas_test;
  ```
* Try to apply our generated `.hcl` file.
  ```shell
  atlas schema apply -u "mariadb://user:pass@localhost:3306" -f atlas_test.hcl
  ```
  ```sql
  -- Planned Changes:
  -- Add new schema named "atlas_test"
  CREATE DATABASE `atlas_test` CHARSET utf8 COLLATE utf8_bin
  -- Create "sample" table
  CREATE TABLE `atlas_test`.`sample` (`id` char(36) NOT NULL DEFAULT "uuid()", `name` varchar(255) NOT NULL DEFAULT "", `createdAt` timestamp NOT NULL DEFAULT current_timestamp(), `updatedAt` timestamp NULL ON UPDATE current_timestamp(), PRIMARY KEY (`id`)) COMMENT "Note id has default uuid()"
  ```
* Insert and read again
   ```sql
  INSERT INTO atlas_test.sample(name) VALUES ('Piyush');
  SELECT * FROM atlas_test.sample;
  ```
  ```
  +--------+--------+---------------------+-----------+
  | id     | name   | createdAt           | updatedAt |
  +--------+--------+---------------------+-----------+
  | uuid() | Piyush | 2022-05-14 16:04:40 | NULL      |
  +--------+--------+---------------------+-----------+
  1 row in set (0.000 sec)
  ```

ofpiyush:
This is an optimistic attempt at a fix for expressions/functions in schema. Feel free to correct and guide me towards a better solve. :)

The wrapping is somewhat redundant when we already have `sql()` but [going by the comments in code](https://github.com/ariga/atlas/blob/60c0ed8bff94979c621c3453426a46c3ad44ee46/sql/mysql/migrate.go#L676-L677), backwards compatibility has made the situation a little messy.

Fixes #795 

Note: Initially, MySQL looked like it was working fine without this change.
But if we tried to inspect -> drop database -> create. We ended up with string default instead of an expression.

### DIFF
--- a/sql/mysql/inspect.go
+++ b/sql/mysql/inspect.go
@@ -542,7 +542,7 @@ var reCurrTimestamp = regexp.MustCompile(`(?i)^current_timestamp(?:\(\d?\))?$`)
 func (i *inspect) myDefaultExpr(c *schema.Column, x string, attr *extraAttr) schema.Expr {
 	// In MySQL, the DEFAULT_GENERATED indicates the column has an expression default value.
 	if i.supportsExprDefault() && attr.defaultGenerated {
-		return &schema.RawExpr{X: x}
+		return &schema.RawExpr{X: sqlx.MayWrap(x)}
 	}
 	switch c.Type.Type.(type) {
 	case *schema.BinaryType:
@@ -556,7 +556,7 @@ func (i *inspect) myDefaultExpr(c *schema.Column, x string, attr *extraAttr) sch
 		// "current_timestamp" is exceptional in old versions
 		// of MySQL for timestamp and datetime data types.
 		if reCurrTimestamp.MatchString(x) {
-			return &schema.RawExpr{X: x}
+			return &schema.RawExpr{X: sqlx.MayWrap(x)}
 		}
 	}
 	return &schema.Literal{V: quote(x)}
--- a/sql/mysql/inspect.go
+++ b/sql/mysql/inspect.go
@@ -542,6 +542,10 @@ var reCurrTimestamp = regexp.MustCompile(`(?i)^current_timestamp(?:\(\d?\))?$`)
 func (i *inspect) myDefaultExpr(c *schema.Column, x string, attr *extraAttr) schema.Expr {
 	// In MySQL, the DEFAULT_GENERATED indicates the column has an expression default value.
 	if i.supportsExprDefault() && attr.defaultGenerated {
+		// CURRENT_TIMESTAMP is an exception for MySQL
+		if _, ok := c.Type.Type.(*schema.TimeType); ok && reCurrTimestamp.MatchString(x) {
+			return &schema.RawExpr{X: x}
+		}
 		return &schema.RawExpr{X: sqlx.MayWrap(x)}
 	}
 	switch c.Type.Type.(type) {
@@ -556,7 +560,7 @@ func (i *inspect) myDefaultExpr(c *schema.Column, x string, attr *extraAttr) sch
 		// "current_timestamp" is exceptional in old versions
 		// of MySQL for timestamp and datetime data types.
 		if reCurrTimestamp.MatchString(x) {
-			return &schema.RawExpr{X: sqlx.MayWrap(x)}
+			return &schema.RawExpr{X: x}
 		}
 	}
 	return &schema.Literal{V: quote(x)}
@@ -620,7 +624,7 @@ func (i *inspect) marDefaultExpr(c *schema.Column, x string) schema.Expr {
 		// "current_timestamp" is exceptional in old versions
 		// of MySQL (i.e. MariaDB in this case).
 		if strings.ToLower(x) == currentTS {
-			return &schema.RawExpr{X: sqlx.MayWrap(x)}
+			return &schema.RawExpr{X: x}
 		}
 	}
 	if !i.supportsExprDefault() {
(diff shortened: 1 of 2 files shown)


rotemtam:
hey @ofpiyush 
thanks for working on this! notice that CI is red, do you mind taking a look at why? 
🙏

ofpiyush:
Yes, I was playing with atlas and the bug just sucked me in. That commit was a lesson in why you shouldn't push code at 3 AM 😅

CURRENT_TIMESTAMP cases were failing. For now, I've added an exception for them.

A better solve for this family of problems will be fairly involved. Comparing the exact query can be brittle but attempting to compare the intent(via AST) behind the query might still result in wrong query and/or lot of edge cases to handle on a case to case basis.

Another such example is a field with default `CONCAT(my_field,' anystr')`. This generates an escaped query (tried with MySQL 8), which again breaks in the inspect -> drop database -> apply scenario.

a8m:
Thanks for the contribution @ofpiyush 🙏 

Code looks great, but the tests in `integration/mysql_test.go` need to be updated with the new inspection format. Let me know if you need any help with this.

> Another such example is a field with default CONCAT(my_field,' anystr'). This generates an escaped query (tried with MySQL 8), which again breaks in the inspect -> drop database -> apply scenario.

Can you elaborate with an example? Did you address it in this PR? If not, that's fine, I can take this as well.

ofpiyush:
@a8m Would you believe that those tests pass in go 1.17? Would be interesting to know why!

I've updated the test cases.

> Can you elaborate with an example? Did you address it in this PR? If not, that's fine, I can take this as well.

I've opened https://github.com/ariga/atlas/issues/803 to describe the bug. I haven't attempted it mostly because I am not sure if we're missing an `unescape` or is it because `unescape` itself isn't doing what it is supposed to.

sql/mysql/inspect.go
```diff
@@ -542,7 +542,11 @@ var reCurrTimestamp = regexp.MustCompile(`(?i)^current_timestamp(?:\(\d?\))?$`)
 func (i *inspect) myDefaultExpr(c *schema.Column, x string, attr *extraAttr) schema.Expr {
 	// In MySQL, the DEFAULT_GENERATED indicates the column has an expression default value.
 	if i.supportsExprDefault() && attr.defaultGenerated {
-		return &schema.RawExpr{X: x}
+		// CURRENT_TIMESTAMP is an exception for MySQL
```
a8m, line 545:
```suggestion
		// Skip CURRENT_TIMESTAMP, because wrapping it with parens will translate it to now().
```

a8m (approved):

a8m:
Thanks for the contribution @ofpiyush 🙏

PR #798, opened 2022-05-14
cmd/atlascmd: use env flag for inspect/apply

rotemtam:


cmd/atlascmd/schema.go
```diff
@@ -115,101 +122,128 @@ const (
...
 
 	// Schema fmt.
 	schemaCmd.AddCommand(SchemaFmt)
 }
 
+// selectEnv returns the Env from the current project file based on the selected
+// argument. If selected is "", or no project file exists in the current directory
+// a zero-value Env is returned.
+func selectEnv(selected string) (*Env, error) {
+	if _, err := os.Stat(projectFileName); selected == "" || os.IsNotExist(err) {
```
masseelch, line 160: Why hit the disk if selected is zero?

masseelch (commented):
I feel like it is more clean to load the env into the flag values before running the command (like in PersistentPreRun e.g.) instead if in each command. 

WDYT?

cmd/atlascmd/schema.go
```diff
@@ -115,101 +122,128 @@ const (
...
+	return LoadEnv(projectFileName, selected)
+}
+
 // CmdInspectRun is the command used when running CLI.
-func CmdInspectRun(cmd *cobra.Command, args []string) {
+func CmdInspectRun(cmd *cobra.Command, _ []string) error {
 	if InspectFlags.Web {
 		schemaCmd.PrintErrln("The Alas UI is not available in this release.")
-		return
+		return errors.New("unsupported")
```
a8m, line 170: Maybe remove the printing and return the message as an error? 

```suggestion
		return errors.New("Atlas UI is not available in this release")
```

cmd/atlascmd/schema.go
```diff
@@ -88,7 +95,7 @@ databases), omit the relevant part from the url, e.g. "mysql://user:pass@localho
 To select specific schemas from the databases, users may use the "--schema" (or "-s" shorthand)
 flag.
 	`,
-		Run: CmdInspectRun,
+		RunE: CmdInspectRun,
```
a8m, line 98: Why did you replace `Run` with `RunE`?

cmd/atlascmd/schema.go
```diff
@@ -115,101 +122,128 @@ const (
...
 	}
-	client, err := sqlclient.Open(cmd.Context(), InspectFlags.URL)
-	cobra.CheckErr(err)
 	defer client.Close()
-	schemas := InspectFlags.Schema
-	activeEnv, err := selectEnv(args)
-	cobra.CheckErr(err)
-	if activeEnv != nil && len(activeEnv.Schemas) > 0 {
-		schemas = activeEnv.Schemas
+	// Select the schemas.
```
a8m, line 189:
```suggestion
```

cmd/atlascmd/schema.go
```diff
@@ -115,101 +122,128 @@ const (
...
+	schemas := activeEnv.Schemas
+	if fl := cmd.Flag(schemaFlag); fl.Changed {
+		schemas = InspectFlags.Schema
 	}
 	if client.URL.Schema != "" {
 		schemas = append(schemas, client.URL.Schema)
 	}
+	// Inspect the realm.
```
a8m, line 197:
```suggestion
```

cmd/atlascmd/schema.go
```diff
@@ -115,101 +122,128 @@ const (
...
 	// Schema fmt.
 	schemaCmd.AddCommand(SchemaFmt)
 }
 
+// selectEnv returns the Env from the current project file based on the selected
+// argument. If selected is "", or no project file exists in the current directory
+// a zero-value Env is returned.
+func selectEnv(selected string) (*Env, error) {
+	if _, err := os.Stat(projectFileName); selected == "" || os.IsNotExist(err) {
+		return &Env{}, nil
```
a8m, line 161: To make the code simpler in CmdInspectRun, I'll suggest doing something like this even though it creates small duplication with CmdApplyRun:

```suggestion
		return &Env{
			URL: InspectFlags.URL,
			Schemas: InspectFlags.Schemas,
		}, nil
```

cmd/atlascmd/schema.go
rotemtam: Honestly just dislike cobra.CheckErr everywhere. Returning an error makes the code seem nicer to me.
Not strongly attached to this

a8m: I don't mind, I wondered why it was needed in this PR. Let me know when it's ready and I'll review this

PR #799, opened 2022-05-14
cmd/atlascmd: env support for migrate commands

rotemtam:


a8m (commented):
The new change adds a lot of complexity which is a bit surprising. Don't you prefer to use cobra hooks (e.g. pre-run) to set the flags?

cmd/atlascmd/schema.go
```diff
@@ -31,15 +39,22 @@ var (
 		Long:  "The `atlas schema` command groups subcommands for working with Atlas schemas.",
 	}
 
+	// SchemaFlags are common flags used by schema commands.
+	SchemaFlags struct {
+		URL     string
+		Schemas []string
+
+		// Deprecated: DSN is an alias for URL.
+		DSN string
```
a8m, line 48: Why it's here? We should not accept it in the project file.

rotemtam: I moved flags common between inspect and apply here. It's not supported in the project file

a8m (commented):
Looks good, minor questions before I approve this. 

Thanks

a8m (approved):

PR #806, opened 2022-05-18
sql/mysql: move charset writing to data_type stage

a8m:


sql/mysql/migrate.go
```diff
@@ -467,6 +467,16 @@ func (s *state) column(b *sqlx.Builder, t *schema.Table, c *schema.Column) error
 		return fmt.Errorf("format type for column %q: %w", c.Name, err)
 	}
 	b.Ident(c.Name).P(typ)
+	if cs := (schema.Charset{}); sqlx.Has(c.Attrs, &cs) {
+		if !supportsCharset(c.Type.Type) {
+			return fmt.Errorf("column %q of type %T does not support the CHARSE attribute", c.Name, c.Type.Type)
```
masseelch, line 472:
```suggestion
			return fmt.Errorf("column %q of type %T does not support the CHARSET attribute", c.Name, c.Type.Type)
```

masseelch (approved):
:100:

PR #808, opened 2022-05-18
sql/migrate: remove GlobStateReader and replace by Executor

masseelch:
This removes the deprecated GlobStateReader and replaces it by the Executor. The Executor is not a StateReader (it replays the migration directory on the dev-db and inspects it).

This enables us in the future to support reading other formats like Liquibase etc as well and ease transition from those tools to Atlas. We will also be able to "rewrite" the Liquibase to the Atlas format (like creating a snapshot e.g.).

This PR is also needed for the CLI to work since the Executor can support multi-schema setups while this was not possible before.


a8m:
Please, fix CI failures and I'll approve it 😄

a8m (approved):

PR #810, opened 2022-05-18
doc/md: document project files

rotemtam:


doc/md/cli/projects.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Project Structure
+id: projects
+slug: /cli/projects
+---
+### Atlas Project Files
```
a8m, line 6:
```suggestion
### Project Files
```

doc/md/cli/projects.md
```diff
@@ -0,0 +1,125 @@
...
+
+Project files provide a convenient way to describe and work with different
+environments which are managed using Atlas.  A project file is a file named
+`atlas.hcl` which is located at the root of the project and contains one or
```
hilakashai:
```suggestion
`atlas.hcl` which is located at the root of the project and contains one or more
```

doc/md/cli/projects.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Project Structure
+id: projects
+slug: /cli/projects
+---
+### Atlas Project Files
+
+Project files provide a convenient way to describe and work with different
+environments which are managed using Atlas.  A project file is a file named
```
a8m, line 9:
```suggestion
Project files provide a convenient way to describe and interact with multiple
environments when working with Atlas. A project file is a file named
```

Not sure it's better, but I like it

doc/md/cli/projects.md
```diff
@@ -0,0 +1,125 @@
...
+`atlas.hcl` which is located at the root of the project and contains one or 
+`env` blocks. For example:
```
a8m, line 11:
```suggestion
`atlas.hcl` and contains one or more `env` blocks. For example:
```

The location does not really matter.

doc/md/cli/projects.md
```diff
@@ -0,0 +1,125 @@
...
+env "dev" {
+  // ... a different env
+}
+```
+
+Once defined, a project's environment can be worked against using the `--env` flag. 
+For example:
+
+```shell
+$ atlas schema apply --env local
```
a8m, line 40:
```suggestion
atlas schema apply --env local
```
We have a copy button.

a8m: I see now that it exists in other places. I'll send a patch to fix it in all other places.

a8m (approved):
💯

hilakashai (approved):
💯

PR #813, opened 2022-05-23
cmd/atlascmd: better err message when missing required var

rotemtam:


a8m:
Test scripts are failing 😍

PR #814, opened 2022-05-23
sql/crdb: basic driver and empty tests

hedwigz:


sql/postgres/driver.go
```diff
@@ -44,6 +44,7 @@ func init() {
 		"postgres",
 		sqlclient.DriverOpener(Open),
 		sqlclient.RegisterCodec(MarshalHCL, EvalHCL),
+		sqlclient.RegisterFlavours("cockroach", "crdb"),
```
a8m, line 47: Let's move it to a standalone driver. A separate sqlcient and a driver.

a8m (changes requested):
See comment.

PR #815, opened 2022-05-23
sql/cockroach: make postgres inspect compatible with crdb

hedwigz:


sql/postgres/driver.go
```diff
@@ -203,6 +205,8 @@ const (
 	TypeInt2     = "int2" // smallint.
 	TypeInt4     = "int4" // integer.
 	TypeInt8     = "int8" // bigint.
+	// cockroach bigint
+	TypeInt64 = "int64"
```
a8m, line 209:
```suggestion
	TypeInt64 = "int64" // cockroach bigint.
```

sql/postgres/driver.go
```diff
@@ -216,6 +220,8 @@ const (
 	TypePath    = "path"
 	TypePolygon = "polygon"
 	TypePoint   = "point"
+	// cockroach geometry
+	TypeGeometry = "geometry"
```
a8m, line 224:
```suggestion
	TypeGeometry = "geometry"
```

a8m (approved):

PR #828, opened 2022-05-30
sql/sqlclient: expose registered urlparser

masseelch:
Will be undone once Ent is based on Atlas driver.

### DIFF
--- a/sql/sqlclient/client.go
+++ b/sql/sqlclient/client.go
@@ -42,6 +42,9 @@ type (
 		// and encoding the schema documents.
 		schemaspec.Marshaler
 		schemahcl.Evaluator
+
+		// The urlParser used to construct the URL for this client.
+		parse urlParser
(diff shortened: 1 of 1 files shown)


sql/sqlclient/client.go
```diff
@@ -198,7 +208,7 @@ func Register(name string, opener Opener, opts ...RegisterOption) {
 	}
 	opt := &registerOptions{
 		// Default URL parser uses the URL as the DSN.
-		parse: func(u *url.URL) *URL { return &URL{URL: u, DSN: u.String()} },
+		parse: func(u *url.URL) *URL { return &URL{URL: u} },
```
a8m, line 212: Why?

a8m:
From looking at the code now - why you can't change the `Path` of the `URL` and then call `sqlclient.Open` again?

PR #830, opened 2022-05-30
sql/crdb: all integration tests

hedwigz:


internal/integration/integration_test.go
```diff
@@ -179,20 +179,23 @@ func testHCLIntegration(t T, full string, empty string) {
...
+	runArgs := []string{
+		"run", "ariga.io/atlas/cmd/atlas",
 		"schema",
 		"inspect",
 		"-d",
 		dsn,
-	)
+	}
+	runArgs = append(runArgs, args...)
+	cmd := exec.Command("go", runArgs...)
```
a8m, line 198:
```suggestion
	cmd := exec.Command("go", append([]string{
		"run", "ariga.io/atlas/cmd/atlas",
		"schema",
		"inspect",
		"-d",
		dsn,
	}, args...)...)
```

internal/integration/postgres_test.go
```diff
@@ -493,7 +492,7 @@ func TestPostgres_HCL_Realm(t *testing.T) {
 	pgRun(t, func(t *pgTest) {
 		t.dropSchemas("second")
 		realm := t.loadRealm()
-		hcl, err := mysql.MarshalHCL(realm)
+		hcl, err := postgres.MarshalHCL(realm)
```
a8m, line 495: 👍

internal/integration/integration_test.go
hedwigz: I don't think it is pretty `args...)...)`

PR #841, opened 2022-06-01
ci: use our crdb image

hedwigz:


masseelch (changes requested):
What about the `docker-compose.yml`?

hedwigz:
> What about the `docker-compose.yml`?

`docker-compose.yml` has the `command` option so that I can start the cockroach image with `start-single-node --insecure` but the GitHub action service doesn't have this option so I built another image to solve that.

a8m:
Can’t you use run.entrypoint or run.args for this? 
https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsentrypoint

.github/workflows/ci.yml
```diff
@@ -480,7 +480,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       cockroach:
-        image: timveil/cockroachdb-single-node:latest
+        image: hedwigz/cockroachdb-single-node:v21.2.11
```
a8m, line 483: I think we can build it in the action and use it here, no?

a8m (approved):

PR #844, opened 2022-06-01
sql/migrate: refactor execution (in preparation for transactions)

masseelch:
This PR refactors the migration execution into two methods instead of one.

1. Compute the pending migration files
2. Execute a migration file onto a database

The above changes are needed to be able to wrap the execution of a migration file in a transaction. Splitting this into the two above methods also enables concurrent migration of multi-tenancy setups and keep them in sync before executing a new migration file.

Wrapping a migration execution in a transaction is should be controlled by the consumer instead of the executor itself.


a8m (approved):
🚀

PR #848, opened 2022-06-08
cmd/atlasci/internal/ci: add basic run structure

a8m:
The general idea is to wrap these analyzers with top-level
runner that can be configured with custom reporter/writer
and setup driver-specific configuration.

On the execution part, the runner should be responsible for
scanning (and parsing, soon) the SQL files in the migration
directory, create a pass (check context) for each new one and
execute the registered and desired analyzers on it.

The reporting should be environment-aware. e.g. GitHub PR,
local development (HCL, Ent, plain SQL, etc).

### DIFF
--- a/cmd/atlasci/internal/ci/run.go
+++ b/cmd/atlasci/internal/ci/run.go
@@ -1,0 +1,64 @@
+// Copyright 2021-present The Atlas Authors. All rights reserved.
+// This source code is licensed under the Apache 2.0 license found
+// in the LICENSE file in the root directory of this source tree.
+
+package ci
+
+import (
+	"context"
+
+	"ariga.io/atlas/sql/migrate"
+	"ariga.io/atlas/sql/sqlcheck"
+	"ariga.io/atlas/sql/sqlclient"
+)
+
+// Runner is used to execute CI jobs.
+type Runner struct {
+	// DevClient configures the "dev driver" to calculate
+	// migration changes by the driver.
+	Dev *sqlclient.Client
+
+	// RunChangeDetector configures the ChangeDetector to
+	// be used by the runner.
+	ChangeDetector ChangeDetector
+
+	// Scan is used for scanning the migration directory.
+	Scan migrate.Scanner
+
+	// Analyzer defines the analysis to be run in the CI job.
+	Analyzer sqlcheck.Analyzer
+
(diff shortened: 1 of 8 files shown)


cmd/atlasci/internal/ci/run.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"context"
+
+	"ariga.io/atlas/sql/migrate"
+	"ariga.io/atlas/sql/sqlcheck"
+	"ariga.io/atlas/sql/sqlclient"
+)
+
+// Runner is used to execute CI jobs.
+type Runner struct {
```
a8m, line 16: I avoid creating wrappers/constructors for now as it's an internal package and it's in development.

sql/sqlcheck/sqlcheck.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2021-present The Atlas Authors. All rights reserved.
+// This source code is licensed under the Apache 2.0 license found
+// in the LICENSE file in the root directory of this source tree.
+
+package sqlcheck
```
a8m, line 5: This package resides in the top-level project because drivers will register their analyzers to it.

cmd/atlasci/internal/ci/run.go
```diff
@@ -0,0 +1,64 @@
...
+		stmt, err := r.Scan.Stmts(f)
+		if err != nil {
+			return err
+		}
+		for _, s := range stmt {
+			if _, err := r.Dev.ExecContext(ctx, s); err != nil {
+				return err
+			}
+		}
+	}
```
masseelch, line 52: Why not use an `migrate.Executor` here? The Executor will be able (soon) to e.g. check for checkpoints and stuff to speed up reaching base. By doing it the way you do, we have to built the logic twice (once we do).

There is a `NoopRevisionReadWriter` for exactly that case. 

For the `Dir` / `Scanner` you can use an "in memory" implementation to serve the base files.

a8m: I missed the `NoopRevisionReadWriter` one. I'll give it a look now.

a8m: I see now that it depends on the hash file. Let's talk about it offline.

PR #854, opened 2022-06-10
sql/sqlite: fix sequence primary key bug

masseelch:
This PR ensures that if in cases where the `sqlite_sequence` table does not exist, this error is not returned back to the caller. Since we add a statement to create an entry in the `sqlite_sequence` table only if there is a PK requiring it, the table will exist when applying the changes.


a8m (approved):
Really good catch! Thanks for fixing this ❤️

PR #858, opened 2022-06-13
sql/sqlcheck: add rename analyzer

omarhachach:
Converted atlas-contrib implementation to an analyzer.
Tests include only a happy-path test. Follows structure of #853

### DIFF
--- a/sql/sqlcheck/sqlcheck.go
+++ b/sql/sqlcheck/sqlcheck.go
@@ -80,6 +80,14 @@ func (a Analyzers) Analyze(ctx context.Context, p *Pass) error {
 	return nil
 }
 
+// AnalyzeFunc implements Analyzer.
+type AnalyzeFunc func(context.Context, *Pass) error
+
+// Analyze calls a(ctx, p).
+func (a AnalyzeFunc) Analyze(ctx context.Context, p *Pass) error {
+	return a(ctx, p)
+}
+
 // ReportWriterFunc is a function that implements Reporter.
 type ReportWriterFunc func(Report)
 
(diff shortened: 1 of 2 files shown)


sql/sqlcheck/sqlcheck_test.go
```diff
@@ -89,3 +122,35 @@ type testFile struct {
 func (t testFile) Name() string {
 	return t.name
 }
+
+type mockDriver struct {
```
a8m, line 126: The mock does not need to have any logic in it. I should get the desired response as an argument and return it.

sql/sqlcheck/sqlcheck.go
```diff
@@ -133,6 +133,56 @@ var Destructive = &driverAware{
...
+						continue
+					}
+					for j := range c.Changes[i+1:] {
+						a, ok := c.Changes[i+1+j].(*schema.AddColumn)
+						if !ok {
+							continue
+						}
+						// Use schema.Differ to see if type has changed, minimal setup to test only type changes
+						fromTbl := schema.NewTable("f").AddColumns(schema.NewColumn("diff").SetType(d.C.Type.Type))
+						toTbl := schema.NewTable("f").AddColumns(schema.NewColumn("diff").SetType(a.C.Type.Type))
```
a8m, line 158: The type is not enough.
The columns should look the same. Hence, they need to have the same attributes.

sql/sqlcheck/sqlcheck.go
```diff
@@ -133,6 +133,56 @@ var Destructive = &driverAware{
...
+				if !ok {
+					continue
+				}
+			Loop:
+				for i := range c.Changes {
+					d, ok := c.Changes[i].(*schema.DropColumn)
+					if !ok {
+						continue
+					}
+					for j := range c.Changes[i+1:] {
```
a8m, line 151: Please, document why we only look forward.
https://github.com/ariga/atlas/blob/master/sql/internal/sqlx/diff.go#L188-L193

a8m:
Something I discussed with @masseelch about it, is that we need to determine (from the statement) if the column was actually dropped and added and not renamed to avoid false-positive reporting. It can be done with regexp for now, but we'll add a proper parser to the statements soon.

a8m (commented):
Great! See comments

sql/sqlcheck/sqlcheck.go
masseelch: True. Also, please make sure to add a test for this.

sql/sqlcheck/sqlcheck.go
```diff
@@ -80,6 +80,14 @@ func (a Analyzers) Analyze(ctx context.Context, p *Pass) error {
 	return nil
 }
 
+// AnalyzeFunc implements Analyzer.
+type AnalyzeFunc func(context.Context, *Pass) error
+
+// Analyze calls a(ctx, p).
+func (a AnalyzeFunc) Analyze(ctx context.Context, p *Pass) error {
+	return a(ctx, p)
```
a8m, line 88:
```suggestion
func (f AnalyzeFunc) Analyze(ctx context.Context, p *Pass) error {
	return f(ctx, p)
```

sql/sqlcheck/sqlcheck.go
```diff
@@ -133,6 +141,58 @@ var Destructive = &driverAware{
...
+						return err
+					}
+					if len(changes) > 0 {
+						continue
+					}
+					diags = append(diags, Diagnostic{
+						Pos:  sc.Pos,
+						Text: fmt.Sprintf("Potential rename from column %q to %q in table %q", d.C.Name, a.C.Name, c.T.Name),
+					})
+					continue Loop // Be greedy, assume first match is most likely correct.
```
a8m, line 182: Why not break instead?

sql/sqlcheck/sqlcheck.go
```diff
@@ -133,6 +141,58 @@ var Destructive = &driverAware{
 	},
 }
 
+// Renames detects potential column renames.
+var Renames = AnalyzeFunc(func(ctx context.Context, p *Pass) error {
+	diags := make([]Diagnostic, 0)
```
a8m, line 146:
```suggestion
	var diags []Diagnostic
```

PR #859, opened 2022-06-13
atlasci/internal/ci: migration analyzer

omarhachach:


sql/sqlcheck/sqlcheck.go
```diff
@@ -133,6 +134,40 @@ var Destructive = &driverAware{
...
+			return err
+		}
+		diags = append(diags, Diagnostic{
+			Pos:  0,
+			Text: fmt.Sprintf("There was a checksum mismatch"),
+		})
+	}
+	if len(diags) > 0 {
+		p.Reporter.WriteReport(Report{
+			Text:        fmt.Sprintf("Migrations directory %q could not be verified", m.dir),
```
masseelch, line 164:
```suggestion
			Text:        fmt.Sprintf("Migration directory %q could not be verified", m.dir),
```

masseelch: Also we might wanna rephrase to something like "Migration directory is invalid". WDYT? @a8m

masseelch (commented):
Looks good, please see comments :blush:

sql/sqlcheck/sqlcheck.go
omarhachach: I thought about if it were possible to check which line/file was changed

cmd/atlasci/internal/ci/run.go
```diff
@@ -156,3 +157,43 @@ func (f *FileReport) WriteReport(r sqlcheck.Report) {
 func (w *TemplateWriter) WriteReport(r SummaryReport) error {
 	return w.T.Execute(w.W, r)
 }
+
+// MigrationAnalyzer implements sqlcheck.Analyzer. It validates the migration dir.
+type MigrationAnalyzer struct {
```
a8m, line 162: Let's call it `ChecksumAnalyzer`.

cmd/atlasci/internal/ci/run.go
```diff
@@ -156,3 +157,43 @@ func (f *FileReport) WriteReport(r sqlcheck.Report) {
...
+	Dir migrate.Dir
+}
+
+// Analyze implements the sqlcheck.Analyzer interface.
+func (m *MigrationAnalyzer) Analyze(_ context.Context, p *sqlcheck.Pass) error {
```
a8m, line 167: Let's return an error instead and not continue to the rest of the analyzers. WDYT?

a8m (approved):

PR #866, opened 2022-06-15
sql/migrate: files containing `atlas:sum none` in their first line wi…

masseelch:
…ll now be excluded from atlas sum file computation

### DIFF
--- a/sql/migrate/migrate.go
+++ b/sql/migrate/migrate.go
@@ -14,8 +14,10 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


sql/migrate/migrate.go
```diff
@@ -842,13 +844,21 @@ type templateFile struct {
 // Name implements the File interface.
 func (f *templateFile) Name() string { return f.n }
 
-// Filename of the migration directory integrity sum file.
-const hashFile = "atlas.sum"
+const (
+	// Filename of the migration directory integrity sum file.
+	hashFile = "atlas.sum"
+	// Directive used it a file should be excluded by the sum computation.
+	directiveNone = "none"
```
a8m, line 851:
```suggestion
	directiveIgnore = "ignore"
```

a8m (approved):

PR #867, opened 2022-06-16
sql/postgres: ensure identity column updates are not harmful

a8m:


sql/postgres/migrate.go
```diff
@@ -630,44 +630,49 @@ func (s *state) columnDefault(b *sqlx.Builder, c *schema.Column) {
...
 			}
 			// The syntax for altering identity columns is identical to sequence_options.
 			// https://www.postgresql.org/docs/current/sql-altersequence.html
-			b.P("SET GENERATED", id.Generation, "SET START WITH", strconv.FormatInt(id.Sequence.Start, 10), "SET INCREMENT BY", strconv.FormatInt(id.Sequence.Increment, 10), "RESTART")
+			b.P("SET GENERATED", toI.Generation, "SET START WITH", strconv.FormatInt(toI.Sequence.Start, 10), "SET INCREMENT BY", strconv.FormatInt(toI.Sequence.Increment, 10))
+			// Skip SEQUENCE RESTART in case the "start value" is less than the "current value" in one
+			// of the states (inspected and desired), because this function is used for both UP and DOWN.
+			if fromI, ok := identity(c.From.Attrs); (!ok || fromI.Sequence.Current < toI.Sequence.Start) && toI.Sequence.Current < toI.Sequence.Start {
+				b.P("RESTART")
+			}
```
masseelch, line 671: This also means when changing the start value in the desired state, nothing will happen, correct? Maybe we should also add some kind of analyzer for this one, checking if there is an sequence change, that can possibly break behaviour.

sql/postgres/migrate_test.go
```diff
@@ -414,6 +414,41 @@ func TestPlanChanges(t *testing.T) {
...
+				Reversible:    true,
+				Transactional: true,
+				Changes: []*migrate.Change{
+					{
+						Cmd:     `ALTER TABLE "users" DROP COLUMN "name", ALTER COLUMN "id" SET GENERATED BY DEFAULT SET START WITH 1024 SET INCREMENT BY 1`,
+						Reverse: `ALTER TABLE "users" ALTER COLUMN "id" SET GENERATED BY DEFAULT SET START WITH 1 SET INCREMENT BY 1, ADD COLUMN "name" character varying NOT NULL`,
+					},
+				},
+			},
+		},
```
masseelch, line 451: Can you add a test for the issue that caused this change? 

Add inc, add entry, drop entry, diff, apply, add entry?

masseelch (commented):
One question, also CI fails :-)

sql/postgres/migrate.go
a8m: > This also means when changing the start value in the desired state, nothing will happen, correct?

Yes.

a8m:
Checked that the new test case fails on master:

```
Error:      	Received unexpected error:
        	pq: duplicate key value violates unique constraint "global_id_pkey"
Test:       	TestPostgres_Ent/GlobalUniqueID/postgres14
```

sql/postgres/migrate_test.go
a8m: done

masseelch (commented):
:100: :rocket: :bear:

masseelch (approved):

Issue #751, opened 2022-04-29
mysql binary type error


calvin2021y:
varbinary is work
```hcl
table "test" {
	schema  = schema.test
	column "bin" {
		type = varbinary(32)
		null = false
	}
}
schema "test" {
	charset = "utf8mb4"
	collate = "utf8mb4_unicode_ci"
}
```

```sql
CREATE TABLE `test`.`test` (`bin` varbinary(32) NOT NULL)
```

binary is not work
```hcl
table "test" {
	schema  = schema.test
	column "bin" {
		type = binary(32)
		null = false
	}
}
schema "test" {
	charset = "utf8mb4"
	collate = "utf8mb4_unicode_ci"
}
```

```sql
CREATE TABLE `test`.`test` (`bin` binary NOT NULL)
```

there is no fixed size 32 in SQL.

Test on today master branch.

a8m:
Hey @calvin2021y 👋 
Thanks for reporting this issue. The fix should be simple.

Would you like to contribute this fix to Atlas? I can support you on our [Discord server](https://discord.gg/zZ6sWVg6NT), but if not that's also fine, I can handle this later today or tomorrow.

calvin2021y:
hi @a8m 
Sorry for not to attend to discord server, I still don't have discord account.

Can you give some tips here how to fix this ?

calvin2021y:
confirm fixed.

PR #869, opened 2022-06-17
sql/mysql:  fix binary with size

calvin2021y:
fix https://github.com/ariga/atlas/issues/751


internal/integration/mysql_test.go
```diff
@@ -186,7 +186,7 @@ func TestMySQL_AddColumns(t *testing.T) {
 			&schema.Column{Name: "b", Type: &schema.ColumnType{Raw: "mediumblob", Type: &schema.BinaryType{T: "mediumblob"}}},
 			&schema.Column{Name: "c", Type: &schema.ColumnType{Raw: "blob", Type: &schema.BinaryType{T: "blob"}}},
 			&schema.Column{Name: "d", Type: &schema.ColumnType{Raw: "longblob", Type: &schema.BinaryType{T: "longblob"}}},
-			&schema.Column{Name: "e", Type: &schema.ColumnType{Raw: "binary", Type: &schema.BinaryType{T: "binary"}}},
+			&schema.Column{Name: "e", Type: &schema.ColumnType{Raw: "binary(255)", Type: &schema.BinaryType{T: "binary(255)"}}},
```
rotemtam, line 189: Please add a case instead of deleting an existing one

rotemtam (commented):
See comment

internal/integration/mysql_test.go
calvin2021y: I am not sure why it not pass without a size.  

If I don't give a size it will provide a default non-zero size for binary ?

rotemtam: If an existing test fails because of a change that was made, it generally should mean that the change is not backward compatible (i.e it changes the existing behavior of the system).

a8m: Just checked it with both MySQL 8 and 5.7 and that's the expected behavior.

#### Version 8
image

#### Version 5.7

image

a8m: 👍

PR #874, opened 2022-06-18
sql/postgres: support dropping unique constraints

a8m:
See testscript.


internal/integration/testdata/postgres/index-unique-constraint.txt
```diff
@@ -0,0 +1,100 @@
+# Create table with UNIQUE constraint. i.e. implicit unique index.
+execsql 'CREATE TABLE script_index_unique_constraint.users (name text, last text, nickname text UNIQUE, UNIQUE(name, last))'
```
rotemtam, line 2: 💯  for this test

rotemtam (approved):

PR #875, opened 2022-06-19
sql/migrate: change Read to Bytes in File interface.

a8m:
Allow reading the files multiple times.


rotemtam (approved):
Sounds reasonable.

PR #876, opened 2022-06-19
cmd/atlasci: adding license subcommand

rotemtam:


cmd/atlasci/main.go
```diff
@@ -29,6 +29,10 @@ import (
 func main() {
 	var opt options
 	check(opt.parse())
+	if opt.license {
+		fmt.Println(license)
+		return
+	}
```
a8m, line 35: You can avoid this field and just call `os.Exit(0)` in parse.

a8m (approved):
💯

cmd/atlasci/main.go
rotemtam: done

PR #889, opened 2022-06-23
sql/sqlcheck: skip temporary resources in destructive analyzer

a8m:


sql/sqlcheck/sqlcheck.go
```diff
@@ -94,74 +94,118 @@ func (f ReportWriterFunc) WriteReport(r Report) {
...
+	SpanAdded
+
+	// SpanDropped describes that a span of
+	// a resource was ended in this file.
+	SpanDropped
+
+	// SpanTemporary indicates that a resource lifetime
+	// was started and ended in this file (CREATE and DROP).
+	SpanTemporary = SpanAdded | SpanDropped
+)
```
a8m, line 117: This may look overkill for one analyzer, but I also use it in the `DataDepend` analyzer (next PR).

masseelch (commented):
While this approach let's us capture the lifetime of a resource, it does not enable us to have analyzers to work on a bunch of information gathered from other analyzers. I still think that approach would make this cleaner and open us for the future.

masseelch (approved):

PR #895, opened 2022-06-29
Fix typo in README.md

ndrpnt:
Or did I completely missed the point?


a8m (approved):
💯 

Thanks for the fix @ndrpnt

PR #900, opened 2022-06-30
sql/migrate: StateReader if Dir has precedence

masseelch:
This change is the first one in a series of PR's to gradually make the migration directory the source of truth for all migration related logic (scanning, formatting, reading state, ...)

In the future one migration directory will be configured for exactly one dialect, migration tool, etc.


a8m (approved):
🚀

PR #906, opened 2022-07-04
Fix columnsOnly logic

junfengye:
add part Attrs len check


a8m:
Hey @junfengye and thanks for the contribution 🙏🏻

Would you mind adding a small test case for this or elaborating what issue it solves?

a8m:
I found out that this was supposed to be a fix for a bug we found yesterday. https://github.com/ariga/atlas/pull/1194.

Thanks for the contribution. Issue was resolved 🙏🏻

PR #907, opened 2022-07-04
schema/schemaspec/schemahcl: support evaluating multiple files

rotemtam:


schema/schemaspec/schemahcl/hcl.go
```diff
@@ -70,12 +69,69 @@ func (s *State) UnmarshalSpec(data []byte, v interface{}) error {
 	return s.Eval(data, v, nil)
 }
 
-// Eval implements the Evaluator interface.
-func (s *State) Eval(data []byte, v interface{}, input map[string]string) error {
-	spec, err := s.eval(data, input)
+// EvalFiles evaluates the files in the provided paths using the input variables and
+// populates v with the result.
+func (s *State) EvalFiles(paths []string, v interface{}, input map[string]string) error {
```
masseelch, line 74: I feel like this should take a slice of byte slices instead

rotemtam: Hcl parser had nice diagnostics that includes file names

schema/schemaspec/schemahcl/context.go
```diff
@@ -87,10 +87,21 @@ func (s *State) setInputVals(ctx *hcl.EvalContext, body hcl.Body, input map[stri
...
+func mergeCtxVar(ctx *hcl.EvalContext, key string, vals map[string]cty.Value) {
+	v, ok := ctx.Variables[key]
+	if ok {
+		v.ForEachElement(func(key cty.Value, val cty.Value) (stop bool) {
+			vals[key.AsString()] = val
+			return false
+		})
+	}
+	ctx.Variables[key] = cty.ObjectVal(vals)
+}
```
a8m, line 103: Why `var` is not a constant if it's used only here?
```suggestion
	mergeCtxVar(ctx, ctxVars)
	return nil
}

func mergeCtxVar(ctx *hcl.EvalContext, vals map[string]cty.Value) {
	const key = "var"
	v, ok := ctx.Variables[key]
	if ok {
		v.ForEachElement(func(key cty.Value, val cty.Value) (stop bool) {
			vals[key.AsString()] = val
			return false
		})
	}
	ctx.Variables[key] = cty.ObjectVal(vals)
}
```

schema/schemaspec/schemahcl/hcl.go
a8m: A `[][]byte` for the file paths? Can you elaborate on why?

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -70,12 +69,69 @@ func (s *State) UnmarshalSpec(data []byte, v interface{}) error {
...
+// using the result.
+func (s *State) eval(parsed *hclparse.Parser, v interface{}, input map[string]string) error {
+	ctx := s.config.newCtx()
+	reg := &blockDef{
+		fields:   make(map[string]struct{}),
+		children: make(map[string]*blockDef),
+	}
+	files := parsed.Files()
+	fileNames := make([]string, 0, len(files))
+	allBlocks := make([]*hclsyntax.Block, 0)
```
a8m, line 95: I'd assume there's a least one block in each file. 
```suggestion
	allBlocks := make([]*hclsyntax.Block, 0, len(files))
```

But in case you want to keep it zero, use nil array instead

```go
var allBlocks []*hclsyntax.Block
```

schema/schemaspec/schemahcl/hcl.go
```diff
@@ -70,12 +69,69 @@ func (s *State) UnmarshalSpec(data []byte, v interface{}) error {
...
+	// Prepare reg and allBlocks.
+	for name, file := range files {
+		fileNames = append(fileNames, name)
+		if err := s.setInputVals(ctx, file.Body, input); err != nil {
+			return err
+		}
+		body := file.Body.(*hclsyntax.Body)
+		for _, blk := range body.Blocks {
+			// variable definition blocks are available in the HCL source but not reachable by reference.
```
a8m, line 104:
```suggestion
			// Variable definition blocks are available in the HCL source but not reachable by reference.
```

a8m (approved):
Great stuff 🚀 

You mentioned a month or two ago that we should go with pure HCL and I'm supporting it.

PR #908, opened 2022-07-04
sql/mysql/mysqlcheck: extend data-depend analysis

a8m:


sql/mysql/internal/mysqlcheck/datadepend.go
```diff
@@ -0,0 +1,192 @@
...
+func (d *DataDepend) Analyze(ctx context.Context, p *sqlcheck.Pass) error {
+	diags := d.Diagnostics(ctx, p)
+	for _, sc := range p.File.Changes {
+		for _, c := range sc.Changes {
+			m, ok := c.(*schema.ModifyTable)
+			if !ok {
+				continue
+			}
+			for _, c := range m.Changes {
+				switch c := c.(type) {
```
a8m, line 51: Keeping it as `switch` and not `if` because in the next PRs I'll add support for column modifications (nullability, charset, etc).

sql/mysql/driver.go
```diff
@@ -6,18 +6,17 @@ package mysql
...
 	"context"
-	"embed"
-	"encoding/json"
 	"fmt"
 	"net/url"
 	"strings"
 	"time"
 
+	"ariga.io/atlas/sql/mysql/internal/mysqlversion"
+
```
masseelch, line 15:
```suggestion
```

sql/mysql/internal/mysqlcheck/datadepend_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"ariga.io/atlas/sql/mysql/internal/mysqlcheck"
+	"ariga.io/atlas/sql/schema"
+	"ariga.io/atlas/sql/sqlcheck"
+	"ariga.io/atlas/sql/sqlcheck/datadepend"
+	"ariga.io/atlas/sql/sqlclient"
+
+	"github.com/stretchr/testify/require"
+)
+
+func init() {
```
masseelch, line 22: Why is this in test?

masseelch (approved):

sql/mysql/internal/mysqlcheck/datadepend_test.go
a8m: Because I'm not sure yet if I want to register all analyzers using `init` in each package or one register function on the command line. 

I will update in next PRs.

PR #910, opened 2022-07-06
cmd/atlas: merge atlasci with atlas migrate

a8m:


cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -50,6 +58,12 @@ var (
 		RevisionSchema string
 		Force          bool
 		Verbose        bool
+		Lint           struct {
+			Format  string // log formatting.
```
masseelch, line 62: This comments are no sentences, therefore I don't like they include interpunctuation (personal pref though)

masseelch (approved):
We call them analyzers but the command is `migrate lint`

What about keeping the naming consistent?

rotemtam (approved):

PR #911, opened 2022-07-06
all: remove the schemaspec package

rotemtam:
Apologize for the huge PR. 

This PR removes the `schemaspec` package and leaves us with a simpler single-syntax implementation for the Atlas DDL based on HCL. 

Some highlights:
* Removing `Unmarshal` (interfaces, implementations, functional interfaces etc) and replacing it with `schemahcl.Eval`
* Docs updated
* Tests updated
* and more.


cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -511,6 +516,14 @@ func to(ctx context.Context, client *sqlclient.Client) (migrate.StateReader, err
 	}
 }
 
+func parseHCLBytes(b []byte) (*hclparse.Parser, error) {
```
a8m, line 519: why do you add there the type to the function? 
```suggestion
func parseHCL(b []byte) (*hclparse.Parser, error) {
```

doc/md/ddl/intro.md
```diff
@@ -160,7 +157,7 @@ show "seinfeld" {
   var test struct {
     Shows []*Show `spec:"show"`
   }
-  err := Unmarshal([]byte(f), &test)
+  err := EvalBytes([]byte(f), &test, nil)
```
a8m, line 160: Again, why do you add the type here and not just leave it as Bytes? It's not like we have a few ways to Eval, no?

a8m (approved):

PR #916, opened 2022-07-07
cmd/atlas/internal/ci: harden change loader

masseelch:
let change loader acquire a lock, handle base state, check for clean state of dev DB and clean up after itself.


sql/migrate/migrate.go
```diff
@@ -616,6 +579,46 @@ func (e *Executor) ReadState(ctx context.Context) (realm *schema.Realm, err erro
 	return
 }
 
+// IsClean checks if the given driver operates on a clean database.
+func IsClean(drv Driver, ctx context.Context) error {
```
a8m, line 583:
```suggestion
func IsClean(ctx context.Context, drv Driver) error {
```

a8m (approved):
💯

PR #917, opened 2022-07-07
cmd/atlas: change migrate apply --to flag to --url

a8m:
The idea is to stay consistent with 'schema apply' and the project file (atlas.hcl)


masseelch (commented):
Update docs as well (go generate)

masseelch (approved):

PR #922, opened 2022-07-08
cmd/atlas/internal/cmdapi: check revisions table existence instead of…

masseelch:
… silently creating it if it does not exist


cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -204,7 +206,8 @@ func init() {
 	MigrateCmd.PersistentFlags().SortFlags = false
 	// Apply flags.
 	MigrateApplyCmd.Flags().StringVarP(&MigrateFlags.LogFormat, migrateFlagLog, "", logFormatTTY, "log format to use")
-	MigrateApplyCmd.Flags().StringVarP(&MigrateFlags.RevisionSchema, migrateFlagRevisionsSchema, "", "", "schema name where the revisions table is to be created")
+	MigrateApplyCmd.Flags().StringVarP(&MigrateFlags.RevisionSchema, migrateFlagRevisionsSchema, "", entmigrate.DefaultRevisionSchema, "schema name where the revisions table resides")
+	MigrateApplyCmd.Flags().BoolVarP(&MigrateFlags.CreateRevisions, migrateFlagCreateRevisions, "", false, "whether to create revisions schema and table if they don't exist yet")
```
a8m, line 210:
```suggestion
	MigrateApplyCmd.Flags().BoolVarP(&MigrateFlags.CreateRevisions, migrateFlagCreateRevisions, "", false, "whether to create revisions schema and table if they do not exist yet")
```

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -525,6 +535,36 @@ func to(ctx context.Context, client *sqlclient.Client) (migrate.StateReader, err
...
+	}
+	// Inspect schema and check if the table does already exist.
+	s, err := sc.InspectSchema(ctx, "", &schema.InspectOptions{Tables: []string{revision.Table}})
+	switch {
+	case err != nil && !schema.IsNotExistError(err):
+		return false, err
+	case schema.IsNotExistError(err):
+		// Schema does not exist.
+		return false, nil
+	}
```
a8m, line 559:
```suggestion
	if err != nil {
		if schema.IsNotExistError(err) {
			err = nil
		}
		return false, err
	}
```
More level of nesting, but feels simpler to read it this way imo.

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -525,6 +535,36 @@ func to(ctx context.Context, client *sqlclient.Client) (migrate.StateReader, err
...
+	if _, ok := s.Table(revision.Table); !ok {
+		// Table does not exist.
+		return false, nil
+	}
+	// Schema and Table are present.
+	return true, nil
```
a8m, line 565:
```suggestion
	_, ok := s.Table(revision.Table)
	// Indicate if both schema and tables exist.
	return ok, nil
```

ditto

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -42,6 +42,7 @@ const (
 	migrateFlagFormat          = "format"
 	migrateFlagLog             = "log"
 	migrateFlagRevisionsSchema = "revisions-schema"
+	migrateFlagCreateRevisions = "create-revisions"
```
a8m, line 45: Don't you think `create-revisions` is not that correct for this flag? Revision is a migration revision but this flag indicates we create the schema and table for storing the revisions. 

cc @rotemtam and @hilakashai for help.

a8m (commented):
The code looks good, I have one concern about the flag name, but Rotem and Hila can help with it.

cmd/atlas/internal/cmdapi/migrate.go
a8m, line 45: Maybe, maybe, `rev-schema` and `create-rev-table` work better together.

rotemtam: Create rev table

masseelch: `create-revisions-table`? I don't like abbreviations in parts of a flag.

PR #923, opened 2022-07-08
cmd/atlas/internal/cmdapi: add dry-run flag to migrate apply

masseelch:
Stacked on #922.

I (and users as well) need an easy way to see what files / statements will be run on the connected database. The `--dry-run` flag enables this by only printing but not executing statements.


a8m (approved):
Looks good.

PR #927, opened 2022-07-09
doc/md: updating install from source instructions

Phuurl:
Current instructions won't install `atlas` in a usable form, and are more for using Atlas in other ways.

This PR updates the docs to use the `go install` command instead, along with specifying version information, which is more likely to achieve what users are expecting when following other links in the docs to this section.


a8m (approved):
💯

a8m:
Thanks for the contribution @Phuurl 🙏

PR #899, opened 2022-06-30
cmd/atlascmd: add supported format for dbmate

iwata:
## What

SSIA

## Why

To integrate with [amacneil/dbmate](https://github.com/amacneil/dbmate) easily.

PR #931, opened 2022-07-11
cmd/atlascmd: add supported format for dbmate

iwata:
to use github.com/amacneil/dbmate easily

Re-open #899


a8m:
Tests fail on 1.17 because it does not recognize the `go.work` file and does not use the `replace` directive there. 

I'm OK with running the unit tests only on 1.18 only. @rotemtam, @masseelch - WDYT?

PR #932, opened 2022-07-11
cmd/atlas/cmdapi: add migrate status command

masseelch:
Adds `atlas migrate status` command to print information about the migration status of a connected database.

image
image
image

### DIFF
--- a/cmd/atlas/internal/cmdapi/migrate.go
+++ b/cmd/atlas/internal/cmdapi/migrate.go
@@ -433,7 +433,7 @@ func CmdMigrateStatusRun(cmd *cobra.Command, _ []string) error {
 		if err != nil {
 			return err
 		}
-		return errors.New("revisions table does not exist")
+		return statusPrint(cmd.OutOrStdout(), sc, avail, nil, nil)
 	}
 	// Currently, only in DB revisions are supported.
 	opts := []entmigrate.Option{entmigrate.WithSchema(MigrateFlags.RevisionSchema)}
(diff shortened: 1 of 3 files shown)


cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -396,6 +411,123 @@ func CmdMigrateNewRun(_ *cobra.Command, args []string) error {
...
+		return err
+	}
+	dirty := migrate.ErrDirty{}
+	pending, err := ex.Pending(context.Background())
+	if err != nil && !errors.As(err, &dirty) && !errors.Is(err, migrate.ErrNoPendingFiles) {
+		return err
+	}
+	var (
+		applied                 = avail[:len(pending)+1]
+		prev, cur, next, latest string
```
a8m, line 459: I don't really understand what each state means. Can you elaborate, please?

masseelch: `prev`: previously applied version (like `20221010235500`)
`cur ` : currently applied version
`next`: next version in line
`latest`: latest version in the migration directory

masseelch: Well, we have a `--dry-run` flag as well even though we could simply look inside of the files. This is nice-to-have. But I don't have strong feelings about those two. The important part is to know the migration state and what currently is applied.

a8m: Can I propose starting with `current` which is the latest applied ("current database version"), and `next` for "next migration file to apply"? Then, we can play with it a bit and improve it if it's needed.

a8m: cc @rotemtam

masseelch: image

a8m (approved):

PR #937, opened 2022-07-12
sql/sqlcheck: move analyzers to be driver-specific

a8m:
Also, detect modify tables changes and ignore report them as destructive. See examples for before/after:

image

image


sql/mysql/mysqlcheck/mysqlcheck.go
```diff
@@ -15,14 +15,15 @@ import (
...
 type (
-	// DataDependOptions defines the configuration options
-	// for the data-dependent changes checker.
+	// DataDependOptions defines the additional configuration
+	// options for the data-dependent changes checker.
 	DataDependOptions struct {
-		// NotNull indicates if the analyzer should check for modification or
-		// addition of not null constraints to
+		// NotNull indicates if the analyzer should check for modification
+		// or addition of not null constraints to columns.
```
masseelch, line 26:
```suggestion
		// or addition of NOT NULL constraints to columns.
```

sql/mysql/mysqlcheck/mysqlcheck.go
```diff
@@ -15,14 +15,15 @@ import (
...
 		NotNull *bool `spec:"not_null,omitempty"`
```
masseelch, line 27: Why is a pointer needed here?

sql/mysql/mysqlcheck/mysqlcheck.go
```diff
@@ -188,3 +182,16 @@ func (d *DataDepend) addColumn(p *sqlcheck.Pass, change *sqlcheck.Change, t *sch
...
+	sqlcheck.Register(mysql.DriverName, func(r *schemahcl.Resource) (sqlcheck.Analyzer, error) {
+		az, err := NewDataDepend(r)
+		if err != nil {
+			return nil, err
+		}
+		return sqlcheck.Analyzers{
+			destructive.New(destructive.Options{}),
+			az,
+		}, nil
+	})
```
masseelch, line 196: This means runtime errors instead of compile time, right?

masseelch (approved):

sql/mysql/mysqlcheck/mysqlcheck.go
a8m: Both happen on initializations because an error is returned in the case of an invalid configuration.

PR #938, opened 2022-07-13
sql/sqlcheck: add datadepend checks for postgres and sqlite

a8m:


internal/integration/testdata/sqlite/cli-migrate-lint-add-notnull.txt
```diff
@@ -0,0 +1,19 @@
...
+
+-- migrations/2.sql --
+ALTER TABLE users ADD COLUMN c2 int NOT NULL;
+
+ALTER TABLE users ADD COLUMN c3 int NOT NULL DEFAULT 1;
+
+-- expected1.txt --
+Data dependent changes detected in file 2.sql:
+
+	L1: Adding a non-nullable "int" column "c2" will fail in case table "users" is not empty
```
masseelch, line 18: Wondering if we should also include the SQL in all reports here: 

```
L1: ALTER TABLE users ADD COLUMN c1 int NOT NULL
    Adding a non-nullable "int" column "c2" will fail in case table "users" is not empty
    
...
```

masseelch (approved):
:rocket:

internal/integration/testdata/sqlite/cli-migrate-lint-add-notnull.txt
a8m: I don't know yet. I plan to improve the printing soon to look more like the `migrate apply` output. I will ask @hilakashai to help with its design.

PR #939, opened 2022-07-13
sql/migrate: change execution logic

masseelch:
Instead of having a state for a migration execution like `ok`, `error`, ... we now have a counter for every executed statement. This will increase for every successfully applied statement. Every next attempt to migrate, will pick up the file where it left off and attempt to execute the new statement.

File has error: 
```
ALTER TABLE `tbl` ADD `col_3` bigint;
aALTER T sABLE `tbl` ADD `col4` bigint;
```
image

After fixing file:
```
ALTER TABLE `tbl` ADD `col_3` bigint;
ALTER TABLE `tbl` ADD `col4` bigint;
```
image

Another attempt:
image


cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -297,40 +297,77 @@ func CmdMigrateApplyRun(cmd *cobra.Command, args []string) error {
...
-		tx, err = c.Tx(cmd.Context(), nil)
+	}
+	for _, f := range pending {
+		if !MigrateFlags.DryRun {
+			// Wrap the file execution in one transaction.
+			tx, err = c.Tx(cmd.Context(), nil)
+			if err != nil {
+				return err
+			}
+			drv = tx.Driver
```
a8m, line 350: What happens in the next iteration when calling `c.Tx()`? Also, we can consider using checkpoints in the future.

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -845,14 +860,15 @@ type LogTTY struct {
...
-	indent4 = strings.Repeat(indent2, 2)
+	cyan         = color.CyanString
+	green        = color.HiGreenString
+	red          = color.HiRedString
+	redBgWhiteFg = color.New(color.FgHiWhite, color.BgHiRed).SprintFunc()
+	yellow       = color.YellowString
+	dash         = yellow("--")
+	arr          = cyan("->")
+	indent2      = strings.Repeat(" ", 2)
+	indent4      = strings.Repeat(indent2, 2)
```
a8m, line 871:
```suggestion
	indent2      = "  "
	indent4      = indent2+indent2
```

😜

sql/migrate/migrate.go
```diff
@@ -436,79 +418,111 @@ func (e *Executor) Pending(ctx context.Context) ([]File, error) {
...
 	}
-	stmts, err := sc.Stmts(m)
+	stmts, err := e.dir.(Scanner).Stmts(m)
 	if err != nil {
-		return r.setGoErr(fmt.Errorf("sql/migrate: execute: scanning statements from file %q: %w", m.Name(), err))
+		return fmt.Errorf("sql/migrate: execute: scanning statements from %q: %w", m.Name(), err)
+	}
+	// If there already is a revision with this version in the database,
+	// and it is partially applied, continue where the last attempt was left off.
+	r, err := e.rrw.ReadRevision(ctx, version)
```
a8m, line 478: Don't we want to ensure the file wasn't changed? e.g. compare hashes?

In case it was changed, should we fail or apply it from the first statement?

sql/migrate/migrate.go
```diff
@@ -436,79 +418,111 @@ func (e *Executor) Pending(ctx context.Context) ([]File, error) {
...
+			err = wrap(err2, err)
 		}
-		if _, err := e.drv.ExecContext(ctx, stmt); err != nil {
-			return r.setSQLErr(
-				fmt.Errorf("sql/migrate: execute: executing statement %q from version %q: %w", stmt, r.Version, err),
-				stmt,
-			)
+	}(ctx, e.rrw, r)
+	e.log.Log(LogFile{r.Version, r.Description, r.Applied})
+	for _, stmt := range stmts[r.Applied:r.Total] {
```
a8m, line 503: See my question above + ensure we don't get an out-of-range error.

cmd/atlas/internal/cmdapi/migrate.go
a8m: Ignore this.

PR #947, opened 2022-07-14
sql/migrate: utilize partial hashes to check for changes in applied s…

masseelch:
…tatements of partially applied files

If a file has been partially applied, we now check if the already applied part of a migration file was edited and refuse to work, if the history was changed.

image

### DIFF
--- a/sql/migrate/migrate.go
+++ b/sql/migrate/migrate.go
@@ -221,6 +221,8 @@ type (
 		Error string
 		// Hash is the check-sum of this migration as stated by the migration directories HashFile.
 		Hash string
+		// PartialHashes contains one hash per statement that has been applied on the database.
+		PartialHashes []string
 		// OperatorVersion holds a string representation of the Atlas operator managing this database migration.
 		OperatorVersion string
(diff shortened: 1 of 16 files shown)


sql/migrate/migrate.go
```diff
@@ -499,18 +512,26 @@ func (e *Executor) Execute(ctx context.Context, m File) (err error) {
...
 	}(ctx, e.rrw, r)
-	e.log.Log(LogFile{r.Version, r.Description, r.Applied})
-	for i, stmt := range stmts {
-		if i < r.Applied {
-			// Skip if already applied.
-			continue
+	if r.Applied > 0 {
+		// If the file has been applied partially before, check if the
+		// applied history didn't change, as this is dangerous.
+		for i := 0; i < r.Applied; i++ {
```
a8m, line 518:
```suggestion
		// If the file has been applied partially before, check if the
		// applied statements were not changed.
		for i := 0; i < r.Applied; i++ {
```

No need for "dangerous". Sounds too dramatic.

a8m (approved):

PR #967, opened 2022-07-19
sql/postgres: support the alias float type

a8m:


rotemtam (approved):
LGTM, maybe add refs to the Postgres docs like you did in MySQL?

PR #969, opened 2022-07-20
doc: add integrations

hilakashai:
* added integrations section
* refactored the Go API to be in one page
* rerouted the Go API pages to the new page
* rerouted the previous TF page to the new page


a8m (approved):
💯

PR #974, opened 2022-07-21
sql/postgres: delete dangling enum types

a8m:


sql/postgres/convert.go
```diff
@@ -181,6 +181,86 @@ func ParseType(typ string) (schema.Type, error) {
 	return t, nil
 }
 
+func columnType(c *columnDesc) (schema.Type, error) {
```
a8m, line 184: Move from a `inspect.go` to `convert.go` close to the rest of the parsing/formatting functions.

sql/postgres/diff.go
```diff
@@ -254,56 +254,6 @@ func (d *diff) typeChanged(from, to *schema.Column) (bool, error) {
 	return changed, nil
 }
 
-// Normalize implements the sqlx.Normalizer interface.
-func (d *diff) Normalize(from, to *schema.Table) error {
```
a8m, line 258: Code is not needed anymore. Already handled by `FormatType`.

giautm (approved):
LGTM, <3

PR #977, opened 2022-07-23
chore: add preallocations and make small cleanup

sashamelentyev:


a8m:
While I appreciate your contribution, I don’t think these small optimizations are needed in this case. Atlas is used to run database migration, inspection, and other schema management work. i.e. it's not something that sits on any critical path (executed as a CLI).

Cleanups are fine to land, but maps/array/other pre-allocations just add extra complexity that is not needed here. To make it clear, I don't mind if someone adds such code when a new feature or fix is created, I just don't want to make it a standard in this repository and prefer to keep our codebase consistent.

sql/postgres/postgrescheck/postgrescheck.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 // NewDataDepend creates new data-depend analyzer.
-func NewDataDepend(*schemahcl.Resource) *datadepend.Analyzer {
+func NewDataDepend() *datadepend.Analyzer {
```
a8m, line 18: This one is needed for other branches I'm working on. Revert, please.

sql/mysql/tidb.go
```diff
@@ -249,7 +249,7 @@ func (i *tinspect) setFKs(s *schema.Schema, t *schema.Table) error {
 }
 
 // columns from the matched regex above.
-func columns(schema *schema.Schema, s string) []string {
+func columns(s string) []string {
```
a8m, line 252: 👍

sql/postgres/postgrescheck/postgrescheck.go
sashamelentyev: Done

PR #980, opened 2022-07-23
schemahcl: fixed invalid identity token

giautm:


rotemtam (commented):
Well done! Can you add an integration test using testscript to verify this works e2e on postgres/mysql?

giautm:
> Well done! Can you add an integration test using testscript to verify this works e2e on postgres/mysql?

yes, please me try. :D

schemahcl/hcl.go
```diff
@@ -541,6 +538,49 @@ func findAttr(attrs []*Attr, k string) (*Attr, bool) {
 	return nil, false
 }
 
+func hclRefTokens(ref string) hclwrite.Tokens {
+	s := strings.Split(strings.ReplaceAll(ref, "$", ""), ".")
+	t := make([]*hclwrite.Token, 0, len(s))
+	for i, s := range s {
+		if i == 0 {
```
rotemtam, line 545: Maybe add comments explaining the cases

schemahcl/hcl.go
```diff
@@ -541,6 +538,49 @@ func findAttr(attrs []*Attr, k string) (*Attr, bool) {
...
+		} else if hclsyntax.ValidIdentifier(s) {
+			t = append(t, hclTraverseAttr(s)...)
+		} else {
+			t = append(t, hclTraverseIndex(cty.StringVal(s))...)
+		}
+	}
+	return t
+}
+
+func hclTraverseAttr(name string) hclwrite.Tokens {
```
rotemtam, line 556: Why "traverse"?

giautm: I used the term from `NewExpressionAbsTraversal`  https://github.com/hashicorp/hcl/blob/v2.13.0/hclwrite/ast_expression.go#L87-L119

schemahcl/hcl.go
giautm: Root element in NewExpressionAbsTraversal

internal/integration/testdata/mysql/column-add-drop-2.txt
```diff
@@ -0,0 +1,110 @@
...
+# Compare database with 2.sql.
+cmpshow users 2.sql
+
+# Apply schema "1.hcl" should migrate database to previous state.
+apply 1.hcl
+cmpshow users 1.sql
+
+# Drop table.
+apply 0.hcl
+! exist users
```
a8m, line 26: Add another assertion using `cmphcl` to ensure HCL writing is correct. Also, no need for another test, you can extend the current one.

schemahcl/hcl.go
```diff
@@ -541,6 +538,49 @@ func findAttr(attrs []*Attr, k string) (*Attr, bool) {
...
+	s := strings.Split(strings.ReplaceAll(ref, "$", ""), ".")
+	t := make([]*hclwrite.Token, 0, len(s))
+	for i, s := range s {
+		if i == 0 {
+			t = append(t, hclRawTokens(s)...)
+		} else if hclsyntax.ValidIdentifier(s) {
+			t = append(t, hclTraverseAttr(s)...)
+		} else {
+			t = append(t, hclTraverseIndex(cty.StringVal(s))...)
+		}
```
a8m, line 551: Use `switch` and inline `hclTraverseAttr` as it's used only here.

schemahcl/hcl.go
```diff
@@ -541,6 +538,49 @@ func findAttr(attrs []*Attr, k string) (*Attr, bool) {
...
+			Bytes: []byte{'.'},
+		},
+		&hclwrite.Token{
+			Type:  hclsyntax.TokenIdent,
+			Bytes: []byte(name),
+		},
+	}
+}
+
+func hclTraverseIndex(v cty.Value) hclwrite.Tokens {
```
a8m, line 569: ditto.

schemahcl/hcl.go
```diff
@@ -541,6 +538,49 @@ func findAttr(attrs []*Attr, k string) (*Attr, bool) {
 	return nil, false
 }
 
+func hclRefTokens(ref string) hclwrite.Tokens {
+	s := strings.Split(strings.ReplaceAll(ref, "$", ""), ".")
```
a8m, line 542: `"c $"` is also a valid (quoted) identifier. Let's handle it as well?

a8m (approved):
🚀

a8m (changes requested):
Accidentally approved 😃

internal/integration/testdata/mysql/column-add-drop-2.txt
a8m: You can extend this one: https://github.com/ariga/atlas/blob/master/internal/integration/testdata/mysql/column-generated-inspect.txt#L5

schemahcl/hcl.go
```diff
@@ -541,6 +538,39 @@ func findAttr(attrs []*Attr, k string) (*Attr, bool) {
 	return nil, false
 }
 
+func hclRefTokens(ref string) hclwrite.Tokens {
+	t := []*hclwrite.Token{}
```
a8m, line 542:
```suggestion
	var t []*hclwrite.Token
```

a8m (approved):
:shipit: ship it!

PR #984, opened 2022-07-25
sql/migrate: merge Dir and Scanner to one interface

a8m:
The next steps are supporting `Baseline` file(s), `Tags`, and `flyway` directory.


cmd/atlas/internal/cmdapi/migrate_test.go
```diff
@@ -582,3 +588,12 @@ func countFiles(t *testing.T, p string) int {
 	require.NoError(t, err)
 	return len(files)
 }
+
+func sed(t *testing.T, r, p string) {
+	args := []string{"-i"}
+	if runtime.GOOS == "darwin" {
+		args = append(args, ".bk")
+	}
```
a8m, line 596: It fails locally on darwin.

masseelch: Darwin has no GNU sed. You need to install it.

sql/migrate/migrate.go
```diff
@@ -140,11 +140,14 @@ func Conn(drv Driver, opts *schema.InspectRealmOption) StateReader {
 }
 
 type (
-	// Dir describes the methods needed for a Planner to manage migration files.
+	// Dir wraps the functionality described a migration directory.
```
rotemtam, line 143: That describes ?

rotemtam (approved):
Minor godoc fix

sql/migrate/migrate.go
a8m, line 143:
```suggestion
	// Dir wraps the functionality used to interact with a migration directory.
```

PR #995, opened 2022-08-02
cmd/atlas/internal: auto populate parseTime option for mysql connections

a8m:


rotemtam (approved):
LGTM 
Maybe a testscript test to verify this doesn't break in the future ?

a8m:
mysql migration is not tested correctly in testscript. I'll add it when I get to it.

Issue #1001, opened 2022-08-03
Index Out Of Bounds Error - atlas apply on crafted hcl document


TopherIsSwell:
I'm not sure what causes this or why, but I can consistently get an index OOB panic from a given (apparently valid) HCL doc. I don't know anything about HCL or atlas, but messing around with it I was able to produce this error which seems like it could be a security issue. Even if the HCL doc is invalid, I feel like this should be caught before attempting to index a null array. 

This hcl document was created from a sqlite3 database and was then applied to a mysql db to discover this bug.

## Expected Result
I would expect that any HCL document (especially a valid one) that can't be applied, would output an well formed error and exit with an error code, without attempting any pointer derefs that may be invalid.

## Observed Result
On sending a specially crafted HCL document to an empty MySQL database, I get a index out-of-bounds panic. While the document may not be valid (it appears to work on sqlite3 targets), indexing out of bounds seems like a bug that should be caught and this document should just be refused instead.

Output from `atlas schema apply`
```
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
ariga.io/atlas/sql/mysql.parseColumn({0x0?, 0xb?})
	/home/topher/go/pkg/mod/ariga.io/atlas@v0.4.3-0.20220713104915-cda1480dd791/sql/mysql/inspect.go:595 +0x45c
ariga.io/atlas/sql/mysql.ParseType({0x0, 0x0})
	/home/topher/go/pkg/mod/ariga.io/atlas@v0.4.3-0.20220713104915-cda1480dd791/sql/mysql/convert.go:109 +0x3e
ariga.io/atlas/schemahcl.(*TypeRegistry).Type(0xc0000a5e60, 0xc000314f60, {0x0, 0x0, 0x1?})
	/home/topher/go/pkg/mod/ariga.io/atlas@v0.4.3-0.20220713104915-cda1480dd791/schemahcl/types.go:285 +0x135
ariga.io/atlas/sql/mysql.convertColumnType(0x8?)
	/home/topher/go/pkg/mod/ariga.io/atlas@v0.4.3-0.20220713104915-cda1480dd791/sql/mysql/sqlspec.go:193 +0x33
ariga.io/atlas/sql/internal/specutil.Column(0xc0000b8d80, 0xc899a8)
	/home/topher/go/pkg/mod/ariga.io/atlas@v0.4.3-0.20220713104915-cda1480dd791/sql/internal/specutil/convert.go:146 +0x215
[...]

PR #1008, opened 2022-08-03
schemahcl: dont allow empty str sql() expr

rotemtam:
Closes https://github.com/ariga/atlas/issues/1001.

### DIFF
--- a/schemahcl/opts.go
+++ b/schemahcl/opts.go
@@ -5,6 +5,7 @@
 package schemahcl
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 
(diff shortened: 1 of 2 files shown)


schemahcl/opts.go
```diff
@@ -103,7 +103,11 @@ func rawExprImpl() function.Function {
 		},
 		Type: function.StaticReturnType(ctyRawExpr),
 		Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
-			t := &RawExpr{X: args[0].AsString()}
+			x := args[0].AsString()
+			if len(x) == 0 {
+				return cty.NilVal, fmt.Errorf("invalid expr: %q", x)
+			}
```
a8m, line 109:
```suggestion
			if len(x) == 0 {
				return cty.NilVal, errors.New("empty expression")
			}
```

a8m (approved):

schemahcl/opts.go
a8m: empty or missing, but there's no need to quote an empty one, right?

PR #1016, opened 2022-08-04
sql/migrate: store baseline migrations in history table

a8m:


sql/migrate/migrate.go
```diff
@@ -182,10 +184,11 @@ type (
 		PartialHashes []string
 		// OperatorVersion holds a string representation of the Atlas operator managing this database migration.
 		OperatorVersion string
-		// Meta holds additional custom meta-data given for this migration.
-		Meta map[string]string
 	}
 
+	// RevisionType defines the type of the revision record in the history table.
+	RevisionType uint
```
rotemtam, line 190: Why not use booleans ? We lose the capability to use the db for filtering nicely this way . And disk  isn't a major concern here

a8m: Because the state is not binary here. There are baseline, exec, resolved, and execs that were resolved. 

Also:
image


We can use `GoType` to avoid this convertion, but I didn't feel it's needed atm - we can always change this, it's an application thing.

rotemtam (approved):

PR #1027, opened 2022-08-08
internal/integration: add testscript for cli migrate diff

masseelch:


internal/integration/postgres_test.go
```diff
@@ -1119,11 +1119,14 @@ create table atlas_types_sanity
 }
 
 func (t *pgTest) url() string {
-	return t.dsn()
+	return t.dsn("")
 }
 
-func (t *pgTest) dsn() string {
-	return fmt.Sprintf("postgres://postgres:pass@localhost:%d/test?sslmode=disable", t.port)
+func (t *pgTest) dsn(dbname string) string {
```
a8m, line 1125: in Postgres, that's a schema name. the database name is "test". It's a bit different from MySQL.

a8m (approved):
Looks good, but I don't see any new tests here.

internal/integration/testdata/mysql/cli-migrate-diff.txt
```diff
@@ -0,0 +1,76 @@
...
+table "users" {
+  schema = schema.script_cli_migrate_diff
+  column "id" {
+    null = false
+    type = bigint
+    auto_increment = true
+  }
+  primary_key  {
+    columns = [column.id]
+  }
```
a8m, line 27: You can configure the `charset` and `collate` (e.g. `latin1` `latin1_swedish_ci`) to avoid the duplicate SQL files below.

PR #1028, opened 2022-08-08
sql/mysql: report error in case table has no columns

a8m:
Avoid getting parse error from MySQL; first column is required by the grammar


ronenlu (approved):

rotemtam (commented):
💯

PR #1031, opened 2022-08-09
internal: database connections for a test are established in the test…

masseelch:
…s itself.

With this changes it is no longer required to have all integration containers running if one only works on one dialect or version. This works by only then spinning up a connection to a database if it is requested the first time. Subsequent calls re-use the existing connection.

If a connection for a container is requested and the container is not running, there still will be an error and the test will fail.

CI has been changed to respect those changes.

### DIFF
--- a/internal/integration/cockroach_test.go
+++ b/internal/integration/cockroach_test.go
@@ -31,11 +31,11 @@ type crdbTest struct {
 	rrw     migrate.RevisionReadWriter
 	version string
 	port    int
-	once    *sync.Once
+	once    sync.Once
 }
 
 var crdbTests = map[string]*crdbTest{
-	"cockroach": {port: 26257, once: &sync.Once{}},
+	"cockroach": {port: 26257},
 }
 
 func crdbRun(t *testing.T, fn func(*crdbTest)) {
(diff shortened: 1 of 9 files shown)


internal/integration/cockroach_test.go
```diff
@@ -31,50 +31,35 @@ type crdbTest struct {
 	rrw     migrate.RevisionReadWriter
 	version string
 	port    int
+	once    *sync.Once
```
a8m, line 34: Pointer is no needed here. Use the default value when creating the struct

internal/integration/postgres_test.go
```diff
@@ -30,53 +30,39 @@ type pgTest struct {
 	rrw     migrate.RevisionReadWriter
 	version string
 	port    int
+	once    *sync.Once
```
a8m, line 33: Ditto.

internal/integration/tidb_test.go
```diff
@@ -25,45 +25,33 @@ import (
...
-		drv, err := mysql.Open(db)
-		if err != nil {
-			log.Fatalln(err)
-		}
-		tidbTests.drivers[version] = &myTest{db: db, drv: drv, version: version, port: port, rrw: &rrw{}}
-	}
-	return cs
+var tidbTests = map[string]*myTest{
+	"tidb5": {port: 4309, once: &sync.Once{}},
+	"tidb6": {port: 4310, once: &sync.Once{}},
```
a8m, line 30: Ditto

a8m (approved):

PR #1035, opened 2022-08-09
sql/migrate: skip adding table qualifiers when connection/work is scoped to a schema

a8m:
This is almost done.

Changes to be made in this PR:
- Add tests for `migrate diff` in MySQL/PostgreSQL 

Changes to be added in the next PRs:
- Add additional validation on the driver side; disallow using this feature programmatically, when the Atlas schema is not scoped to 1 database schema.
- Accept additional flag to allow changing the schema.
- Add docs and examples for Ent and golang-migrate.

### DIFF
--- a/cmd/atlas/internal/cmdapi/migrate.go
+++ b/cmd/atlas/internal/cmdapi/migrate.go
@@ -429,7 +429,8 @@ func CmdMigrateDiffRun(cmd *cobra.Command, args []string) error {
 	}
 	opts := []migrate.PlannerOption{migrate.PlanFormat(f)}
 	if dev.URL.Schema != "" {
-		opts = append(opts, migrate.PlanSchema(dev.URL.Schema))
+		// Disable tables qualifier in schema-mode.
+		opts = append(opts, migrate.PlanWithSchemaQualifier(""))
 	}
 	// Plan the changes and create a new migration file.
 	pl := migrate.NewPlanner(dev.Driver, dir, opts...)
@@ -437,8 +438,14 @@ func CmdMigrateDiffRun(cmd *cobra.Command, args []string) error {
 	if len(args) > 0 {
 		name = args[0]
 	}
+	plan, err := func() (*migrate.Plan, error) {
+		if dev.URL.Schema != "" {
+			return pl.PlanSchema(cmd.Context(), name, desired.StateReader)
+		}
+		return pl.Plan(cmd.Context(), name, desired.StateReader)
+	}()
 	var cerr migrate.NotCleanError
-	switch plan, err := pl.Plan(cmd.Context(), name, desired.StateReader); {
+	switch {
 	case errors.Is(err, migrate.ErrNoPlan):
 		cmd.Println("The migration directory is synced with the desired state, no changes to be made")
 		return nil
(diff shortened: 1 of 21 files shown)


giautm (commented):
💯

internal/integration/testdata/mysql/cli-migrate-diff.txt
```diff
@@ -57,16 +57,16 @@ table "users" {
 
 -- 2.sql --
 -- modify "users" table
-ALTER TABLE `script_cli_migrate_diff`.`users` ADD COLUMN `create_time` timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6);
+ALTER TABLE `users` ADD COLUMN `create_time` timestamp(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6);
```
rotemtam, line 60: 💯 
😍

rotemtam (approved):

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -429,16 +429,23 @@ func CmdMigrateDiffRun(cmd *cobra.Command, args []string) error {
 	}
 	opts := []migrate.PlannerOption{migrate.PlanFormat(f)}
 	if dev.URL.Schema != "" {
-		opts = append(opts, migrate.PlanSchema(dev.URL.Schema))
+		// Disable tables qualifier in schema-mode.
+		opts = append(opts, migrate.PlanWithSchemaQualifier(""))
```
masseelch, line 433: I think the API is a little weird here. Atlas does plan with qualifier by default, correct? Why not have another option `PlanWithoutSchemaQualifier()` (that maybe call the above)?

sql/migrate/migrate.go
```diff
@@ -263,37 +278,48 @@ func PlanFormat(fmt Formatter) PlannerOption {
...
 }
 
-// PlanWithoutChecksum disables the hash-sum functionality for the migration directory.
-func PlanWithoutChecksum() PlannerOption {
+// PlanWithChecksum allows setting if the hash-sum functionality
+// for the migration directory is enabled or not.
+func PlanWithChecksum(b bool) PlannerOption {
 	return func(p *Planner) {
-		p.sum = false
+		p.sum = b
```
masseelch, line 285: TBH, I don't like this. Because setting `PlanWithChecksum(true)` is a no-op.

masseelch (approved):
:bear:

cmd/atlas/internal/cmdapi/migrate.go
a8m: > Atlas does plan with qualifier by default, correct?

Only in realm mode. 

> PlanWithoutSchemaQualifier

hehe, that's funny, because I started the code this way, but since I want to provide a flag (with default to ""), I decided to keep only one option.

sql/migrate/migrate.go
a8m: Kinda agree, but I thought about how to simplify the code by passing the flag directly, I can revert it. 

```
PlanWithChecksum(!MigrateFlag.DisableChecksum)
```

cmd/atlas/internal/cmdapi/migrate.go
masseelch: oki

sql/migrate/migrate.go
masseelch: I think this is a case where we "gain" 2 less lines of code without any real advantage. 

```go
if MigrateFlag.DisableChecksum {
  opts = append(opts, PlanWithoutChecksum())
}
```

Does not seem much simpler, except maybe getting rid of the `opts` variable.

PR #1058, opened 2022-08-16
sql/sqltool: support replaying of flyway migration files

masseelch:
Waits on #1057


a8m (approved):
Fix CI before landing.

PR #1060, opened 2022-08-16
cmd/atlas: make schema diff commands work with realms

ronenlu:
currently the command: `atlas schema diff --from mysql://root:pass@localhost:3306 --to mysql://root:pass@localhost:33`
will return: `Error: mysql: schema "" was not found`


cmd/atlas/internal/cmdapi/diff.go
```diff
@@ -54,19 +55,30 @@ func cmdDiffRun(cmd *cobra.Command, flags *diffCmdOpts) {
...
-	toSchema, err := toC.InspectSchema(ctx, toName, nil)
-	cobra.CheckErr(err)
-	// SchemaDiff checks for name equality which is irrelevant in the case
-	// the user wants to compare their contents, if the names are different
-	// we reset them to allow the comparison.
-	if fromName != toName {
-		toSchema.Name = ""
-		fromSchema.Name = ""
+	var diff []schema.Change
+	// if one of schema names is empty we compare realms
```
a8m, line 59: Add validation that both connections/urls are the same type. Schema or a realm.

cmd/atlas/internal/cmdapi/diff.go
```diff
@@ -54,19 +56,33 @@ func cmdDiffRun(cmd *cobra.Command, flags *diffCmdOpts) {
...
+	if fromName == "" || toName == "" {
+		if fromName != toName {
+			cobra.CheckErr(fmt.Errorf("can not diff realm with schame, from schema: %s, to schema: %s", fromName, toName))
```
masseelch, line 63:
```suggestion
			cobra.CheckErr(fmt.Errorf("can not diff realm with schema, from schema: %s, to schema: %s", fromName, toName))
```

cmd/atlas/internal/cmdapi/diff.go
```diff
@@ -54,19 +56,33 @@ func cmdDiffRun(cmd *cobra.Command, flags *diffCmdOpts) {
 	defer toC.Close()
 	fromName := fromC.URL.Schema
 	toName := toC.URL.Schema
```
a8m, line 58: toName and fromName and not good names anymore. Use `fromS`, `toS` instead.

cmd/atlas/internal/cmdapi/diff.go
```diff
@@ -54,19 +56,33 @@ func cmdDiffRun(cmd *cobra.Command, flags *diffCmdOpts) {
...
-	cobra.CheckErr(err)
-	// SchemaDiff checks for name equality which is irrelevant in the case
-	// the user wants to compare their contents, if the names are different
-	// we reset them to allow the comparison.
-	if fromName != toName {
-		toSchema.Name = ""
-		fromSchema.Name = ""
+	var diff []schema.Change
+	// if one of schema names is empty we compare realms
+	if fromName == "" || toName == "" {
```
a8m, line 61: switch statements are powerful in Go, use it here to make the code simpler.

```go
switch {
case fromS == "" && toS == "":
// compare realm.
case fromS == "":
	return fmt.Errorf("cannot diff schema %q with a database connection", fromS)
case toS == "":
	return fmt.Errorf("cannot diff a database connection with a schema %q ", toS)
default: // compare schemas.
}
```

a8m (approved):
💯 

Can you add testscript for this in another PR? Let me know if you need any help.

PR #1063, opened 2022-08-17
sql/sqltool: support replay of existing flyway migration directories

masseelch:
This PR is the first step in understanding existing Flyway migration directories and therefore in "importing" them into the Atlas migration file format.

By default, Flyway skips hidden directories. For now, this PR only skips hidden directories on *nix machines. Windows does not use the `.` prefix for marking directories as hidden. Windows code remains untested.


sql/sqltool/hidden_windows.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2021-present The Atlas Authors. All rights reserved.
+// This source code is licensed under the Apache 2.0 license found
+// in the LICENSE file in the root directory of this source tree.
+
+package sqltool
```
a8m, line 5: No build flag for windows?

sql/sqltool/tool.go
```diff
@@ -374,6 +387,99 @@ var (
...
+	case 'B':
+		switch {
+		case ff.baseline != "" && flywayVersion(ff.baseline) == flywayVersion(path):
+			return fmt.Errorf("sql/sqltool: flyway: found duplicate baseline migrations %q and %q", ff.baseline, path)
+		case ff.baseline == "" || flywayVersion(ff.baseline) < flywayVersion(path):
+			ff.baseline = path
+		default:
+			return nil
+		}
+		// In case we set a new baseline, remove all versioned files with a version smaller than the new baseline.
```
a8m, line 411: +1

internal/integration/script_test.go
```diff
@@ -45,6 +46,7 @@ func TestMySQL_Script(t *testing.T) {
 				"execsql":     t.cmdExec,
 				"atlas":       t.cmdCLI,
 				"clearSchema": t.clearSchema,
+				"sleep":       sleep,
```
a8m, line 49: Let's not sleep in tests. Use a different name instead of the version.

sql/sqltool/hidden_windows.go
masseelch: The name itself is enough. `hidden_windows.go` will only be compiled if `GOOS == windows`.

a8m (approved):

PR #1078, opened 2022-08-21
cmd/atlas: use tags instead of releases for check for update

rotemtam:


cmd/atlas/internal/update/update.go
```diff
@@ -152,3 +160,14 @@ func saveStore(path string, l LatestRelease, t time.Time) error {
...
+
+type tag struct {
+	Name       string `json:"name"`
+	ZipballURL string `json:"zipball_url"`
+	TarballURL string `json:"tarball_url"`
+	Commit     struct {
+		SHA string `json:"sha"`
+		URL string `json:"url"`
+	} `json:"commit"`
+	NodeID string `json:"node_id"`
```
a8m, line 172: keep only the name.

giautm (approved):

PR #1084, opened 2022-08-22
sql/sqlspec/destructive: enable error by default

a8m:


sql/sqlcheck/sqlcheck.go
```diff
@@ -79,6 +79,16 @@ type (
 	ReportWriter interface {
 		WriteReport(Report)
 	}
+
+	// Options defines a generic configuration options for analyzers.
+	Options struct {
```
rotemtam, line 84: Does this belong in the PR?

rotemtam (approved):
Looks good see q

sql/sqlcheck/sqlcheck.go
a8m: Yes. See destructive.Analyzer.

PR #1100, opened 2022-08-26
cmd/atlas/internal/migrate: code-generate mapping between revision st…

masseelch:
…ructs


cmd/atlas/internal/migrate/ent/convert.tmpl
```diff
@@ -0,0 +1,36 @@
+{{/* gotype: entgo.io/ent/entc/gen.Graph */}}
+
+{{ define "convert" }}
```
a8m, line 3: Nice. Can you please move it under a `template` directory?

a8m (approved):

cmd/atlas/internal/migrate/ent/entc.go
```diff
@@ -26,6 +26,9 @@ func main() {
 			gen.FeatureExecQuery,
 			gen.FeatureSchemaConfig,
 		},
+		Templates: []*gen.Template{
+			gen.MustParse(gen.NewTemplate("").ParseFiles("convert.tmpl")),
+		},
 	})
```
a8m, line 32: Maybe pass it as an [`entc.Option`](https://github.com/ent/ent/blob/master/entc/entc.go#L159)? It will look cleaner:

```suggestion
	}, entc.TemplateFiles("convert.tmpl"))
```

PR #1104, opened 2022-08-29
sql/migrate: pending files are all files after the last fully applied one

masseelch:
Next PR will add the possibility to rollback more than one file in a transaction and therefore we need to re-apply more than one file in that case.


a8m (commented):
Can you please elaborate on the use case here? Also, let’s add test cases for it.

masseelch:
> Can you please elaborate on the use case here? Like what is the difference from the previous logic.
> 
> Also, let’s add test cases for it.

The previous logic would look at the revision, take the last one (regardless of state), find the version matching in the migration directory and pick up the execution there. 

Now, we want to add a transaction mode to group all pending migrations. In that case, we have to go back to the last fully applied revision and then start over from there (since everything would have been rolled back).

a8m (approved):

ronenlu (approved):

PR #1106, opened 2022-08-29
cmd/atlas/internal/cmdapi: introduce transaction mode for migration execution

masseelch:
This PR introduces the `--tx-mode` flag to specify the transaction strategy when applying pending migration files: 

- `all`: All pending migration files will be wrapped into one transaction. In case of an error, all files will be rolled back.
- `file`: Each pending migration file will be wrapped in its own transaction. In case of an error, only the currently executed file gets rolled back.
- `none`: No transaction. No rollbacks.

In addition, this PR changes the previous (buggy) behavior, where a rolled back migration was not reflected by the revision table. Now all rolled back revisions will get the `Applied` field set to zero. The previous PR already altered the logic to compute pending files to take this into account. 

Postgres testscript has all of the above tested. Handling of special cases like MySQL implicit commits has to be discussed.


a8m:
> Postgres testscript has all of the above tested. Handling of special cases like MySQL implicit commits has to be discussed.

Can you please add test cases for SQLite and MySQL as well? This feels to me like one the most critical part in versioned migration.

Also, what are the open questions about implicit commit and MySQL?

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -350,44 +354,128 @@ func CmdMigrateApplyRun(cmd *cobra.Command, args []string) error {
...
-				if err2 := tx.Rollback(); err2 != nil {
-					err = fmt.Errorf("%v: %w", err2, err)
+		if err := mux.mayRollback(ex.Execute(cmd.Context(), f), f); err != nil {
+			return err
+		}
+		if err := mux.mayCommit(); err != nil {
+			return err
+		}
+	}
+	if err := mux.commit(); err != nil {
```
a8m, line 376: My concern about using a separate connection for the revision-writer is that we lose the atomicity of the database. I think we can use SAVEPOINT to have both the information in the revisions table, but also maintain consistency. 

This of course can be added in future PRs because that’s the current behavior..

a8m (approved):

PR #1108, opened 2022-08-29
cmd/atlas/internal/migrate: revisions work on same connection as the migration execution

masseelch:
This enables the revision access work within the same transaction as the execution. This way we ensure atomicity on the database connection and correctness for the revision information to match the actual state of the database.

Logging information about migration attempts, rollbacks, errors and stuff like this will be separated from the actual revision information and be done somewhere else (e.g. nebula) and this will follow in another PR.

Since this PR was required for the `--tx-mode` flag to work properly the missing test for SQLite has been added as well.


ronenlu (approved):

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -354,40 +346,41 @@ func CmdMigrateApplyRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	var (
-		mux = tx{ctx: cmd.Context(), c: c, rrw: rrw, files: pending}
+		tx  = tx{c: c, rrw: rrw}
```
a8m, line 349: Try to avoid this shadowing.

cmd/atlas/internal/migrate/migrate.go
```diff
@@ -24,19 +25,21 @@ const DefaultRevisionSchema = "atlas_schema_revisions"
...
 	Option func(*EntRevisions) error
 )
 
-// NewEntRevisions creates a new EntRevisions with the given sqlclient.Client. It is important to call
-// EntRevisions.Init to initialize the underlying Ent client.
-func NewEntRevisions(ac *sqlclient.Client, opts ...Option) (*EntRevisions, error) {
+// NewEntRevisions creates a new EntRevisions with the given sqlclient.Client.
+func NewEntRevisions(ctx context.Context, ac *sqlclient.Client, opts ...Option) (*EntRevisions, error) {
+	// MySQL connections need to set the parseTime value.
+	if ac.Name == dialect.MySQL && strings.ToLower(ac.URL.Query().Get("parseTime")) != "true" {
```
a8m, line 40: Let’s handle it on Open. It’s weird to require this from users.

cmd/atlas/internal/migrate/migrate.go
```diff
@@ -119,78 +110,37 @@ func (r *EntRevisions) ReadRevisions(ctx context.Context) ([]*migrate.Revision,
...
-	// the database until the transaction wrapping the migration execution has been committed.
-	return r.ac.Name == dialect.SQLite
-}
-
-func (r *EntRevisions) openSchema(ctx context.Context) (*sqlclient.Client, error) {
-	u := r.ac.URL.URL
-	if r.ac.Name == dialect.MySQL && !u.Query().Has("parseTime") {
-		v := *u
-		q := v.Query()
-		q.Set("parseTime", "true")
```
a8m, line 189: See my comment above.

internal/integration/mysql_test.go
```diff
@@ -1278,7 +1278,7 @@ func (t *myTest) dsn(dbname string) string {
 	if t.mariadb() {
 		d = "mariadb"
 	}
-	return fmt.Sprintf("%s://root%s@localhost:%d/%s", d, pass, t.port, dbname)
```
a8m, line 1281: Remove and let Atlas handle it for the users.

a8m (approved):
💯

internal/integration/mysql_test.go
masseelch: This is in the test without a client open in between. Here it is valid to add the parseTime by hand.

PR #1117, opened 2022-09-01
cmd/atlas/internal/cmdapi: add migrate clean command

masseelch:
As a safety guard this command requires the `--local-url` flag as well as a confirmation from the user, unless the `--auto-approve` flag is given.

### DIFF
--- a/cmd/atlas/internal/cmdapi/schema.go
+++ b/cmd/atlas/internal/cmdapi/schema.go
@@ -100,10 +100,10 @@ migration.`,
 	SchemaClean = &cobra.Command{
 		Use:   "clean [flags]",
 		Short: "Removes all objects from the connected database.",
-		Long: `'atlas migrate clean' drops all objects in the connected database and leaves it in an empty state.
-As a safety feature, 'atlas migrate clean' will ask for confirmation before attempting to execute any SQL.`,
-		Example: `  atlas migrate clean -u mysql://user:pass@localhost:3306/dbname
-  atlas migrate clean --auto-approve --env local `,
+		Long: `'atlas schema clean' drops all objects in the connected database and leaves it in an empty state.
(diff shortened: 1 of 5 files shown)


cmd/atlas/internal/cmdapi/schema.go
```diff
@@ -279,6 +305,68 @@ func CmdApplyRun(cmd *cobra.Command, _ []string) error {
...
+	defer c.Close()
+	// Acquire a lock.
+	if l, ok := c.Driver.(schema.Locker); ok {
+		unlock, err := l.Lock(cmd.Context(), "atlas_migrate_execute", 0)
+		if err != nil {
+			return fmt.Errorf("acquiring database lock: %w", err)
+		}
+		// If unlocking fails notify the user about it.
+		defer cobra.CheckErr(unlock())
+	}
```
a8m, line 324: I don't feel like it is needed for local development. We don't use it for `schema apply` as well.

cmd/atlas/internal/cmdapi/schema.go
```diff
@@ -279,6 +305,68 @@ func CmdApplyRun(cmd *cobra.Command, _ []string) error {
...
+		if c.Comment != "" {
+			cmd.Println("--", strings.ToUpper(c.Comment[:1])+c.Comment[1:])
+		}
+		cmd.Println(c.Cmd)
+	}
+	if CleanFlags.AutoApprove || promptUser() {
+		if err := c.ApplyChanges(cmd.Context(), drop); err != nil {
+			return err
+		}
+	}
```
a8m, line 366: I don't know how much it's needed, but some stuff here can be moved out to shared functions on the same file and used by the 2 commands (apply and clean).

a8m (approved):
💯

PR #1120, opened 2022-09-02
cmd/atlas/internal: change default behavior of revision table location

masseelch:
This PR changes the default location of the revision information table. The old behavior was to store the table in a separate schema called `atlas_schema_revisions`. Now, if the connection is bound to a schema (e.g. when working with Ent), the table will be stored in that schema, unless the `--revision-schema` flag is given.

To not destroy pre-existing projects, we require the flag to be present, if the connection is schema bound, no table exists in that schema but we can find one in the old default schema.

- [x] add test for the above

### DIFF
--- a/cmd/atlas/internal/cmdapi/migrate.go
+++ b/cmd/atlas/internal/cmdapi/migrate.go
@@ -331,13 +331,13 @@ func CmdMigrateApplyRun(cmd *cobra.Command, args []string) error {
 			s, err := c.InspectSchema(cmd.Context(), defaultRevisionSchema, opts)
 			switch {
 			case schema.IsNotExistError(err):
-				goto proceed
 			case err != nil:
 				return err
-			}
-			if _, ok := s.Table(revision.Table); ok {
-				fmt.Fprintf(cmd.OutOrStderr(), `We couldn't find a revision table in the connected schema but 
-found one in the schema 'atlas_schema_revisions' and cannot.
+			default:
+				if _, ok := s.Table(revision.Table); ok {
+					fmt.Fprintf(cmd.OutOrStderr(),
+						`We couldn't find a revision table in the connected schema but found one in 
+the schema 'atlas_schema_revisions' and cannot determine the desired behavior.
 
 As a safety guard, we require you to specify whether to use the existing
 table in 'atlas_schema_revisions' or create a new one in the connected schema
@@ -345,12 +345,12 @@ by providing the '--revisions-schema' flag or deleting the 'atlas_schema_revisio
 schema if it is unused.
 
 `)
-				cmd.SilenceUsage = true
-				return errors.New("ambiguous revision table")
+					cmd.SilenceUsage = true
+					return errors.New("ambiguous revision table")
+				}
 			}
 		}
 	}
-proceed:
 	// Get the correct log format and destination. Currently, only os.Stdout is supported.
 	l, err := logFormat(cmd.OutOrStdout())
 	if err != nil {
--- a/cmd/atlas/internal/cmdapi/migrate.go
+++ b/cmd/atlas/internal/cmdapi/migrate.go
@@ -422,6 +422,7 @@ schema if it is unused.
 
 `)
 					cmd.SilenceUsage = true
+					cmd.SilenceErrors = true
 					return errors.New("ambiguous revision table")
 				}
 			}
@@ -444,7 +445,7 @@ func revisionSchemaName(c *sqlclient.Client) string {
 	case c.URL.Schema != "":
 		return c.URL.Schema
 	default:
-		return "atlas_schema_revisions"
+		return defaultRevisionSchema
 	}
 }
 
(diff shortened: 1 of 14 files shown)


cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -305,6 +308,43 @@ func CmdMigrateApplyRun(cmd *cobra.Command, args []string) error {
...
+As a safety guard, we require you to specify whether to use the existing
+table in 'atlas_schema_revisions' or create a new one in the connected schema
+by providing the '--revisions-schema' flag or deleting the 'atlas_schema_revisions'
+schema if it is unused.
+
+`)
+				cmd.SilenceUsage = true
+				return errors.New("ambiguous revision table")
+			}
+		}
```
a8m, line 345:
```suggestion
		if _, ok := s.Table(revision.Table); !ok {
			// Check for the old default schema. If it does not exist, we have no problem.
			s, err := c.InspectSchema(cmd.Context(), defaultRevisionSchema, opts)
			switch {
			case schema.IsNotExistError(err):
			case err != nil:
				return err
			default:
				if _, ok := s.Table(revision.Table); ok {
					fmt.Fprintf(cmd.OutOrStderr(), `We couldn't find a revision table in the connected schema but 
found one in the schema 'atlas_schema_revisions' and cannot.

As a safety guard, we require you to specify whether to use the existing
table in 'atlas_schema_revisions' or create a new one in the connected schema
by providing the '--revisions-schema' flag or deleting the 'atlas_schema_revisions'
schema if it is unused.

`)
					cmd.SilenceUsage = true
					return errors.New("ambiguous revision table")
			}
		}
```

a8m: > found one in the schema 'atlas_schema_revisions' and cannot.

Cannot what?

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -305,6 +308,43 @@ func CmdMigrateApplyRun(cmd *cobra.Command, args []string) error {
...
+	}
+proceed:
```
a8m, line 347: remove and see my suggestion above.

a8m (commented):
Please, add test cases and I'll re-review.

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -378,8 +381,71 @@ func CmdMigrateApplyRun(cmd *cobra.Command, args []string) error {
...
+					fmt.Fprintf(cmd.OutOrStderr(),
+						`We couldn't find a revision table in the connected schema but found one in 
+the schema 'atlas_schema_revisions' and cannot determine the desired behavior.
+
+As a safety guard, we require you to specify whether to use the existing
+table in 'atlas_schema_revisions' or create a new one in the connected schema
+by providing the '--revisions-schema' flag or deleting the 'atlas_schema_revisions'
+schema if it is unused.
+
+`)
```
a8m, line 423:
```suggestion
schema if it is unused.
`)
```

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -378,8 +381,71 @@ func CmdMigrateApplyRun(cmd *cobra.Command, args []string) error {
...
+					cmd.SilenceUsage = true
+					return errors.New("ambiguous revision table")
```
a8m, line 425: We can also silent error which is preferred here because we have an explicit message.

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -378,8 +381,71 @@ func CmdMigrateApplyRun(cmd *cobra.Command, args []string) error {
...
+const defaultRevisionSchema = "atlas_schema_revisions"
+
+func revisionSchemaName(c *sqlclient.Client) string {
+	switch {
+	case MigrateFlags.RevisionSchema != "":
+		return MigrateFlags.RevisionSchema
+	case c.URL.Schema != "":
+		return c.URL.Schema
+	default:
+		return "atlas_schema_revisions"
```
a8m, line 447:
```suggestion
		return defaultRevisionSchema
```

internal/integration/integration_test.go
```diff
@@ -175,12 +174,113 @@ func testHCLIntegration(t T, full string, empty string) {
 	require.Empty(t, t.realm().Schemas[0].Tables)
 }
 
-func testCLISchemaInspect(t T, h string, dsn string, eval schemahcl.Evaluator, args ...string) {
-	err := initCLI()
+func testCLIMigrateApplyBC(t T, dialect string) {
```
a8m, line 177: 😍

a8m (approved):
🚀

cmd/atlas/internal/cmdapi/migrate.go
masseelch: Following the exsting message structure here: https://github.com/ariga/atlas/blob/893f650a741c65c403e72f05702c50d3d214a260/cmd/atlas/internal/cmdapi/migrate.go#L762-L772

PR #1121, opened 2022-09-02
cmd/atlas/internal/cmdapi: do not allow args below zero in migrate apply

masseelch:


cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -284,6 +284,9 @@ func CmdMigrateApplyRun(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
+		if n < 1 {
+			return fmt.Errorf("cannot apply '%d' migration files", n)
```
a8m, line 288: Maybe, “n must be a positive number”?

a8m (approved):

cmd/atlas/internal/cmdapi/migrate.go
masseelch: But n is not declared to the user and if you do `atlas migrate apply 0` and get the error `cannot apply 0 files ` it is clear to the user what the issue is, don't you think?

a8m: 👍🏻

PR #1132, opened 2022-09-07
Typo fix for `Supported Databases` in Readme

fuziontech:
What's on the tin!


rotemtam (approved):
Thanks for the fix! ❤️

PR #1147, opened 2022-09-15
cmd/atlas/internal/cmdapi: add atlas migrate import command

masseelch:
This PR adds the `atlas migrate import` command to import existing migraiton directories from supported tools (e.g. golang-migrate) into the Atlas format.

Comments directly preceeding a SQL statement are preserved. Other comments are lost.

- [] Create Documentation
- [] Testcript


cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -53,6 +53,8 @@ const (
 	migrateApplyFromVersion     = "from"
 	migrateApplyBaselineVersion = "baseline"
 	migrateApplyTxMode          = "tx-mode"
+	migrateFlagImportFrom       = "from"
+	migrateFlagImportTo         = "to"
```
a8m, line 57: You can use `migrateFlagTo`. I'd say, move the `from` to be shared flag as well.

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -170,6 +176,36 @@ This command should be used whenever a manual change in the migration directory
...
+			if errors.Is(err, migrate.ErrChecksumNotFound) {
+				// Consider this case valid.
+				return nil
+			}
+			if err != nil {
+				printChecksumErr(cmd.OutOrStderr())
+				cmd.SilenceUsage = true
+				return err
+			}
+			return nil
```
a8m, line 205:
```suggestion
			if err := migrate.Validate(dir); err != nil && !errors.Is(err, migrate.ErrChecksumNotFound) {
				printChecksumErr(cmd.OutOrStderr())
				cmd.SilenceUsage = true
				return nil
			}
			return nil
```

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -251,16 +291,17 @@ func init() {
 	MigrateApplyCmd.Flags().SortFlags = false
 	cobra.CheckErr(MigrateApplyCmd.MarkFlagRequired(migrateFlagURL))
 	MigrateApplyCmd.MarkFlagsMutuallyExclusive(migrateApplyFromVersion, migrateApplyBaselineVersion)
-	cobra.CheckErr(MigrateApplyCmd.Flags().MarkHidden(migrateFlagDirFormat))
-	cobra.CheckErr(MigrateApplyCmd.Flags().MarkHidden(migrateFlagForce))
-	cobra.CheckErr(MigrateApplyCmd.Flags().MarkHidden(migrateFlagSchema))
 	// Diff flags.
```
a8m, line 294: Why? Also, can we remove the `--force` one?

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -623,6 +664,76 @@ func CmdMigrateHashRun(*cobra.Command, []string) error {
 	return migrate.WriteSumFile(dir, sum)
 }
 
+// CmdMigrateImportRun is the command executed when running the CLI with 'migrate import' args.
+func CmdMigrateImportRun(*cobra.Command, []string) error {
+	// Importing an Atlas project does not make any sense.
+	if MigrateFlags.DirFormat == formatAtlas {
+		return fmt.Errorf("cannot import an migration directory already in %q format", formatAtlas)
```
a8m, line 671:
```suggestion
		return fmt.Errorf("cannot import a migration directory already in %q format", formatAtlas)
```

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -623,6 +664,76 @@ func CmdMigrateHashRun(*cobra.Command, []string) error {
 	return migrate.WriteSumFile(dir, sum)
 }
 
+// CmdMigrateImportRun is the command executed when running the CLI with 'migrate import' args.
+func CmdMigrateImportRun(*cobra.Command, []string) error {
+	// Importing an Atlas project does not make any sense.
```
a8m, line 669:
```suggestion
```
Self-explained by the error.

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -623,6 +664,76 @@ func CmdMigrateHashRun(*cobra.Command, []string) error {
...
+		return err
+	case len(ff) != 0:
+		return errors.New("target migration directory must be empty")
+	}
+	ff, err = src.Files()
+	switch {
+	case err != nil:
+		return err
+	case len(ff) == 0:
+		return errors.New("nothing to import")
```
a8m, line 698: Should we error here? I'd expect no error.

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -623,6 +664,76 @@ func CmdMigrateHashRun(*cobra.Command, []string) error {
...
+		plan := &migrate.Plan{
+			Version: f.Version(),
+			Name:    f.Desc(),
+			Changes: make([]*migrate.Change, len(stmts)),
+		}
+		for i, s := range stmts {
+			cmd := strings.TrimSuffix(s.Text, ";")
+			for _, c := range s.Comments {
+				cmd += "\n" + c
+			}
```
a8m, line 716: Why the comments are attached to the end of the statements?

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -623,6 +664,76 @@ func CmdMigrateHashRun(*cobra.Command, []string) error {
...
+		if err != nil {
+			return err
+		}
+		for _, f := range files {
+			if err := trgt.WriteFile(f.Name(), f.Bytes()); err != nil {
+				return err
+			}
+		}
+	}
+	// Compute the sum file.
```
a8m, line 729: Redundant comment.
```suggestion
```

cmd/atlas/internal/cmdapi/migrate_test.go
```diff
@@ -36,6 +36,22 @@ func TestMigrate(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestMigrate_Import(t *testing.T) {
+	for _, tool := range []string{"dbmate", "flyway", "golang-migrate", "goose", "liquibase"} {
```
a8m, line 40: Check/compare the expected output?

doc/md/reference.md
```diff
@@ -69,7 +69,10 @@ atlas migrate
 #### Flags
 ```
       --dir string           select migration directory using URL format (default "file://migrations")
+      --dir-format string    set migration file format (default "atlas")
       --env string           set which env from the project file to use
+      --force                force a command to run on a broken migration directory state
+      --schema strings       set schema names
```
a8m, line 75: What this one is used for?

sql/migrate/dir.go
```diff
@@ -54,6 +54,8 @@ type (
 		Bytes() []byte
 		// Stmts returns the set of SQL statements this file holds.
 		Stmts() ([]string, error)
+		// StmtDecls returns the set of SQL statements this file holds alongside its preceding comments.
+		StmtDecls() ([]*Stmt, error)
```
a8m, line 58: 💯

cmd/atlas/internal/cmdapi/migrate.go
masseelch: Because this marks the flags hidden for ALL commands and not only the apply command. Cobra is a little weird here.

Keeping this means having the flags not show in any help command.

masseelch: > Why? Also, can we remove the `--force` one?

Yes, but not in this PR.

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -623,6 +660,85 @@ func CmdMigrateHashRun(*cobra.Command, []string) error {
 	return migrate.WriteSumFile(dir, sum)
 }
 
+// CmdMigrateImportRun is the command executed when running the CLI with 'migrate import' args.
+func CmdMigrateImportRun(cmd *cobra.Command, _ []string) error {
+	if MigrateFlags.DirFormat == formatAtlas {
+		return fmt.Errorf("cannot import a migration directory already in %q format", formatAtlas)
+	}
+	// Create migration directories.
```
a8m, line 668: Feels unnecessary and not correct. `dir(false)` won't create a dir.

doc/website/sidebars.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
                 {type: 'doc', id: 'versioned/lint', label: 'Migration Linting'},
                 {type: 'doc', id: 'versioned/new', label: 'Manual Migrations'},
                 {type: 'doc', id: 'versioned/apply', label: 'Migration Applying'},
+                {type: 'doc', id: 'versioned/import', label: 'Migration Directory Import'},
```
a8m, line 41:
```suggestion
                {type: 'doc', id: 'versioned/import', label: 'Migration Import'},
```

A sidebar cannot be that long.

a8m (approved):
Excellent work 💯

cmd/atlas/internal/cmdapi/migrate.go
masseelch: Comment is misleading. Create means instantiate in this case. Comment itself is a little useless.

PR #1150, opened 2022-09-16
sql/sqlcheck/datadepend: report column nullability changes that might fail

a8m:


cmd/atlas/internal/lint/lint.go
```diff
@@ -269,34 +269,23 @@ func (d *DevLoader) inspect(ctx context.Context) (*schema.Realm, error) {
...
+		return nil, errors.New("driver does not support locking")
 	}
-	s2 := make([]*migrate.Stmt, len(s1))
-	for i := range s1 {
-		p, err := pos(f, s1[i])
-		if err != nil {
-			return nil, err
-		}
-		s2[i] = &migrate.Stmt{Pos: p, Text: s1[i]}
+	name := "atlas_lint"
```
rotemtam, line 278: this prevents other linters from running, but not migrations from executing, right?

doc/md/lint/analyzers.md
```diff
@@ -164,4 +182,41 @@ zero (default) value. For example:
...
+
+2\. Backfill `NULL` values with a default value:
+
+```sql {1-2}
+-- backfill previous rows
+UPDATE `users` SET `a` = 1 WHERE `a` IS NULL;
+-- disable the enforcement of foreign-keys constraints
+PRAGMA foreign_keys = off;
+-- create "new_users" table
+CREATE TABLE `new_users` (`a` int NOT NULL);
```
rotemtam, line 213: the SQLite example is more complicated than other DBs here, maybe use another DB since it diverts focus

rotemtam (approved):
looks good see 2 comments

cmd/atlas/internal/lint/lint.go
a8m: right

doc/md/lint/analyzers.md
a8m: This is an SQLite-specific example using the "temporary table" because SQLite is treated a bit differently. See https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes.

That's why we have MF104 and LT101 checks.

PR #1153, opened 2022-09-19
sql/postgres: Add support for concurrent indexes

ivanvanderbyl:
In the process of resolving https://github.com/ent/ent/issues/1624 I've added support for creating indexes concurrently.


sql/postgres/inspect.go
```diff
@@ -772,6 +772,16 @@ type (
...
 
+	// IndexBuildParams describes additional parameters for index creation.
+	IndexBuildParams struct {
+		schema.Attr
+
+		// Concurrently describes the CONCURRENTLY clause to instruct Postgres
+		// to build the index concurrently without blocking the current table.
+		// https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY
+		Concurrently bool
+	}
```
a8m, line 783: I'm thinking, maybe we can name the attribute as `Concurrently` to allow using it in other schema changes like `schema.DropIndex` (https://www.postgresql.org/docs/current/sql-dropindex.html).

sql/postgres/sqlspec.go
```diff
@@ -334,6 +334,14 @@ func convertIndex(spec *sqlspec.Index, t *schema.Table) (*schema.Index, error) {
...
 		idx.Attrs = append(idx.Attrs, &IndexInclude{Columns: include})
 	}
+
+	if attr, ok := spec.Attr("concurrently"); ok {
+		concurrently, err := attr.Bool()
+		if err != nil {
+			return nil, err
+		}
+		idx.Attrs = append(idx.Attrs, &IndexBuildParams{Concurrently: concurrently})
+	}
```
a8m, line 344: I suggest not adding it to the HCL schema. The reason is that spec/schema defines the _"what"_ (the schema structure in the database) and this attribute is something that presents the _"how"_ (how to apply the change).

We have a rough design of how to hook into the apply/plan step in the declarative workflow and change the "how", but since you're using Ent, this should not be a blocker for you, right? 
```suggestion
```

a8m (commented):
Thanks for the contribution @ivanvanderbyl! 🙏 
Looks great, but I added 2 suggestions.

sql/postgres/inspect.go
ivanvanderbyl: Done. I added support for `DROP INDEX CONCURRENTLY` as well.

sql/postgres/inspect.go
```diff
@@ -772,6 +772,14 @@ type (
 		Columns []*schema.Column
 	}
 
+	// Concurrently describes the CONCURRENTLY clause to instruct Postgres
+	// to build or drop the index concurrently without blocking the current table.
+	// https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY
+	Concurrently struct {
+		schema.Attr
+		Concurrently bool
+	}
```
a8m, line 781:
```suggestion
	Concurrently struct {
		schema.Attr
	}
```
The flag there is not needed. If it exists in the attached attributes, it's enough. 
See `NoInherit` as an example.

sql/postgres/migrate.go
```diff
@@ -787,6 +787,11 @@ func (s *state) addIndexes(t *schema.Table, indexes ...*schema.Index) {
 			b.P("UNIQUE")
 		}
 		b.P("INDEX")
+		if c := (Concurrently{}); sqlx.Has(idx.Attrs, &c) {
+			if c.Concurrently {
+				b.P("CONCURRENTLY")
+			}
+		}
```
a8m, line 794:
```suggestion
		if c := (Concurrently{}); sqlx.Has(idx.Attrs, &c) {
			b.P("CONCURRENTLY")
		}
```
See my comment above.

sql/postgres/migrate.go
```diff
@@ -797,6 +802,11 @@ func (s *state) addIndexes(t *schema.Table, indexes ...*schema.Index) {
 			Comment: fmt.Sprintf("create index %q to table: %q", idx.Name, t.Name),
 			Reverse: func() string {
 				b := s.Build("DROP INDEX")
+				if c := (Concurrently{}); sqlx.Has(idx.Attrs, &c) {
+					if c.Concurrently {
+						b.P("CONCURRENTLY")
+					}
+				}
```
a8m, line 809:
```suggestion
				if c := (Concurrently{}); sqlx.Has(idx.Attrs, &c) {
					b.P("CONCURRENTLY")
				}
```
See the comment above.

sql/postgres/migrate_test.go
```diff
@@ -395,6 +395,17 @@ func TestPlanChanges(t *testing.T) {
...
 								},
 							},
+							&schema.AddIndex{
+								I: &schema.Index{
+									Name: "concurrently",
+									Parts: []*schema.IndexPart{
+										{C: users.Columns[0]},
+									},
+									Attrs: []schema.Attr{
+										&Concurrently{Concurrently: true},
```
a8m, line 405:
```suggestion
										&Concurrently{},
```

sql/postgres/sqlspec.go
```diff
@@ -334,6 +334,7 @@ func convertIndex(spec *sqlspec.Index, t *schema.Table) (*schema.Index, error) {
 		}
 		idx.Attrs = append(idx.Attrs, &IndexInclude{Columns: include})
 	}
+
```
a8m, line 337:
```suggestion
```

ivanvanderbyl:
Thanks for the suggestions, makes sense :)

a8m (approved):

a8m:
Thanks for the contribution, @ivanvanderbyl ❤️

PR #1159, opened 2022-09-22
cmd/atlas/internal/cmdapi: add atlas migrate set-revision

masseelch:
This PR introduces the `atlas migrate set-revision version` command.

It will manipulate the revision table such, that, with linear history in mind, we end up with all migration versions up-to and including the given version marked as applied.

**Example**

```
Migrations: 1, 2, 4, 5, 6
Revisions : 1, 2, 4
                  ^
             has an error
```

Running `atlas migrate set-revision 2` will result in:
```
Migrations: 1, 2, 4, 5, 6
Revisions : 1, 2
```
Running `atlas migrate set-revision 4` will result in:
```
Migrations: 1, 2, 4, 5, 6
Revisions : 1, 2, 4
```
Running `atlas migrate set-revision 7 --ignore-missing-migration ` will result in:
```
Migrations: 1, 2, 4, 5, 6
Revisions : 1, 2, 4, 5, 6, 7
```
Running `atlas migrate set-revision 3 --ignore-missing-migration && atlas migrate set-revision 6` will result in:
```
Migrations: 1, 2, 4, 5, 6
Revisions : 1, 2, 3, 4, 5, 6
```

The `--ignore-missing-migration` flag is a safe-guard against typos. If the user wants to mark a revision as applied, that has no matching migration, he has to provide that flag to the command.

### DIFF
--- a/cmd/atlas/internal/cmdapi/migrate.go
+++ b/cmd/atlas/internal/cmdapi/migrate.go
@@ -803,12 +803,20 @@ func CmdMigrateSetRevisionRun(cmd *cobra.Command, args []string) error {
 	if err := checkRevisionSchemaClarity(cmd, client); err != nil {
 		return err
 	}
+	// Ensure revision table exists.
+	rrw, err := entRevisions(cmd.Context(), client)
+	if err != nil {
+		return err
+	}
+	if err := rrw.Migrate(cmd.Context()); err != nil {
+		return err
+	}
 	// Wrap manipulation in a transaction.
 	tx, err := client.Tx(cmd.Context(), nil)
 	if err != nil {
 		return err
 	}
-	rrw, err := entRevisions(cmd.Context(), tx.Client)
+	rrw, err = entRevisions(cmd.Context(), tx.Client)
 	if err != nil {
 		return err
 	}
--- a/cmd/atlas/internal/cmdapi/migrate.go
+++ b/cmd/atlas/internal/cmdapi/migrate.go
@@ -35,24 +35,23 @@ import (
 )
 
 const (
-	migrateFlagURL                  = "url"
-	migrateFlagDevURL               = "dev-url"
-	migrateFlagDir                  = "dir"
-	migrateFlagDirFormat            = "dir-format"
-	migrateFlagLog                  = "log"
-	migrateFlagRevisionsSchema      = "revisions-schema"
-	migrateFlagDryRun               = "dry-run"
-	migrateFlagTo                   = "to"
-	migrateFlagFrom                 = "from"
-	migrateFlagSchema               = "schema"
-	migrateLintLatest               = "latest"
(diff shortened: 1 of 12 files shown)


cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -201,6 +205,19 @@ This command should be used whenever a manual change in the migration directory
 		Args:    cobra.MaximumNArgs(1),
 		RunE:    CmdMigrateNewRun,
 	}
+	// MigrateSetRevisionCmd represents the 'atlas migrate set-revision' command.
+	MigrateSetRevisionCmd = &cobra.Command{
+		Use:   "set-revision <version> [flags]",
```
a8m, line 210: If we set the revision, the argument should be called `revision`. But I would go with "version" instead of "revision".

Having said that, I don't like dashes in commands and I also think it's clear enough to read it this way:

```console
atlas migrate set <version>
```

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -201,6 +205,19 @@ This command should be used whenever a manual change in the migration directory
...
+		Short: "Override the revision history.",
```
a8m, line 211:
```suggestion
		Short: "Set the current version of a migration history table",
```

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -201,6 +205,19 @@ This command should be used whenever a manual change in the migration directory
...
+		Long: `'atlas migrate set-revision' edits the revision table to consider all migrations up to and including the given version
+to be applied. This command is usually used after manually making changes to the managed database.`,
+		Example: `  atlas migrate set-revision 3 --url mysql://user:pass@localhost:3306/
+  atlas migrate set-revision 4 --env local
+  atlas migrate set-revision 1.2.4 --url mysql://user:pass@localhost:3306/my_db --revision-schema my_revisions
+  atlas migrate set-revision foo --env dev --ignore-missing-migration`,
```
a8m, line 217: Feels too long and unnecessary

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -749,6 +776,151 @@ func CmdMigrateNewRun(_ *cobra.Command, args []string) error {
...
+		sum, err := dir.Checksum()
+		if err != nil {
+			return err
+		}
+		for _, f := range pending {
+			h, err := sum.SumByName(f.Name())
+			if err != nil {
+				return err
+			}
+			if err := rrw.WriteRevision(cmd.Context(), &migrate.Revision{
```
a8m, line 886: Why it's needed? Why not mark only the latest one?

a8m (approved):

PR #1164, opened 2022-09-26
sql/sqlite: check foreign keys once before committing a transaction

masseelch:


sql/sqlite/driver.go
```diff
@@ -181,19 +186,90 @@ func openTx(ctx context.Context, db *sql.DB, opts *sql.TxOptions) (*sqlclient.Tx
...
+	}
+	// Re-enable foreign key checks if they were enabled before.
+	if _, err2 := db.ExecContext(ctx, "PRAGMA foreign_keys = on"); err2 != nil {
+		err2 = fmt.Errorf("sql/sqlite: set 'foreign_keys = on': %w", err)
+		if err != nil {
+			return fmt.Errorf("%v: %w", err2, err)
+		}
+		return err2
+	}
+	return nil
```
a8m, line 230: You ignore the case where the above command succeeded, but you still have an error to return.

sql/sqlite/driver.go
```diff
@@ -181,19 +186,89 @@ func openTx(ctx context.Context, db *sql.DB, opts *sql.TxOptions) (*sqlclient.Tx
...
+	return vs, nil
+}
+
+// equalViolations compares the foreign key violations before starting a transaction with the ones afterwards.
+// It returns violations found in v2 that are not in v1.
+func newViolations(v1, v2 []violation) (vs []violation) {
+	for _, v := range v2 {
+		if !contains(v1, v) {
+			vs = append(vs, v)
+		}
```
a8m, line 258: `slices.Index` can be used here, but we don't use `slices` as a dependency in this module (only in `cmd/atlas`).

sql/sqlite/driver.go
```diff
@@ -181,19 +186,89 @@ func openTx(ctx context.Context, db *sql.DB, opts *sql.TxOptions) (*sqlclient.Tx
...
+	}
+	return vs
+}
+
+func contains(hs []violation, n violation) bool {
```
a8m, line 263:
```suggestion
func contains(vs []violation, n violation) bool {
```

masseelch: hs -> haystack, n -> needle

PR #1165, opened 2022-09-27
sql/sqlite: only check foreign-keys in before transaction commit, if …

masseelch:
…they were initially enabled

Currently checks are run every time instead of only if fks are enabled.

### DIFF
--- a/sql/sqlite/driver.go
+++ b/sql/sqlite/driver.go
@@ -170,7 +170,9 @@ type violation struct {
 	row, index int
 }
 
-// OpenTx opens a transaction. If foreign keys are enabled, it disables them, checks for constraint violations before doing so, disabling
+// OpenTx opens a transaction. If foreign keys are enabled, it disables them, checks for constraint violations,
+// opens the transaction and before committing ensures no new violations have been introduced by whatever Atlas was
+// doing.
 func OpenTx(ctx context.Context, db *sql.DB, opts *sql.TxOptions) (*sqlclient.Tx, error) {
 	var on sql.NullBool
 	if err := db.QueryRowContext(ctx, "PRAGMA foreign_keys").Scan(&on); err != nil {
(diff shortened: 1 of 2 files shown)


sql/sqlite/driver.go
```diff
@@ -170,48 +170,59 @@ type violation struct {
...
 	if err := db.QueryRowContext(ctx, "PRAGMA foreign_keys").Scan(&on); err != nil {
 		return nil, fmt.Errorf("sql/sqlite: querying 'foreign_keys' pragma: %w", err)
 	}
+	return OpenTx(ctx, on.Bool, db, opts)
+}
+
+// OpenTx opens a transaction. If foreign keys are enabled, it disables them, checks for constraint violations,
+// opens the transaction and before committing ensures no new violations have been introduced by whatever Atlas was
+// doing.
+func OpenTx(ctx context.Context, on bool, db *sql.DB, opts *sql.TxOptions) (*sqlclient.Tx, error) {
```
a8m, line 185: The sql.DB will block you from using it in Ent.

a8m: See: https://github.com/ent/ent/blob/master/dialect/sql/schema/atlas.go#L782

a8m (approved):

giautm (commented):
LGTM,

sql/sqlite/driver.go
```diff
@@ -179,44 +181,74 @@ func OpenTx(ctx context.Context, db *sql.DB, opts *sql.TxOptions) (*sqlclient.Tx
...
+
+// Tx wraps schema.ExecQuerier with the transaction methods.
+type Tx interface {
+	schema.ExecQuerier
+	Commit() error
+	Rollback() error
+}
+
+// CommitFn takes a transaction and ensures to toggle foreign keys back on after tx.Commit is called.
+func CommitFn(ctx context.Context, db schema.ExecQuerier, tx Tx, on bool) (func() error, error) {
```
a8m, line 212: Fn => Func.

PR #1167, opened 2022-10-01
cmd/atlas/internal/cmdapi: vercheck for updates

rotemtam:


cmd/atlas/internal/cmdapi/cmdapi.go
```diff
@@ -23,6 +23,32 @@ var (
 		Use:          "atlas",
 		Short:        "A database toolkit.",
 		SilenceUsage: true,
+		PostRunE: func(cmd *cobra.Command, _ []string) error {
```
masseelch, line 26: Considering we wanna add some information to the update, like "this fixes a critical bug", don't we wanna check before the CLI work instead if after it?

rotemtam: Which tools? My thought here is to not block the operation the user wants to perform.

cmd/atlas/internal/cmdapi/cmdapi.go
```diff
@@ -23,6 +23,32 @@ var (
...
+			// Users may skip update checking behavior.
+			if v := os.Getenv(AtlasNoUpdateNotifier); v != "" {
+				return nil
+			}
+			// Skip if the current binary version isn't set (dev mode).
+			if !semver.IsValid(version) {
+				return nil
+			}
+			path, err := homedir.Expand("~/.atlas/release.json")
```
yonidavidson, line 35: We should consider in the near future passing this configuration logic to `https://github.com/spf13/viper`  (reading files, comparing to env params, and setting "modes" for CLI)

cmd/atlas/internal/cmdapi/vercheck.go
```diff
@@ -0,0 +1,134 @@
+package cmdapi
```
a8m, line 1: Add license header.

cmd/atlas/internal/cmdapi/cmdapi.go
a8m, line 26: PostRun will block the CLI. Start the check in the background and on PostRun check if there's something to report. 

I'd say that it should be enabled only in development (not all operations, e.g. schema/migrate apply).

cmd/atlas/internal/cmdapi/vercheck.go
```diff
@@ -0,0 +1,134 @@
...
+		st, err := json.Marshal(s)
+		if err != nil {
+			return nil, err
+		}
+		if err := os.WriteFile(v.statePath, st, os.ModePerm); err != nil {
+			return nil, err
+		}
+	}
+	return &p, nil
+}
```
a8m, line 104: neat

cmd/atlas/internal/cmdapi/cmdapi.go
```diff
@@ -144,3 +116,8 @@ func parse(version string) (string, string) {
 	}
 	return version, u
 }
+
+// Version returns the current Atlas binary version.
+func Version() string {
+	return version
```
rotemtam, line 119: did this in order to avoid breaking the binary build which expects version at the specific location

cmd/atlas/internal/cmdapi/cmdapi.go
rotemtam: its popular but very "big"

a8m (approved):
Looks good. Consider moving this logic from the msin file

cmd/atlas/main.go
```diff
@@ -26,9 +33,72 @@ import (
...
+		if err != nil {
+			return
+		}
+		var b bytes.Buffer
+		if err := vercheck.Notify.Execute(&b, payload); err != nil {
+			return
+		}
+		updateMessage = b.String()
+	}()
+	return
```
a8m, line 103:
```suggestion
func noText() string {return ""}
// checkForUpdate checks for version updates and security advisories for Atlas.
func checkForUpdate() (func() string) {
	done := make(chan struct{})
	version := cmdapi.Version()
	// Users may skip update checking behavior.
	if v := os.Getenv(envNoUpdate); v != "" {
		return noText
	}
	// Skip if the current binary version isn't set (dev mode).
	if !semver.IsValid(version) {
		return noText
	}
	path, err := homedir.Expand(versionFile)
	if err != nil {
		return noText
	}
	var message string
	go func() {
		defer close(done)
		vc := vercheck.New(vercheckURL, path)
		payload, err := vc.Check(version)
		if err != nil {
			return
		}
		var b bytes.Buffer
		if err := vercheck.Notify.Execute(&b, payload); err != nil {
			return
		}
		message = b.String()
	}()
	return func() string {
		select {
		case u := <-done:
		case <-time.After(time.Millisecond * 500):
		}
		return message
	}
```

PR #1175, opened 2022-10-06
schemahcl: add support for the for_each meta argument and stdlib functions

a8m:
See: https://www.terraform.io/language/meta-arguments/for_each and https://www.terraform.io/language/functions.

### DIFF
--- a/schemahcl/hcl.go
+++ b/schemahcl/hcl.go
@@ -222,11 +222,33 @@ func (s *State) resource(ctx *hcl.EvalContext, file *hcl.File) (*Resource, error
 		if err != nil {
 			return nil, err
 		}
-		resource, err := s.toResource(ctx, blk, []string{blk.Type})
-		if err != nil {
-			return nil, err
+		switch {
+		case blk.Body != nil && blk.Body.Attributes[forEachAttr] != nil:
+			forEach, diags := blk.Body.Attributes[forEachAttr].Expr.Value(ctx)
+			if diags.HasErrors() {
+				return nil, diags
+			}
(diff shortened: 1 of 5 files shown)


schemahcl/hcl.go
```diff
@@ -222,11 +222,33 @@ func (s *State) resource(ctx *hcl.EvalContext, file *hcl.File) (*Resource, error
...
+		switch {
+		case blk.Body != nil && blk.Body.Attributes[forEachAttr] != nil:
+			forEach, diags := blk.Body.Attributes[forEachAttr].Expr.Value(ctx)
+			if diags.HasErrors() {
+				return nil, diags
+			}
+			if t := forEach.Type(); !t.IsSetType() && !t.IsObjectType() {
+				return nil, fmt.Errorf("schemahcl: for_each does not support %s type", t.FriendlyName())
+			}
+			delete(blk.Body.Attributes, forEachAttr)
```
masseelch, line 234: Why this?

a8m: This is a meta argument/attribute and should not be populated to the decoded resource.

schemahcl/stdlib.go
```diff
@@ -0,0 +1,235 @@
...
+				objT := cty.Object(argsT)
+				return cty.CapsuleVal(ctyNilType, &objT), nil
+			},
+		}),
+	}
+	return ctx
+}
+
+// standard functions exist in schemahcl language.
+func stdFuncs() map[string]function.Function {
```
rotemtam, line 99: 😍

rotemtam (approved):

PR #1176, opened 2022-10-06
doc/md/guides: Add article "Indexes with Included Columns in PostgreSQL"

priteshtm:


doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
+---
+id: included-columns
+title: Indexes with Included Columns in PostgreSQL
+slug: /guides/postgres/included-columns
+---
+
+With PostgreSQL, we can create indexes with `INCLUDE` clauses, which are types of indexes that specify a list of columns to be included in the index as non-key columns. If used correctly, indexes with Included columns improve performance and reduce total costs.
```
hilakashai, line 7:
```suggestion
With PostgreSQL, we can create indexes with `INCLUDE` clauses, which are types of indexes that specify a list of columns to be included in the index as non-key columns. If used correctly, indexes with included columns improve performance and reduce total costs.
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+
+```sql
+CREATE [UNIQUE] INDEX index_name
+ON table_name(key_column_list)
+INCLUDE(included_column_list);
+```
+
+### How do they work?
+
+A B-Tree index is made of two structures, 1) the B-tree and 2) The doubly linked list at the leaf node level of the B-tree. In an index with included columns, records of the columns mentioned in the `INCLUDE` clause are included in the ‘doubly linked list at the leaf node level of the B-tree’, corresponding to the heap(rows/tuples stored in a table)
```
hilakashai, line 19:
```suggestion
A B-Tree index is made of two structures:
1. The B-tree
2. The doubly linked list at the leaf node level of the B-tree
In an index with included columns, records of the columns mentioned in the `INCLUDE` clause are included in the doubly linked list at the leaf node level of the B-tree, corresponding to the heap (rows/tuples stored in a table).
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+
+:::info
+Each index is stored separately from the table's main data area, which is called the table's heap in PostgreSQL terminology. On the other hand, Leaf pages are the pages on the lowest level of the index tree. To know more about B-tree index structure in PostgreSQL, visit PostgreSQL documentation [here](https://www.postgresql.org/docs/current/btree-implementation.html#BTREE-STRUCTURE)
```
hilakashai, line 22:
```suggestion
Each index is stored separately from the table's main data area, which in PostgreSQL this is known as the table's _heap_.  Leaf pages are the pages on the lowest level of the index tree. To learn more about the B-tree index structure in PostgreSQL, visit the PostgreSQL documentation [here](https://www.postgresql.org/docs/current/btree-implementation.html#BTREE-STRUCTURE). 
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+-[ RECORD 1500 ]------------------------------------
+id         | 1500
+savings    | 69 873
+first_name | Imani
+last_name  | Noble
+email      | imaninoble@hotmail.net
+bank       | BG45LBAX41796917951361
+```
+
+Now, suppose we want to find the ID of a user by their email address. Let’s check the performance of the query with WHERE clause without any index, with the following command:
```
hilakashai, line 81:
```suggestion
Now, suppose we want to find the ID of a user by their email address. Let’s check the performance of the query with a WHERE clause without any index, with the following command:
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+   Filter: ((email)::text = 'd-abbott3425@google.edu'::text)
+   Rows Removed by Filter: 1499
+ Planning Time: 0.053 ms
+ Execution Time: 0.195 ms
+(5 rows)
+
+Time: 0.626 ms
+```
+
+Notice that the total cost is 38.75 units. If we want to use a unique index to accelerate the query, we can create it on the “email” column with the following command:
```
hilakashai, line 109:
```suggestion
Notice that the total cost is 38.75 units. If we want to use a unique index to accelerate the query, we can create it on the `email` column with the following command:
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+time=0.200..0.203 rows=1 loops=1)
+   Index Cond: ((email)::text = 'd-abbott3425@google.edu'::text)
+ Planning Time: 0.081 ms
+ Execution Time: 0.259 ms
+(4 rows)
+
+Time: 1.470 ms
+```
+
+Notice that total cost is now 8.29 units. The performance of the query has improved by creating a primary key index on “email” column, compared to 38.75 units without using any index. The engine still has to fetch the “first_name” and “last_name” columns from the table (also known as “Heap fetches”).
```
hilakashai, line 146:
```suggestion
Notice that the total cost is now 8.29 units. The performance of the query has improved by creating a primary key index on `email` column, compared to 38.75 units without using any index. The engine still has to fetch the `first_name` and `last_name` columns from the table (also known as "heap fetches").
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+```sql
+DROP INDEX emails_idx;
+```
+```console title="Output"
+DROP INDEX
+Time: 3.856 ms
+```
+
+Suppose we want to accelerate the same query using the INCLUDE clause.
+In the following command, we have created an index with an INCLUDE clause that precisely covers “first_name” and “last_name” columns which are part of the query for which we are trying to improve performance.
```
hilakashai, line 158:
```suggestion
In the following command, we will create an index with an `INCLUDE` clause that precisely covers `first_name` and `last_name` columns which are part of the query for which we are trying to improve performance.
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+
+A B-Tree index is made of two structures, 1) the B-tree and 2) The doubly linked list at the leaf node level of the B-tree. In an index with included columns, records of the columns mentioned in the `INCLUDE` clause are included in the ‘doubly linked list at the leaf node level of the B-tree’, corresponding to the heap(rows/tuples stored in a table)
+
+:::info
+Each index is stored separately from the table's main data area, which is called the table's heap in PostgreSQL terminology. On the other hand, Leaf pages are the pages on the lowest level of the index tree. To know more about B-tree index structure in PostgreSQL, visit PostgreSQL documentation [here](https://www.postgresql.org/docs/current/btree-implementation.html#BTREE-STRUCTURE) 
+:::
+
+### When do we need them?
+
+Let's demonstrate an example where an index with an INCLUDE clause may be useful, by contrasting it with a unique index without an INCLUDE clause.
```
hilakashai, line 27:
```suggestion
Let's demonstrate an example where an index with an `INCLUDE` clause may be useful, by contrasting it with a unique index without an `INCLUDE` clause.
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+   Index Cond: ((email)::text = 'd-abbott3425@google.edu'::text)
+ Planning Time: 0.081 ms
+ Execution Time: 0.259 ms
+(4 rows)
+
+Time: 1.470 ms
+```
+
+Notice that total cost is now 8.29 units. The performance of the query has improved by creating a primary key index on “email” column, compared to 38.75 units without using any index. The engine still has to fetch the “first_name” and “last_name” columns from the table (also known as “Heap fetches”).
+Let's drop the existing index to demonstrate next section in the article
```
hilakashai, line 147:
```suggestion
Let's drop the existing index to demonstrate the next section in the article:
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+---------------------------------------
+ Index Only Scan using emails_idx on bankdb  (cost=0.28..4.29 rows=1 width=37) (ac
+tual time=0.228..0.231 rows=1 loops=1)
+   Index Cond: (email = 'd-abbott3425@google.edu'::text)
+   Heap Fetches: 0
+ Planning Time: 0.233 ms
+ Execution Time: 0.283 ms
+(5 rows)
+```
+Notice that the total cost is now 4.29, which is significantly lower compared to 8.29 while using a unique index without the INCLUDE clause. Because, the query only scanned the index in order to get the data. As a result, “Heap fetches” is also zero, which means the query does not access any tables to retrieve the records.
```
hilakashai, line 195:
```suggestion
Notice that the total cost is now 4.29, which is significantly lower, compared to 8.29 which we got while using a unique index without the `INCLUDE` clause. We were able to reduce the total cost because the query only scanned the index in order to get the data. As a result, `heap fetches` is also zero, which means the query does not access any tables to retrieve the records. 
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+   Index Cond: (email = 'd-abbott3425@google.edu'::text)
+   Heap Fetches: 0
+ Planning Time: 0.233 ms
+ Execution Time: 0.283 ms
+(5 rows)
+```
+Notice that the total cost is now 4.29, which is significantly lower compared to 8.29 while using a unique index without the INCLUDE clause. Because, the query only scanned the index in order to get the data. As a result, “Heap fetches” is also zero, which means the query does not access any tables to retrieve the records. 
+
+:::info
+You might be wondering why we didn’t just use `CREATE INDEX ON bankdb(email,first_name,last_name)` instead of using the `INCLUDE` clause? One of the advantages by using the `INCLUDE` clause is having fewer levels of B-tree. All the `INCLUDE` columns are stored in the doubly linked list of the B-tree index.
```
hilakashai, line 198:
```suggestion
You might be wondering why we didn’t just use `CREATE INDEX ON bankdb(email,first_name,last_name)` instead of using the `INCLUDE` clause. One of the advantages of using the `INCLUDE` clause is having fewer levels in a  B-tree. All `INCLUDE` columns are stored in the doubly linked list of the B-tree index.
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+:::
+
+### Advantages of using Indexes with an INCLUDE clause:
+1. B-tree part of the B-tree index has fewer levels as they do not contain include columns
```
hilakashai, line 202: This is confusing :) 
Maybe just write:
1. The B-tree has fewer levels because they do not contain include columns
Do you think that has the same meaning?

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+2. Greatly improved performance
```
hilakashai, line 203:
```suggestion
2. Greatly improves performance
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,316 @@
...
+3. Has the ability to return the contents of non-key columns without having to visit the index's table
+
+### Limitation of using Indexes with included columns
+- Expressions are not supported as included columns since they cannot be used in index-only scans.
+
+## Managing indexes with included columns is easy with Atlas
+
+Managing indexes and database schemas in PostgreSQL can be confusing and error-prone. Atlas is an open-source project which allows us to manage our database using a simple and easy-to-understand declarative syntax (similar to Terraform). We will now learn how to manage  indexes with included columns using Atlas.
```
hilakashai, line 211:
```suggestion
Managing indexes and database schemas in PostgreSQL can be confusing and error-prone. Atlas is an open-source project which allows us to manage our database using a simple and easy-to-understand declarative syntax (similar to Terraform). We will now learn how to manage indexes with included columns using Atlas.
```

hilakashai (commented):
Great work!!
See comments :)

doc/md/guides/postgres/included-columns.md
priteshtm: Yes it has same meaning, replaced with your suggested sentence! Thanks! :)

rotemtam (approved):

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,319 @@
...
+
+```sql
+DROP INDEX emails_idx;
+```
+```console title="Output"
+DROP INDEX
+Time: 3.856 ms
+```
+
+Suppose we want to accelerate the same query using the INCLUDE clause.
```
a8m, line 160:
```suggestion
Suppose we want to accelerate the same query using the `INCLUDE` clause.
```

doc/md/guides/postgres/included-columns.md
```diff
@@ -0,0 +1,319 @@
...
+```hcl title="schema.hcl"
+table "bankdb" {
+  schema = schema.public
+  column "id" {
+    null = false
+    type = serial
+  }
+  column "savings" {
+    null    = true
+    type    = character_varying(100)
```
a8m, line 236: Please, use `varchar` (shorter form) instead: https://atlasgo.io/atlas-schema/sql-types#string-1

rotemtam (approved):

PR #1182, opened 2022-10-09
cmd/atlas/internal/cmdapi/migrate: set dir value from env in migrate command

unknowingknow:
According to this [guide](https://atlasgo.io/atlas-schema/projects#projects-with-versioned-migrations), `dir` can be set from a project file, but it seems not to work.


a8m (approved):
Thanks for the contribution, @unknowingknow 🙏🏻

PR #1194, opened 2022-10-12
sql/internal/specutil: add index-part attributes to spec

a8m:
Also, in case of columns-only (not expression) indexes. For example, index a column prefix


giautm (approved):
LGTM, <3

rotemtam (approved):
❤️

PR #1211, opened 2022-10-19
cmd/atlas: support multi-env logging

a8m:


rotemtam (approved):
💯

cmd/atlas/internal/migrate/report.go
```diff
@@ -476,6 +482,20 @@ func jsonFn(v any, args ...string) (string, error) {
 	return string(b), err
 }
 
+func jsonMerge(objects ...string) (string, error) {
+	var r map[string]any
+	for i := range objects {
+		if err := json.Unmarshal([]byte(objects[i]), &r); err != nil {
```
giautm, line 488: Interesting, I don't know that we can merge the JSON like this. 💯

giautm (commented):
💯

cmd/atlas/internal/migrate/report.go
a8m: ❤️

PR #1219, opened 2022-10-21
sql/spanner: Add initial parsing and diffing implementations

tmc:
WIP

This adds a minimal implementation of parsing and diffing to close the HCL->SQL loop.


tmc:
__

PR #1231, opened 2022-10-26
schemahcl: replace schemahcl.Value with cty.Value

a8m:


schemahcl/hcl_test.go
```diff
@@ -361,83 +361,6 @@ func TestMultiFile(t *testing.T) {
 	}, test.People[1])
 }
 
-func TestDynamicBlocks(t *testing.T) {
```
rotemtam, line 364: is this added elsewhere?

schemahcl/opts.go
```diff
@@ -188,28 +188,14 @@ func typeFuncSpecImpl(_ *function.Spec, typeSpec *TypeSpec) function.ImplFunc {
...
-				lst := &ListValue{}
-				for _, arg := range args {
-					v, err := extractValue(arg)
-					if err != nil {
-						return cty.NilVal, err
-					}
-					lst.V = append(lst.V, v)
-				}
-				t.Attrs = append(t.Attrs, &Attr{K: attr.Name, V: lst})
+				t.Attrs = append(t.Attrs, &Attr{K: attr.Name, V: cty.ListVal(args)})
```
rotemtam, line 191: nice

schemahcl/hcl_test.go
a8m: No. I decided to remove it as it's not needed anymore, and it just adds complexity to the current package.

rotemtam (approved):

PR #1240, opened 2022-10-27
sql/spanner: fix change ordering

tmc:
sql/spanner was missing a call to `sqlx.DetachCycles` which performs a topological sort of the changes to order them correctly.

### DIFF
--- a/internal/integration/spanner_test.go
+++ b/internal/integration/spanner_test.go
@@ -67,8 +67,8 @@ func TestSpanner_AddDropTable(t *testing.T) {
 		t.dropIndexes("idx_author_id", "idx_id_author_id_unique")
 
 		t.migrate(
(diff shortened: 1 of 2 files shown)


internal/integration/spanner_test.go
```diff
@@ -67,8 +67,8 @@ func TestSpanner_AddDropTable(t *testing.T) {
 		t.dropIndexes("idx_author_id", "idx_id_author_id_unique")
 
 		t.migrate(
-			&schema.AddTable{T: usersT},
 			&schema.AddTable{T: postsT},
+			&schema.AddTable{T: usersT},
```
tmc, line 71: I moved posts first here to show that the test still passes.

a8m (approved):

PR #1243, opened 2022-10-28
integration/spanner: Add HCL integration test

tmc:
This includes fixing up column parsing which this test surfaced as having a few issues.

### DIFF
--- a/sql/spanner/convert.go
+++ b/sql/spanner/convert.go
@@ -106,6 +106,7 @@ func ParseType(c string) (schema.Type, error) {
 
 // parseColumn attempts to populate a columnDesc.
 func parseColumn(s string) (*columnDesc, error) {
+	var err error
 	cd := &columnDesc{}
 	// split up type into, base type, size, and other modifiers.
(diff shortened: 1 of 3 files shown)


sql/spanner/convert.go
```diff
@@ -106,21 +106,22 @@ func ParseType(c string) (schema.Type, error) {
...
-		cd.size = size
+	if len(m) > 2 && m[2] != "" {
 		if m[2] == "MAX" {
 			cd.maxSize = true
+		} else {
+			cd.size, err = strconv.Atoi(m[2])
+			if err != nil {
+				return nil, fmt.Errorf("parseColumn: unable to convert %q to int: %w", m[2], err)
+			}
 		}
```
a8m, line 125:
```suggestion
		if m[2] == "MAX" {
			cd.maxSize = true
		} else if cd.size, err = strconv.Atoi(m[2]); err != nil {
			return nil, fmt.Errorf("parseColumn: unable to convert %q to int: %w", m[2], err)
		}
```

a8m (approved):

Issue #1250, opened 2022-10-31
How to spin up an ephemeral local docker container with specify the db??


chokoGtr:
It was mentioned that ephemeral docker is available for local execution of `atlas migrate lint`.
https://atlasgo.io/concepts/url

I want to spin up an ephemeral local docker container with specify the db.
Like `mysql://user:pass@localhost:3306/database`.
Is there any way to specify the db?

rotemtam:
Hi @chokoGtr,

Thanks for reaching out!

I'm not sure I understand your question, but if I understand correctly, you're asking how to specify the engine for the ephemeral docker container used as a dev db? 

There are some examples on the page you linked to:

<img width="976" alt="Screen Shot 2022-10-31 at 21 51 01" src="https://user-images.githubusercontent.com/1522681/199098065-12d6abb4-91ad-4347-a01d-7f308536cea7.png">

Does this answer your question?

a8m:
Issue was fixed with #1253.

PR #1253, opened 2022-11-01
cmd/atlas/internal/docker: allow setting database name in url

a8m:
Fixed https://github.com/ariga/atlas/issues/1250.


masseelch (approved):
:100:

PR #1266, opened 2022-11-06
sql/schema: Add String() method to ChangeKind

tmc:
This adds a `String() string` method to `ChangeKind` via [golang.org/x/tools/cmd/stringer](https://pkg.go.dev/golang.org/x/tools/cmd/stringer).


sql/schema/generate.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2021-present The Atlas Authors. All rights reserved.
+// This source code is licensed under the Apache 2.0 license found
+// in the LICENSE file in the root directory of this source tree.
+//go:generate stringer -type ChangeKind
```
a8m, line 4: Would you mind moving this command above the type definition? https://github.com/ariga/atlas/blob/master/sql/schema/migrate.go#L186

Thanks 🙏

a8m (approved):
💯 

Thanks a lot, @tmc

PR #1280, opened 2022-11-10
doc/md: add ariga cloud analyzer to doc

ronenlu:


crossworth:
I think this improves the performance as well.

> Saving a few bytes per tuple in a relation consisting of millions of rows may result in saving some significant storage space. Besides, if we can fit more tuple in a data page, performance can be improved due to lesser I/O activity.

From https://www.enterprisedb.com/postgres-tutorials/data-alignment-postgresql

I was curious about it and made a few tests with a table with 100_000_000 rows (using the example table provided on this PR).

**accounts_1**: Table without alignment
**accounts_2**: Table with alignment

Size difference:
```
accounts_1|accounts_2|
----------+----------+
7118 MB   |6366 MB   |
```
**~10.56% difference**

Query execution performance:
**accounts_1**: https://app.pgmustard.com/#/explore/e7241155-67e4-41c2-881b-def89b205d9b
**accounts_2**: https://app.pgmustard.com/#/explore/1522ab71-c91f-4f79-ba3f-bafe0df93927

It's not a _huge_ performance improvement, **2s and 1GB buffers read** (with default Postgres settings), I think using more columns would yield better results and of course billions of records should make the difference more visible. But hey, any performance improvement is a good improvement, specially in this case, where we don't really "loose" anything and have the storage space improvement as well.

Anyway, I think this analyzers are great and hope to use them in the future.


<details>
  <summary>Commands used</summary>
  Feel free to point problems or test with bigger tables.
  
  ```sql
CREATE TABLE accounts_1 (
    id bigint PRIMARY KEY,
    premium boolean,
    balance integer,
    age     smallint                  
);

CREATE TABLE accounts_2 (
    id bigint PRIMARY KEY,
    balance integer,
    age smallint,
    premium boolean             
);

INSERT INTO accounts_1 SELECT id, random() > 0.5, floor(random() * 10000 + 1000)::int, floor(random() * 50 + 20)::int FROM generate_series(1,1e8) AS gs(id);

INSERT INTO accounts_2 SELECT id, balance, age, premium from accounts_1;

vacuum full analyze;

SELECT pg_size_pretty( pg_total_relation_size('accounts_1')) as accounts_1, pg_size_pretty( pg_total_relation_size('accounts_2')) as accounts_2;

-- executed the commands below 4 times and took the 4 execution to be used on pgmustard
explain (analyze, buffers, verbose, settings, format json) select * from accounts_1 where age > 30 and premium is true

explain (analyze, buffers, verbose, settings, format json) select * from accounts_2 where age > 30 and premium is true
  ```
</details>

a8m (approved):
💯

PR #1282, opened 2022-11-12
fix build from source

Weilence:
Need to remove `go.work` which makes it impossible to compile cli from source


rotemtam:
Hey,
We compile from source multiple times a day. Can you share more details ?

Weilence:
image
go version go1.19.3 linux/amd64

in go.mod file
image

it seems like ignore `go.work` when build with `go install ariga.io/atlas/cmd/atlas@latest`.
but if `go install` in local, it's work well

a8m:
Hey @Weilence, and thanks for raising this issue. I forgot to upgrade the Atlas package version in `cmd/atlas` before tagging master. This PR and v0.8.2 should fix this issue: https://github.com/ariga/atlas/pull/1284.

In case you're curious why we're maintaining 2 different modules in this repository, the reason is to keep the Atlas package (`ariga.io.io/atlas`) leaner than the binary and avoid causing Ent users to download the binary dependencies when they are not needed (e.g. drivers).

image

Closing, but please let me know if you still have trouble with it 🙏

PR #1283, opened 2022-11-13
sql/schema: add RemoveAttr helper to schema dsl

a8m:
A small helper I need to use in Ent's codebase.


internal/integration/testdata/mysql/column-charset.txt
```diff
@@ -39,6 +39,11 @@ CREATE TABLE `users` (
   `name` varchar(255) CHARACTER SET hebrew NOT NULL
 )
 
+-- maria107/1.sql --
```
a8m, line 42: MariaDB's latest version added the collation to the table description.

rotemtam (approved):

PR #1285, opened 2022-11-14
cmd/atlas: unlock advisory lock after the command finished

a8m:
image


yonidavidson (approved):
I would validate migration run once, but LGTM

a8m:
Merged, but I’ll improve the flakiness in the **old** tests in next PRs.

PR #1292, opened 2022-11-17
feat/support-cue

sourcec0de:


tmc:
I think starting with either an issue or a [discussion](https://github.com/ariga/atlas/discussions) would be a good way to describe what you envision in terms of how atlas and cue may interact.

PR #1297, opened 2022-11-18
cmd/atlas: add --amend to 'migrate new' and 'migrate diff' 

a8m:
If I need to choose between `--interactive`/`-i` and `--amend`, I prefer the latter. Because the command is not really interactive and just gives you the option to amend/edit the files after they were created.

### DIFF
--- a/cmd/atlas/internal/cmdapi/migrate.go
+++ b/cmd/atlas/internal/cmdapi/migrate.go
@@ -14,6 +14,7 @@ import (
 	"io/fs"
 	"net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -1577,6 +1591,43 @@ func migrateEnvsRun[F any](run func(*cobra.Command, []string, F) error, cmd *cob
...
+
+// edit allows editing the file content using editor.
+func edit(name string, src []byte) ([]byte, error) {
+	path := filepath.Join(os.TempDir(), name)
+	if err := os.WriteFile(path, src, 0644); err != nil {
+		return nil, fmt.Errorf("write source content to temp file: %w", err)
+	}
+	defer os.Remove(path)
+	editor := "vi"
+	if e := os.Getenv("EDITOR"); e != "" {
```
a8m, line 1614: See: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_editor

rotemtam (approved):

PR #1300, opened 2022-11-21
doc/md/guides: Add doc about index operator classes in PostgreSQL

priteshtm:


a8m (approved):
💯

doc/md/guides/postgres/index-operator-classes.md
```diff
@@ -0,0 +1,289 @@
...
+
+Time: 243.573 ms
+```
+Notice that the index `internet_provider_idx` that we created was not used in order to execute this query. Instead, the `Parallel Seq Scan` operation was performed. As a result, the total execution time and cost are still too high.
+
+:::info
+In a parallel sequential scan, the table's blocks will be divided into ranges and shared among the cooperating processes. Each worker process will complete the scanning of its given range of blocks before requesting an additional range of blocks. To learn more about Parallel Plans in PostgreSQL, visit the official documentation [here](https://www.postgresql.org/docs/current/parallel-plans.html).
+:::
+
+Now, you might be wondering why the index that we created was not being used in the execution of this query. This is a point where having knowledge about usage of operator classes becomes important.
```
hilakashai, line 132:
```suggestion
Now, you might be wondering why the index that we created was not used in the execution of this query. This when having knowledge about the usage of operator classes becomes important. 
```

doc/md/guides/postgres/index-operator-classes.md
```diff
@@ -0,0 +1,289 @@
...
+---------------------------------------------------------------------------------------------------------------------------------------------
+ Index Scan using internet_provider_idx on internet_provider  (cost=0.42..8.45 rows=100 width=44) (actual time=0.025..0.131 rows=78 loops=1)
+   Index Cond: (((subscriber_name)::text ~>=~ 'Shirley C'::text) AND ((subscriber_name)::text ~<~ 'Shirley D'::text))
+   Filter: ((subscriber_name)::text ~~ 'Shirley C%'::text)
+ Planning Time: 0.103 ms
+ Execution Time: 0.153 ms
+(5 rows)
+```
+
+Amazing! This time, Index Scan was performed using `internet_provider_idx`. As a result the cost, planning time, and execution time for our query have reduced significantly, as we expected.
```
hilakashai, line 175:
```suggestion
Amazing! This time, Index Scan was performed using `internet_provider_idx`. As a result, the cost, planning time, and execution time for our query have reduced significantly, as we expected. 
```

doc/md/guides/postgres/index-operator-classes.md
```diff
@@ -0,0 +1,289 @@
...
+  }
+  primary_key {
+    columns = [column.id]
+  }
+}
+schema "public" {
+}
+```
+
+Now, lets add the following index definition to the file:
```
hilakashai, line 229:
```suggestion
Now, let's add the following index definition to the file:
```

PR #1302, opened 2022-11-22
sql/mysql: support native-UUID data type for MariaDB since version 10.7.0

michaelboke:
I added support for creating migrations from `ent` where you want to use the native UUID. This data type is supported since version 10.7.0 in maria db. It offers better performance than the varchar(36) solution and similar performance as the binary(16) solution but a readable uuid that does not need conversion.

Unfortunately, Atlas was reporting that the uuid type was an unsupported type when i tried to define it as a custom schemaType inside ent. This PR should fix that.

Inside ent i created a model like this

```go
type Customer struct {
	ent.Schema
}

func (Customer) Fields() []ent.Field {
	return []ent.Field{
		field.UUID("id", uuid.UUID{}).
			SchemaType(map[string]string{
				dialect.MySQL: "uuid",
			}).
			Default(uuid.New).
			StorageKey("id"),
                          //....
	}
}
```

and then run the creation of the migration (according to the docs)

```go
ctx := context.Background()
	dir, err := sqltool.NewGolangMigrateDir("ent/migrate/migrations")
	if err != nil {
		log.Fatalf("failed creating atlas migration directory: %v", err)
	}
	opts := []schema.MigrateOption{
		schema.WithDir(dir),                         // provide migration directory
		schema.WithMigrationMode(schema.ModeReplay), // provide migration mode
		schema.WithDialect(dialect.MySQL),           // Ent dialect to use
	}
	if len(os.Args) != 2 {
		log.Fatalln("migration name is required. Use: 'go run -mod=mod ent/migrate/main.go <name>'")
	}
	err = migrate.NamedDiff(ctx, "mysql://test:test@localhost:3306/test", os.Args[1], opts...)
	if err != nil {
		log.Fatalf("failed generating migration file: %v", err)
	}
```

I probably still need to add some tests for the changes but I would love to hear your opinions first on this change.

### DIFF
--- a/sql/mysql/convert.go
+++ b/sql/mysql/convert.go
@@ -244,7 +244,6 @@ func ParseType(raw string) (schema.Type, error) {
 		return &schema.UUIDType{
 			T: t,
 		}, nil
-
 	default:
 		return &schema.UnsupportedType{
 			T: t,
--- a/sql/mysql/convert.go
+++ b/sql/mysql/convert.go
@@ -99,6 +99,8 @@ func FormatType(t schema.Type) (string, error) {
 		if p := t.Precision; p != nil && *p > 0 {
 			f = fmt.Sprintf("%s(%d)", f, *p)
 		}
+	case *schema.UUIDType:
+		f = strings.ToLower(t.T)
 	case *schema.UnsupportedType:
 		// Do not accept unsupported types as we should cover all cases.
 		return "", fmt.Errorf("unsupported type %q", t.T)
@@ -238,6 +240,11 @@ func ParseType(raw string) (schema.Type, error) {
 		return &schema.SpatialType{
 			T: t,
 		}, nil
+	case TypeUUID:
+		return &schema.UUIDType{
+			T: t,
+		}, nil
+
 	default:
 		return &schema.UnsupportedType{
 			T: t,
--- a/sql/mysql/inspect.go
+++ b/sql/mysql/inspect.go
@@ -260,7 +260,6 @@ func (i *inspect) addColumn(s *schema.Schema, rows *sql.Rows) error {
 		}
 		c.Attrs = append(c.Attrs, a)
 	}
-
 	if attr.onUpdate != "" {
 		c.Attrs = append(c.Attrs, &OnUpdate{A: attr.onUpdate})
 	}
--- a/sql/mysql/inspect.go
+++ b/sql/mysql/inspect.go
@@ -260,6 +260,7 @@ func (i *inspect) addColumn(s *schema.Schema, rows *sql.Rows) error {
 		}
 		c.Attrs = append(c.Attrs, a)
 	}
+
 	if attr.onUpdate != "" {
 		c.Attrs = append(c.Attrs, &OnUpdate{A: attr.onUpdate})
 	}
--- a/sql/mysql/internal/mysqlversion/mysqlversion.go
+++ b/sql/mysql/internal/mysqlversion/mysqlversion.go
@@ -84,13 +84,6 @@ func (v V) SupportsIndexComment() bool {
 	return v.Maria() || v.GTE("5.5.3")
 }
 
-// SupportsUUIDDataType reports if the version supports uuid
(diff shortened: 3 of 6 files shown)


sql/mysql/inspect.go
```diff
@@ -260,6 +261,11 @@ func (i *inspect) addColumn(s *schema.Schema, rows *sql.Rows) error {
 		}
 		c.Attrs = append(c.Attrs, a)
 	}
+
+	if typ.String == "uuid" && !i.SupportsUUIDDataType() {
+		return errors.New("uuid datatype is not supported for the current database version")
+	}
+
```
a8m, line 268: No need for this check. If it comes from the database, we can assume it's valid. 

```suggestion
```

a8m (commented):
Thanks for the contribution, @michaelboke! That's appreciated.

Please, see the minor comment.

sql/mysql/inspect.go
michaelboke: Thanks for the review, that makes sense I removed the check. As i was assuming that from the ENT project the schema is created on the fly and not coming directly from an existing database.

sql/mysql/convert.go
```diff
@@ -238,6 +240,11 @@ func ParseType(raw string) (schema.Type, error) {
 		return &schema.SpatialType{
 			T: t,
 		}, nil
+	case TypeUUID:
+		return &schema.UUIDType{
+			T: t,
+		}, nil
+
```
a8m, line 247:
```suggestion
```

sql/mysql/inspect.go
```diff
@@ -260,6 +260,7 @@ func (i *inspect) addColumn(s *schema.Schema, rows *sql.Rows) error {
 		}
 		c.Attrs = append(c.Attrs, a)
 	}
+
```
a8m, line 263:
```suggestion
```

sql/mysql/internal/mysqlversion/mysqlversion.go
```diff
@@ -84,6 +84,13 @@ func (v V) SupportsIndexComment() bool {
 	return v.Maria() || v.GTE("5.5.3")
 }
 
+// SupportsUUIDDataType reports if the version supports uuid
+// data types, as from maria db 10.7.0 they are supported and offer
+// better performance over the varchar(32) and equal to binary(16) solutions.
+func (v V) SupportsUUIDDataType() bool {
+	return v.Maria() && v.GTE("10.7.0")
+}
+
```
a8m, line 93:
```suggestion
```

This method is not used anymore.

a8m (approved):

sql/schema/schema.go
```diff
@@ -247,6 +247,11 @@ type (
 		T string
 	}
 
+	// A UUIDType defines a UUID type.
+	UUIDType struct {
+		T string
+	}
```
a8m, line 253: That's great. I'll move the `postgres.UUIDType` to this new definition as well.

sql/mysql/internal/mysqlversion/mysqlversion.go
michaelboke: ah missed that one removed it :+1:

sql/mysql/inspect.go
michaelboke: done

sql/mysql/convert.go
michaelboke: done

sql/mysql/internal/mysqlversion/mysqlversion.go
a8m: Hey @michaelboke, I think you forgot to push the local changes because I still see this function.

giautm: @michaelboke: remind, to release this PR. :D

michaelboke: @a8m @giautm thanks for the notice!, i now really pushed the changes to GH 😄

a8m (approved):
🚀

PR #1310, opened 2022-11-25
sql/spanner: Use new spanner emulator and touch up tests

tmc:
As of 1.4.8 the Spanner emulator has DEFAULT support and this is a small change to include more use of defaults.


tmc:
I believe that #1311 needs to go in for `generate-cmp` to pass.

a8m (approved):

a8m:
Can I merge it, or do you want to fix it before?

tmc:
> Can I merge it, or do you want to fix it before?

Since the previous step can (and does) update go.work.sum this step fails: `git checkout -` without a `-f` -- since that check is explicitly using master as long as master and this (or any other) branch cause different content in go.work.sum that step will fail.

tmc:
(no-op push to re-trigger CI)

PR #1311, opened 2022-11-25
cmd/atlas: Update go.mod and go.sum files

tmc:
This is the result of `go work sync` which is necessary for the `generate-cmp` CI check to pass on this branch, as far as I can tell.


tmc:
Closing in favor of #1312

PR #1326, opened 2022-12-04
sql/spanner: Add and enhance FK AddDrop

tmc:
This adds sorting of changes as inspired by the TiDB driver.

Will rebase after #1310 lands.


a8m (approved):
💯

PR #1327, opened 2022-12-04
spanner: Add Sanity integration test

tmc:
This adds the TestSpanner_Sanity integration test which is meant to cover all reasonable column types to show that inspection is producing the expected output.

### DIFF
--- a/internal/integration/spanner_test.go
+++ b/internal/integration/spanner_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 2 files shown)


internal/integration/spanner_test.go
```diff
@@ -302,6 +305,155 @@ schema "default" {
...
+		if diff != "" {
+			t.Errorf("mismatch (-want +got): %s", diff)
+		}
+	})
+
+	// t.Run("ImplicitIndexes", func(t *testing.T) {
+	// 	stRun(t, func(t *spannerTest) {
+	// 		testImplicitIndexes(t, t.db)
+	// 	})
+	// })
```
a8m, line 454: Remove this in next PR.

PR #1329, opened 2022-12-05
cmd/atlas: basic json logging for schema inspect

a8m:
image

### DIFF
--- a/sql/mysql/sqlspec.go
+++ b/sql/mysql/sqlspec.go
@@ -27,12 +27,12 @@ type doc struct {
 
 // evalSpec evaluates an Atlas DDL document into v using the input.
 func evalSpec(p *hclparse.Parser, v any, input map[string]cty.Value) error {
-	var d doc
-	if err := hclState.Eval(p, &d, input); err != nil {
-		return err
-	}
 	switch v := v.(type) {
 	case *schema.Realm:
+		var d doc
+		if err := hclState.Eval(p, &d, input); err != nil {
+			return err
+		}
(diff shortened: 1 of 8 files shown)


sql/mysql/sqlspec.go
```diff
@@ -27,12 +27,12 @@ type doc struct {
...
-	var d doc
-	if err := hclState.Eval(p, &d, input); err != nil {
-		return err
-	}
 	switch v := v.(type) {
 	case *schema.Realm:
+		var d doc
+		if err := hclState.Eval(p, &d, input); err != nil {
+			return err
+		}
```
masseelch, line 35: Move this after the switch to avoid duplicate code (here, in case of schema.Schema and in default case

sql/postgres/sqlspec.go
```diff
@@ -42,12 +42,12 @@ func init() {
 
 // evalSpec evaluates an Atlas DDL document into v using the input.
 func evalSpec(p *hclparse.Parser, v any, input map[string]cty.Value) error {
-	var d doc
-	if err := hclState.Eval(p, &d, input); err != nil {
-		return err
-	}
 	switch v := v.(type) {
 	case *schema.Realm:
+		var d doc
```
masseelch, line 47: Dito

sql/mysql/sqlspec.go
a8m: The code is different in the cases.

masseelch (approved):

PR #1345, opened 2022-12-12
sql/mysql: implement reverse for DROP TABLE commands

a8m:


masseelch (approved):
This is a BC, isn't? Not sure if we can put this as a "bugfix" if suddenly a user, who expects drop tables to not be generated like before.

a8m:
This is BC, but wait for the next PR. It might be controversial ;)

PR #1346, opened 2022-12-12
sql: support multi-statement reverse commands

a8m:


sql/migrate/migrate.go
```diff
@@ -46,15 +46,29 @@ type (
 		// A Comment describes the change.
 		Comment string
 
-		// Reverse contains the "reversed statement" if
-		// command is reversible.
-		Reverse string
+		// Reverse contains the "reversed" statement(s) if
+		// the command is reversible.
+		Reverse any // string | []string
```
masseelch, line 51: Why not `[]string`? What code would we break by using `[]string`?

a8m: All codebase uses `string` except for one case. I preferred `any` for this reason, but also there's a chance we'll add support for `*Change` in the future (e.g., adding a comment per statement).

sql/mysql/migrate.go
```diff
@@ -40,7 +40,7 @@ func (p *planApply) PlanChanges(_ context.Context, name string, changes []schema
 		return nil, err
 	}
 	for _, c := range s.Changes {
-		if c.Reverse == "" {
+		if stmts, _ := c.ReverseStmts(); len(stmts) == 0 {
```
masseelch, line 43: Why ignore the error? It is not guaranteed we have `string | []string` here, is it?

a8m: I feel comfortable ignoring the error in this case because we build the changes in this file.

masseelch: By using a `DiffHook` a user can add a `Reverse` that is not a `string` or `[]string`.

rotemtam (approved):
GJ

PR #1354, opened 2022-12-16
doc: component for installation instructions

masseelch:
Creates a reusable markdown-component for the installation instructions.

The component itself has updated instructions about using the https://atlasgo.sh script to install / update Atlas.

image

Manual instructions simply list all available builds and then the user can decide what to download.

image


rotemtam (approved):

doc/md/getting-started/getting-started.mdx
```diff
@@ -5,74 +5,18 @@ sidebar_label: Quick Introduction
...
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
+import InstallationInstructions from '../components/_installation_instructions.mdx'
 import Discord from '../../website/src/assets/icons/discord-white.svg'
 
 Atlas CLI is an open-source tool designed to help software engineers, DBAs and DevOps
 practitioners to manage their database schemas.  Atlas users can use the [Atlas DDL](../atlas-schema/sql-resources)
 (data-definition language) to describe the desired database schema and use the command-line
 tool to plan and apply the migrations to their systems.
+Discord
```
a8m, line 15: What is this?

doc/md/getting-started/manual-installation.mdx
```diff
@@ -0,0 +1,65 @@
...
+slug: /manual-installation
+---
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+
+### Manual Installation
+
+If you don't trust the shell installation script or want to do a manual installation for some other reason, follow the
+below instructions.
```
a8m, line 12: The sentence feels odd to me. Also, I feel like we can add it under CLI reference and not in the getting started, but add a link to it in the `InstalltionInstruction` component. 
image

a8m (changes requested):
See comments.

doc/md/getting-started/getting-started.mdx
masseelch: Artifact

doc/md/getting-started/manual-installation.mdx
masseelch: I don't think so. I think everywhere we have the installation instructions it should be clear that the user can install without the script as well. 

I'd be okay with dropping the sentence though.

a8m: Why not add it to the `InstalltionInstruction` as a "manual installation" tab? Since this is quite a rare use case I don't think it should reside in the second document.

a8m (approved):
💯

PR #1359, opened 2022-12-22
doc/install: add docker to instructions component

a8m:


masseelch (approved):
Great! We could add some info about using docker as an alias for  `atlas`?

PR #1375, opened 2022-12-28
.github/workflows: create quick boot mysql images

ronenlu:


.github/workflows/cd-push-docker-mysql.yaml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build & Push ${{ matrix.dialect }} Docker Image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./ops/mysql/Dockerfile
+          push: true
+          tags: arigaio/${{ matrix.dialect }}-${{ github.sha }},arigaio/${{ matrix.dialect }}-latest
```
giautm, line 28: I dont think you have a valid image tag here. Double : in its name

giautm (changes requested):

.github/workflows/cd-push-docker-mysql.yaml
ronenlu: what do you suggest?
just arigaio/${{ matrix.dialect }} ?

ops/mysql/Dockerfile
```diff
@@ -0,0 +1,18 @@
+ARG DIALECT=mysql:8.0
```
a8m, line 1: Move this file under `.github`. There's no need to have it under the top level repository.

.github/workflows/cd-push-docker-mysql.yaml
giautm: We should add a prefix. `arigaio/eph-${{ matrix.dialect }}`. Or we need to separate the dialect and its version?

a8m: I prefer `arigaio/mysql:<tag>`.

ops/mysql/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+FROM $DIALECT as builder
+
+ARG MYSQL_DATABASE=test
+ARG MYSQL_USER=user
+ARG MYSQL_PASSWORD=pass
+ARG MYSQL_ROOT_PASSWORD=pass
+
+ENV MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD MYSQL_DATABASE=$MYSQL_DATABASE MYSQL_USER=$MYSQL_USER MYSQL_PASSWORD=$MYSQL_PASSWORD
+
+# That file does the DB initialization but also runs mysql daemon, by removing the last line it will only init
```
rotemtam, line 12:
```suggestion
# Remove the last line of the entry point script, leaving the initialization code but omitting actually starting the db.
```

a8m (approved):

PR #1379, opened 2023-01-02
cmd/atlas/docker: use mysql/mariadb images that boot up faster

a8m:
Also, using 'docker kill' sends SIGKILL and not SIGTERM results faster teardown

### DIFF
--- a/cmd/atlas/internal/docker/docker.go
+++ b/cmd/atlas/internal/docker/docker.go
@@ -15,6 +15,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


.github/workflows/cd-push-docker-mysql.yaml
```diff
@@ -7,7 +7,6 @@ on:
     branches:
       - master
   schedule:
-    # Runs at 00:00 on Sunday
```
masseelch, line 10: Why remove this comment? Not everyone has cron learned by heart :-D

giautm (approved):

cmd/atlas/internal/docker/docker.go
```diff
@@ -227,7 +236,7 @@ func (c *Config) Run(ctx context.Context) (*Container, error) {
 
 // Close stops and removes this container.
 func (c *Container) Close() error {
-	return exec.Command("docker", "stop", c.ID).Run() //nolint:gosec
+	return exec.Command("docker", "kill", c.ID).Run() //nolint:gosec
```
masseelch, line 239: This might create zombie processes. This is uncool.

masseelch (changes requested):

masseelch (approved):

PR #1393, opened 2023-01-08
sql/postgres: clear enum types after replaying migrations directory

a8m:
i.e., restore function used by Snapshoter

### DIFF
--- a/sql/postgres/migrate.go
+++ b/sql/postgres/migrate.go
@@ -200,7 +200,21 @@ func (s *state) addTable(ctx context.Context, add *schema.AddTable) error {
 
 // dropTable builds and executes the query for dropping a table from a schema.
 func (s *state) dropTable(ctx context.Context, drop *schema.DropTable) error {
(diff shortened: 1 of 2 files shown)


sql/postgres/migrate.go
```diff
@@ -782,8 +797,9 @@ func (s *state) enumExists(ctx context.Context, ns *schema.Schema, e *schema.Enu
 	return rows.Next(), rows.Err()
 }
 
-// mayDropEnum drops dangling enum types form the schema.
-func (s *state) mayDropEnum(alter *alterChange, ns *schema.Schema, e *schema.EnumType) error {
+// mayDropEnum drops dangling enum types form the schema. An optional
```
giautm, line 800:
```suggestion
// mayDropEnum drops dangling enum types from the schema. An optional
```

giautm (approved):
LGTM, one typo found. :D

sql/postgres/migrate.go
a8m: <3

PR #1396, opened 2023-01-10
doc: add examples for down migrations

a8m:


masseelch (commented):
I feel like this should be part of the versioned docs. There is no need for someone doing declarative style to have some info about rolling back versions here. WDYT?

masseelch (approved):

Issue #1417, opened 2023-01-23
[BUG]: Check constraint does not get excluded on exclude []


andreimc:
on the following schema:

```
schema "public" {
}
table "table" {
  schema = schema.public
  column "id" {
    null    = false
    type    = uuid
    default = sql("gen_random_uuid()")
  }
  column "kind" {
    null = false
    type = text
  }

  primary_key {
    columns = [column.id]
  }

  check "CHK_kind" {
    expr = "((kind = 'affirmation') OR (kind = 'reversal'))"
  }
}
```

with the following atlas.hcl config:

```
env "fromdbs" {
  src = "./schema_constraint.hcl"

  // note the url is for the dev db which should only be used when modifying migrations
  url = "postgresql://postgres:password@localhost:6432/fromdb?sslmode=disable"

  exclude = ["*.*.CHK_kind"]
}
```

```
atlas schema inspect --env fromdbs > without_cons.hcl
```

Constraint is still there.

I expect the constraint to be filtered also

andreimc:
I have created a merge request for this one.

a8m:
Thanks for reporting this and for the contribution, @andreimc. It's appreciated 🙏

PR #1418, opened 2023-01-23
bug #1417: Fix exclude constraint by name

andreimc:
check constraints couldn't be excluded.


sql/internal/sqlx/exclude.go
```diff
@@ -140,6 +140,15 @@ func excludeT(t *schema.Table, pattern string) (err error) {
...
 	})
+
+	t.Attrs, err = filter(t.Attrs, func(c schema.Attr) (bool, error) {
+		t := c.(*schema.Check)
+		match, err := filepath.Match(pattern, t.Name)
+		if !match || err != nil {
+			return false, err
+		}
+		return true, nil
+	})
```
a8m, line 151: Skip the attribute if it's not a `Check` type.
```suggestion
	t.Attrs, err = filter(t.Attrs, func(c schema.Attr) (bool, error) {
		c, ok := c.(*schema.Check)
		if !ok {
			return false, nil
		}
		match, err := filepath.Match(pattern, t.Name)
		if !match || err != nil {
			return false, err
		}
		return true, nil
	})
```

sql/internal/sqlx/exclude_test.go
```diff
@@ -94,6 +94,17 @@ func TestExcludeRealm_Tables(t *testing.T) {
 	require.Empty(t, r.Schemas[2].Tables)
 }
 
+func TestExcludeRealm_Checks(t *testing.T) {
+	r := schema.NewRealm(
+		schema.New("s1").AddTables(
+			schema.NewTable("t1").AddChecks(schema.NewCheck().SetName("c1")),
+		),
+	)
+
```
a8m, line 103:
```suggestion
```

cmd/atlas/go.work.sum
```diff
@@ -1,149 +1,657 @@
+bazil.org/fuse v0.0.0-20200407214033-5883e5a4b512 h1:SRsZGA7aFnCZETmov57jwPrWuTmaZK6+4R4v5FUe1/c=
```
a8m, line 1: I don't know how did you get this, but you can remove this from this PR.

a8m (commented):
🚀 

Thanks for the contribution, @andreimc. I added two little comments. Please, address them, and I'll merge this.

andreimc:
@a8m resolved :) I did a go build cmd main.go and it changed the sum, reverted that

andreimc:
@a8m would be good to have something like a Makefile to do the linting building etc 👍  was a bit of guess work to make this change. A contributing guide would help!

a8m:
Merged with https://github.com/ariga/atlas/pull/1426. Thanks for the contribution, @andreimc. It's appreciated 🙏

andreimc:
Thanks mate.

Issue #1414, opened 2023-01-20
Adding replication slots and publication forces me to run in tx-mode=none


advanderveer:
As part of a manual migration file that I want to apply with atlas we're creating a replication slot and publication like so:
```SQL
-- setup a replication slot for change capture
SELECT pg_create_logical_replication_slot('cl_cdc', 'pgoutput');
-- setup a replication publication for change capture
CREATE PUBLICATION cl_cdc FOR ALL TABLES
```
But when I run this with `atlas migrate apply` I get the following error:
```
Migrating to version 20230120064406 from 20230111102646 (1 migrations in total):

  -- migrating version 20230120064406
    -> SELECT pg_create_logical_replication_slot('cl_cdc', 'pgoutput');
    pq: cannot create logical replication slot in transaction that has performed writes

  -------------------------
  -- 314.7615ms
  -- 0 migrations ok (1 with errors)
  -- 0 sql statements ok (1 with errors)
Error: sql/migrate: sql/migrate: execute: executing statement "SELECT pg_create_logical_replication_slot('cl_cdc', 'pgoutput');" from version "20230120064406": pq: cannot create logical replication slot in transaction that has performed writes: sql/migrate: execute: write revision: pq: current transaction is aborted, commands ignored until end of transaction block
exit status 1
```

The only way to solve this right now is to switch to tx-mode=none but that nullifies the transactional DDL features we like from using postgres: https://wiki.postgresql.org/wiki/Transactional_DDL_in_PostgreSQL:_A_Competitive_Analysis

Is there an easy solution I'm missing? Or do we need a way to disable transactions per-file, using an annotation, like the "delimiter" annotation?

a8m:
Hey @advanderveer 👋
 
> Is there an easy solution I'm missing? Or do we need a way to disable transactions per-file, using an annotation, like the "delimiter" annotation?

That sounds like a valid use case. In many cases, users don't have control over the number of files that will be executed with `migrate apply`, and it feels cleaner to handle this problem when the file is written rather than during deployment.

advanderveer:
Hey @a8m thank you for the quick response, and what you say is true. Is this feature something that would be easy for me to contribute? Or do you wanna add it at a later stage?

advanderveer:
Very cool, thank you!

PR #1423, opened 2023-01-25
sql/migrate: support reading file directives

a8m:
Initial work for #1414.

### DIFF
--- a/sql/migrate/dir.go
+++ b/sql/migrate/dir.go
@@ -179,6 +179,37 @@ func (f LocalFile) Bytes() []byte {
 	return f.b
 }
 
+// Directive returns the (global) file directives that match the provided name.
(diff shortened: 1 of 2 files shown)


rotemtam (approved):

sql/migrate/dir.go
```diff
@@ -179,6 +179,37 @@ func (f LocalFile) Bytes() []byte {
 	return f.b
 }
 
+// Directive returns the (global) file directives that match the provided name.
+// File directives are located at the top of the file and should not be associated with any
+// statement. Hence, double new lines are used to separate file directives from its content.
```
masseelch, line 184: I am no fan of different meanings for \n and \n\n. is there another solution we can think of?

masseelch (approved):

PR #1430, opened 2023-01-26
cmd/atlas/internal: support file-level nolint directives

a8m:


doc/md/versioned/lint.mdx
```diff
@@ -104,6 +105,39 @@ statement.
 ALTER TABLE `t1` DROP COLUMN `c1`, ADD COLUMN `d1` varchar(255) NOT NULL;
 ```
 
+</TabItem>
+
+
+<TabItem value="file-level">
+
+The `--atlas:nolint` directive can be used at the file level to exclude the file from analysis, or to exclude specific
```
masseelch, line 113: It think "file level" is not clear enough at first side. What about sth like 


```suggestion
The `--atlas:nolint` directive can be used to exclude the whole file (given at top) from analysis, or to exclude specific
```
Something like the above?

masseelch (approved):

rotemtam (approved):

PR #1459, opened 2023-02-14
cmd/atlas: allow using ent/schemas as desired states

a8m:
e.g., 
```console
atlas migrate diff \
  --to "ent://path/to/schema?globalid=1" \
  --dev-url docker://mysql/8/test

atlas migrate diff \
  --to ent://path/to/schema \
  --dev-url docker://mysql/8/test
```


cmd/atlas/internal/cmdext/cmdext.go
```diff
@@ -192,3 +201,63 @@ func blockError(name string, b *hclsyntax.Block) func(string, ...any) error {
...
+func EntSchemaState(ctx context.Context, opts *ReadStateOptions) (migrate.StateReader, error) {
+	if opts.Dev == nil || len(opts.URLs) != 1 {
+		return nil, errors.New("schema url and dev database are required")
+	}
+	path := filepath.Join(opts.URLs[0].Host, opts.URLs[0].Path)
+	graph, err := entc.LoadGraph(path, &gen.Config{})
+	if err != nil {
+		return nil, fmt.Errorf("loading schema %s: %w", path, err)
+	}
+	m, err := entschema.NewMigrate(
```
masseelch, line 237: Why not use `NewMigrateURL()`? You can pass in the URL from the args directly.

a8m: I want to reuse the same database connection.

cmd/atlas/internal/cmdapi/migrate.go
```diff
@@ -380,10 +395,6 @@ func migrateDiffRun(cmd *cobra.Command, args []string, flags migrateDiffFlags) e
 	}
 	// Plan the changes and create a new migration file.
 	pl := migrate.NewPlanner(dev.Driver, dir, opts...)
-	var name string
-	if len(args) > 0 {
-		name = args[0]
-	}
```
masseelch, line 386: This variable is used laters right?

cmd/atlas/internal/cmdext/cmdext.go
```diff
@@ -192,3 +201,134 @@ func blockError(name string, b *hclsyntax.Block) func(string, ...any) error {
...
+// EntLoader is a StateLoader for loading ent.Schema's as StateReader's.
+type EntLoader struct{}
+
+// LoadState returns a migrate.StateReader that reads the schema from an ent.Schema.
+func (l EntLoader) LoadState(ctx context.Context, opts *LoadStateOptions) (migrate.StateReader, error) {
+	switch {
+	case opts.Dev == nil, len(opts.URLs) != 1:
+		return nil, errors.New("schema url and dev database are required")
+	case opts.URLs[0].Query().Has("globalid"):
+		return nil, errors.New("globalid is not supported by this command. Use 'migrate diff' instead")
```
masseelch, line 262: This confuses me. Not sure if miss some context but this shouldn't be here, right?

cmd/atlas/internal/cmdapi/migrate.go
a8m: Right. I moved it above

cmd/atlas/internal/cmdext/cmdext.go
a8m: Ent schema with `globalid` is supported only by `migrate diff` and should be rejected otherwise, because Atlas doesn’t support appending values into the `ent_types` table.

masseelch: Ah, now I got it.

PR #1464, opened 2023-02-19
cmd/lint: fixed atlas.sum report on empty dir

giautm:
Signed-off-by: Giau. Tran Minh <hello@giautm.dev>

### DIFF
--- a/cmd/atlas/internal/lint/run.go
+++ b/cmd/atlas/internal/lint/run.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"strings"
 	"text/template"
 
@@ -87,8 +88,7 @@ func (r *Runner) summary(ctx context.Context) error {
 		return r.sum.StepError(stepIntegrityCheck, fmt.Sprintf("File %s is invalid", migrate.HashFileName), err)
 	default:
 		// If the hash file exists, it is valid.
-		if f, err := r.Dir.Open(migrate.HashFileName); err == nil {
-			f.Close()
+		if _, err := fs.Stat(r.Dir, migrate.HashFileName); err == nil {
 			r.sum.StepResult(stepIntegrityCheck, fmt.Sprintf("File %s is valid", migrate.HashFileName), nil)
 		}
 	}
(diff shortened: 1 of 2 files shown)


cmd/atlas/internal/lint/run.go
```diff
@@ -86,7 +86,11 @@ func (r *Runner) summary(ctx context.Context) error {
 		r.sum.Files = append(r.sum.Files, &FileReport{Name: migrate.HashFileName, Error: err.Error()})
 		return r.sum.StepError(stepIntegrityCheck, fmt.Sprintf("File %s is invalid", migrate.HashFileName), err)
 	default:
-		r.sum.StepResult(stepIntegrityCheck, fmt.Sprintf("File %s is valid", migrate.HashFileName), nil)
+		// If the hash file exists, it is valid.
+		if f, err := r.Dir.Open(migrate.HashFileName); err == nil {
+			f.Close()
+			r.sum.StepResult(stepIntegrityCheck, fmt.Sprintf("File %s is valid", migrate.HashFileName), nil)
+		}
```
a8m, line 93: Why this code is needed if we handle ErrChecksumNotFound?

giautm: Because the dir is empty, the original logic of Validate was allow an empty dir is valid.

This test case run on an empty dir. So it report atlas.sum is valid, but the file is not existing.

cmd/atlas/internal/lint/run.go
```diff
@@ -86,7 +86,11 @@ func (r *Runner) summary(ctx context.Context) error {
 		r.sum.Files = append(r.sum.Files, &FileReport{Name: migrate.HashFileName, Error: err.Error()})
 		return r.sum.StepError(stepIntegrityCheck, fmt.Sprintf("File %s is invalid", migrate.HashFileName), err)
 	default:
-		r.sum.StepResult(stepIntegrityCheck, fmt.Sprintf("File %s is valid", migrate.HashFileName), nil)
+		// If the hash file exists, it is valid.
+		if f, err := r.Dir.Open(migrate.HashFileName); err == nil {
+			f.Close()
+			r.sum.StepResult(stepIntegrityCheck, fmt.Sprintf("File %s is valid", migrate.HashFileName), nil)
```
a8m, line 92:
```suggestion
		if _, err := fs.Stat(r.Dir, migrate.HashFileName); err == nil {
			r.sum.StepResult(stepIntegrityCheck, fmt.Sprintf("File %s is valid", migrate.HashFileName), nil)
```

a8m (approved):
Fix the comment and land it 🙏🏻

a8m (approved):

PR #1467, opened 2023-02-20
cmd/atlas: experiment function for running lint

a8m:


cmd/atlas/atlasapi/atlasapi.go
```diff
@@ -0,0 +1,57 @@
...
+		Dev:            dev,
+		Dir:            dir,
+		Analyzers:      az,
+		ChangeDetector: lint.LatestChanges(dir, latest),
+		ReportWriter: reporterFunc(func(r *SummaryReport) error {
+			report = r
+			return nil
+		}),
+	}
+	if err = r.Run(context.Background()); err != nil {
```
giautm, line 42:
```suggestion
	if err = r.Run(ctx); err != nil {
```

masseelch (approved):

rotemtam (approved):

PR #1471, opened 2023-02-22
doc/md/declarative: Add Quick-start docs about declarative migrations with single schema

priteshtm:


a8m:
See https://github.com/ariga/atlas/pull/1471

PR #1478, opened 2023-02-24
schemahcl: limit type variables scope to specific location

a8m:
Limiting the variable's scope to specific attributes provides type-safety to the schema. i.e., Eval fails on "parsing" and not on the decoding stage, which improves the error reporting. 

From:
```
Error: schemahcl: failed reading spec as *postgres.doc: set field "type": converting cty.Value to *schemahcl.Type: incorrect type type
```

To:
```
Error: schema1.hcl:6,12-18: Unknown column.type; There is no type named "string".
```


rotemtam (approved):
Well done

PR #1490, opened 2023-02-28
feat/atlas react website

dimagreat:
Update main page with new deisgn

### DIFF
--- a/doc/website/package-lock.json
+++ b/doc/website/package-lock.json
@@ -8,6 +8,7 @@
       "name": "website",
       "version": "0.0.0",
       "dependencies": {
+        "@ariga/atlas-website": "0.0.6",
         "@docusaurus/core": "^2.0.0-rc.1",
         "@docusaurus/plugin-client-redirects": "^2.0.0-rc.1",
         "@docusaurus/preset-classic": "^2.0.0-rc.1",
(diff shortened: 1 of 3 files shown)


a8m (approved):

sbellone:
Hello, I'm from the Algolia crawler team (that powers DocSearch), and we've noticed recently bursts in bandwidth coming from your website. The timing seems to indicate that it can be related to this change.
Basically most of your pages now weight 1.55MB:
```
$ curl https://atlasgo.io/getting-started/guides/testing/guides -I
HTTP/2 200 
content-type: text/html
content-length: 1548783
```

The problem is that even the 404 have this behaviour, they are actually responding 200 and have a 1.55MB content-length, to only later display a 404 message. Also they expose relative links, leading to infinite paths such as: https://atlasgo.io/cloud/getting-started/guides/cloud/getting-started, https://atlasgo.io/cli/declarative/apply/guides/mysql/getting-started/getting-started, ...

a8m:
Thanks for reporting this to us, @sbellone. I appreciate it 🙏 
We'll start by fixing these relative links immediately and investigate the 404 behavior. I'll keep you updated on our progress.

giautm:
Hello @sbellone, we pushed a fix to production today. The not found page no longer returns 1.55MB in size anymore. And the status returns correctly to 404.

Thanks for your report here. We appreciate it.

```shell
$ curl https://atlasgo.io/getting-started/guides/testing/guides -I
HTTP/2 404
content-type: text/html
content-length: 12061
```

sbellone:
Thanks for notifying! I can confirm that your DocSearch crawler is back to crawling a small amount of pages which all have a reasonable payload :slightly_smiling_face:

PR #1492, opened 2023-02-28
sql/migrate: archive and unarchive dir

rotemtam:


cmd/atlas/internal/cmdext/cmdext.go
```diff
@@ -459,3 +461,91 @@ func dirFormatter(dir migrate.Dir) migrate.Formatter {
 		return migrate.DefaultFormatter
 	}
 }
+
+// ArchiveDir returns a tar archive of the given directory.
+func ArchiveDir(dir migrate.Dir) ([]byte, error) {
```
a8m, line 466: I think the API is small enough (and elegant to live in sql/migrate).

cmd/atlas/internal/cmdext/cmdext.go
```diff
@@ -459,3 +461,91 @@ func dirFormatter(dir migrate.Dir) migrate.Formatter {
...
+		if err := append2Tar(tw, f.Name(), f.Bytes()); err != nil {
+			return nil, err
+		}
+	}
+	return buf.Bytes(), nil
+}
+
+// UnarchiveDir extracts the tar archive into the given directory. If the archive contains
+// a file named "atlas.sum", it will be used to verify the checksum of the directory.
+func UnarchiveDir(arc []byte, dir migrate.Dir) error {
```
a8m, line 496: I would say, change the API to `UnarchiveDir([]byte) (migrate.Dir, err)` and always return MemDir, because in case of an error (below), the LocalDir will end up in a stale state.

cmd/atlas/internal/cmdext/cmdext.go
```diff
@@ -459,3 +461,91 @@ func dirFormatter(dir migrate.Dir) migrate.Formatter {
...
+			data, err := io.ReadAll(tr)
+			if err != nil {
+				return err
+			}
+			if err := dir.WriteFile(h.Name, data); err != nil {
+				return err
+			}
+		}
+	}
+	if sumf != nil {
```
a8m, line 523: migrate.Validate

a8m (commented):
🚀

sql/migrate/dir.go
```diff
@@ -555,3 +556,79 @@ func (m *memFile) Mode() fs.FileMode          { return 0 }
 func (m *memFile) ModTime() time.Time         { return time.Time{} }
 func (m *memFile) IsDir() bool                { return false }
 func (m *memFile) Sys() interface{}           { return nil }
+
+// ArchiveDir returns a tar archive of the given directory.
+func ArchiveDir(dir Dir) ([]byte, error) {
+	var buf bytes.Buffer
```
masseelch, line 562: I think we should validate the dir here once. Atlas promises to not work on broken files.

sql/migrate/dir.go
```diff
@@ -555,3 +556,79 @@ func (m *memFile) Mode() fs.FileMode          { return 0 }
...
+	}
+	return buf.Bytes(), nil
+}
+
+// UnarchiveDir extracts the tar archive into the given directory. If the archive contains
+// a file named "atlas.sum", it will be used to verify the checksum of the directory.
+func UnarchiveDir(arc []byte) (Dir, error) {
+	tr := tar.NewReader(bytes.NewReader(arc))
+	var sum bool
+	md := &MemDir{}
```
giautm, line 594:
```suggestion
	var (
		md = &MemDir{}
		sum bool
		tr  = tar.NewReader(bytes.NewReader(arc))
	)
```

sql/migrate/dir.go
rotemtam: but we generate the sumfile ourselves in line 566

giautm: You can try to call Open(HashFileName), and check exist error.

giautm: @masseelch: I just thought, maybe we need to accept an invalid dir and then later report it using atlas lint?

giautm (approved):

sql/migrate/dir.go
a8m: I don’t think it’s the responsibility of this function to validate the Dir. However, it does make sense to call migrate.Validate in Unarchive as we build and return a Dir, and I expect to get a valid one.

sql/migrate/dir.go
```diff
@@ -555,3 +556,73 @@ func (m *memFile) Mode() fs.FileMode          { return 0 }
...
 func (m *memFile) Sys() interface{}           { return nil }
+
+// ArchiveDir returns a tar archive of the given directory.
+func ArchiveDir(dir Dir) ([]byte, error) {
+	var buf bytes.Buffer
+	tw := tar.NewWriter(&buf)
+	defer tw.Close()
+
+	sumF, err := dir.Open(HashFileName)
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
```
a8m, line 567:
```suggestion
	if err != nil {
```
We can keep it simple and fail here.

sql/migrate/dir.go
```diff
@@ -555,3 +556,73 @@ func (m *memFile) Mode() fs.FileMode          { return 0 }
...
+		}
+		data, err := io.ReadAll(tr)
+		if err != nil {
+			return nil, err
+		}
+		if err := md.WriteFile(h.Name, data); err != nil {
+			return nil, err
+		}
+	}
+	return md, nil
```
a8m, line 613: Add migrate.Validate here.

sql/migrate/dir.go
rotemtam: I want to skip errnotexist so we can archive directories without sumfiles, shouldn't we support that ?

a8m: kk

sql/migrate/dir.go
masseelch: I think it is fine to not validate. Esepcially if the dir archived was not valid. It should return exactly what was compressed in the first place.

PR #1494, opened 2023-03-01
sql: support planning with indentation 

a8m:


sql/internal/sqlx/sqlx.go
```diff
@@ -216,6 +216,8 @@ type Builder struct {
 	bytes.Buffer
 	QuoteChar byte    // quoting identifiers
 	Schema    *string // schema qualifier
+	Indent    string  // indentation prefix
```
masseelch, line 219: Usually a prefix is something else than the indention. E.g. I can have a prefix of '+' and indent '  ': 

```
+  create table t (
+    t int
+  )
```

Not sure if we want that as well to mimic `json.MarshalIndent` behavior.

a8m: The comment is wrong. I'll change it

masseelch (approved):

PR #1516, opened 2023-03-08
sql/sqlite/sqlitecheck: add backwards-incompatible checks

a8m:


cmd/atlas/internal/sqlparse/parseutil/parseutil.go
```diff
@@ -64,8 +64,8 @@ func RenameIndex(modify *schema.ModifyTable, r *Rename) {
 
 // RenameTable patches DROP/ADD table commands to RENAME.
 func RenameTable(changes schema.Changes, r *Rename) schema.Changes {
-	i := changes.IndexDropTable(r.From)
-	j := changes.IndexAddTable(r.To)
+	i := changes.LastIndexDropTable(r.From)
+	j := changes.LastIndexAddTable(r.To)
```
masseelch, line 68:
```suggestion
	var (
		i = changes.LastIndexDropTable(r.From)
		j = changes.LastIndexAddTable(r.To)
	)
```

cmd/atlas/internal/sqlparse/sqliteparse/sqliteparse.go
```diff
@@ -128,6 +129,18 @@ func (s *Stmt) TableUpdate(t *schema.Table) (*Update_stmtContext, bool) {
 	return u, true
 }
 
+// CreateView reports if the statement is a CREATE VIEW command with the given name.
+func (s *Stmt) CreateView(name string) (*Create_view_stmtContext, bool) {
```
masseelch, line 133: What is this typename?

cmd/atlas/internal/sqlparse/parseutil/parseutil.go
a8m: I actually don't like to use var block here, because I use them in the`if` right after.

cmd/atlas/internal/sqlparse/sqliteparse/sqliteparse.go
```diff
@@ -214,8 +250,14 @@ func isnull(t antlr.Tree) bool {
 }
 
 func unquote(s string) string {
-	if len(s) < 2 || s[0] != '`' || s[len(s)-1] != '`' {
-		return s
+	switch {
+	case len(s) < 2:
+	case s[0] == '`' && s[len(s)-1] == '`':
+		return s[1 : len(s)-1]
```
masseelch, line 256: I think strconv.Unquote works on backticks as well.

a8m: Thanks for this. I'll fix it.

a8m: "`" cannot be escaped, so calling Unquote is unnecessary. But better merging these two together.

masseelch (approved):

PR #1531, opened 2023-03-15
Update image.md

michaelcaulley:
docs: update docker command to match the format of docker run 
 IMAGE [COMMAND]

The `migrations` before the image name might be a typo. It didn't work for me and I can't find any documentation that helps me understand what it might be doing.


a8m (approved):
Thanks for fixing this, @michaelcaulley

PR #1548, opened 2023-03-21
Add support for LibSQL

haaawk:
Example usage:

```
atlas schema inspect -u "libsql://<db url>?authToken=<token>"
```

```
atlas schema apply -u "libsql://<db url>?authTokne=<token>" --to "sqlite:///test.db"
```


CodingDoug:
Hoping to see this in the libsql driver since it surfaces here: https://github.com/libsql/libsql-client-go/issues/25

CodingDoug:
I noticed that the inspection of a schema includes internal tables.

- libsql_wasm_func_table
- _litestream_seq
- _litestream_lock

It probably should filter those out, maybe in the same way the CLI filters the content of `.tables`.  It might be good to have this knowledge coded into a single place to apply to other similar tools.

This is related to:

- https://github.com/chiselstrike/turso-cli/issues/278
- https://github.com/chiselstrike/turso-cli/pull/304

haaawk:
Added support for integration tests.
To run them one can do:
```
export export LIBSQL_URL="libsql://<db url>?authToken=<token>"
cd internal/integration
go test -count=1 -p=1 -parallel=1 -run='TestSQLite_' ./...
```
It is needed to use -p=1 and -parallel=1 because the tests reuse a single db and can't work concurrently.
For now, I'm filtering internal tables in Atlas driver to unblock the effort until Libsql supports hiding internal tables.

haaawk:
I removed the commit that was making Atlas ignore litestream tables

internal/integration/sqlite_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"ariga.io/atlas/sql/schema"
 	"ariga.io/atlas/sql/sqlite"
 
+	_ "github.com/libsql/libsql-client-go/libsql"
```
a8m, line 23: Hey! Why are the tests are here? We agreed the tests will reside on a separate repo?

a8m, line 23: Hey! Why these changes are necessary? We agreed the tests will reside on a separate repo.

haaawk: I'm sorry - I must have misunderstood you. I thought that only libsql specific tests should live in separate repo. This PR makes it possible to run existing Atlas sqlite tests against libsql. It does not change any tests* - just adds ability to select different target. Default target is still sqlite and no one will experience any difference in tests behaviour unless they explicitly set the right env variables. I was hoping you would agree to merge this to master together with libsql support. That would help us avoid merge conflicts in case you decide to change sqlite tests in the future.

As we agreed, I created a separate repo that runs the tests once a day -> https://github.com/chiselstrike/atlas-integration-ci
This was my understanding of the separate repo idea. We (Libsql maintainers) maintain and monitor the integration repo to make sure integration keeps working.

There are too alternatives to what I did:
1. I copy all sqlite tests to the integration repo but then it creates unnecessary maintenance cost of keeping them synced with upstream
2. In integration repo, I could apply my patch to the current atlas master. This on the other hand will break when Atlas sqlite tests are changed

Would it be ok to merge this @a8m ? 

* - I had to restructure one test but functionally it's the same as before. I just had to wrap the code into `t.Run`

a8m: Why testing libsql is needed here? Atlas does not have any special handling besides opening a connection. i.e., the validation in the other repository is enough IMO.

Sorry for the misunderstanding and for being too picky here, we use the same (package) tests for our internal testing (and commercial drivers) and I prefer to not have it there. 

> I had to restructure one test but functionally it's the same as before. I just had to wrap the code into t.Run

Can you elaborate on why it's needed?

sql/sqlite/driver.go
```diff
@@ -60,6 +62,15 @@ func init() {
...
 	)
+	sqlclient.Register(
+		LibsqlDriverName,
+		sqlclient.DriverOpener(Open),
+		sqlclient.RegisterCodec(MarshalHCL, EvalHCL),
+		sqlclient.RegisterURLParser(sqlclient.URLParserFunc(func(u *url.URL) *sqlclient.URL {
+			uc := &sqlclient.URL{URL: u, DSN: "wss://" + strings.TrimPrefix(u.String(), u.Scheme+"://"), Schema: mainFile}
+			return uc
+		})),
+	)
```
a8m, line 73: I think a better way, that will also support our platform is to register a flavor (of `sqlite`) named `libsql` and just handle it differently using the URL parser/opener. Otherwise, this will require us to handle libsql in all our other integrations - which is much bigger effort.

a8m:
Closed with #1716

PR #1557, opened 2023-03-26
sql/postgres: change the Concurrently attribute to a schema.Clause

a8m:
The `Concurrently` option only affects how the index is created and doesn't provide any information about the index itself. 

cc @ivanvanderbyl

### DIFF
--- a/sql/postgres/migrate_test.go
+++ b/sql/postgres/migrate_test.go
@@ -534,13 +534,24 @@ func TestPlanChanges(t *testing.T) {
 							},
 							&schema.AddIndex{
 								I: &schema.Index{
-									Name: "concurrently",
+									Name: "add_con",
 									Parts: []*schema.IndexPart{
 										{C: users.Columns[0]},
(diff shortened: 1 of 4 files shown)


sql/postgres/migrate_test.go
```diff
@@ -581,8 +596,8 @@ func TestPlanChanges(t *testing.T) {
 						Reverse: `DROP INDEX "include_key"`,
 					},
 					{
-						Cmd:     `CREATE INDEX CONCURRENTLY "concurrently" ON "users" ("id")`,
-						Reverse: `DROP INDEX CONCURRENTLY "concurrently"`,
+						Cmd:     `CREATE INDEX CONCURRENTLY "concurrently1" ON "users" ("id")`,
+						Reverse: `DROP INDEX CONCURRENTLY "concurrently1"`,
```
rotemtam, line 600: ?

rotemtam (approved):

PR #1559, opened 2023-03-27
sql/postgres/postgrescheck: add concurrent index commands analyzer

a8m:
An analyzer that detects two cases:
1. Missing the `CONCURRENTLY` option in `CREATE INDEX`/`DROP INDEX` commands and report accordingly. Error (red CI) if was set this way. Note the code skips tables that were created in the same file.
2. Warning about missing `atlas:txmode none` directive in case the migration file contains concurrent commands (invalid operation).

Future improvements: As concurrent creation and deletion operations cannot run within a transaction, it is recommended that users remove unrelated statements (such as `ALTER TABLE`) from the file. This is because other commands can, and should, be executed within a transaction.

### DIFF
--- a/sql/postgres/postgrescheck/postgrescheck.go
+++ b/sql/postgres/postgrescheck/postgrescheck.go
@@ -5,10 +5,15 @@
 package postgrescheck
 
 import (
+	"context"
+	"errors"
 	"fmt"
 
 	"ariga.io/atlas/schemahcl"
+	"ariga.io/atlas/sql/internal/sqlx"
+	"ariga.io/atlas/sql/migrate"
 	"ariga.io/atlas/sql/postgres"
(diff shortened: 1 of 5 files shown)


sql/postgres/postgrescheck/postgrescheck.go
```diff
@@ -54,6 +59,136 @@ func init() {
...
+
+// Analyze implements sqlcheck.Analyzer.
+func (a *ConcurrentIndex) Analyze(_ context.Context, p *sqlcheck.Pass) error {
+	var (
+		notx  bool
+		notxC int
+		diags []sqlcheck.Diagnostic
+	)
+	// The txmode directive, currently defined in cmd/atlas,
+	// might be moved to sql/migrate in the future.
```
masseelch, line 128: Why is this comment here?

PR #1586, opened 2023-04-13
schema/inspect: support inspect for non-database resource

giautm:
This PR updates the `atlas schema inspect` command to support other resources: sql-file, hcl-file, and Ent's schema. When you do inspect those resources, you need to provide a dev-DB for Atlas to replay changes on it.

With an unexpected side effect, to fix what we believe to be a bug with the inspect command. The `--exclude` flag when used with schema connection can only exclude sub-resources such as: table, column, index, ... and no longer exclude the current schema.

Those commands are invalid
```shell
atlas schema inspect -u "mysql://user:pass@localhost:3306/dbname" --exclude=dbname
atlas schema inspect -u "mysql://user:pass@localhost:3306/dbname" --exclude=dbname.*
atlas schema inspect -u "mysql://user:pass@localhost:3306/dbname" --exclude=dbname.table1
```

Need to update to this:
```shell
atlas schema inspect -u "mysql://user:pass@localhost:3306/dbname" --exclude=*
atlas schema inspect -u "mysql://user:pass@localhost:3306/dbname" --exclude=table1
```

For connection to realm, the exclude flag doesn't affect by this PR. So the bellow command still working fine.
```shell
atlas schema inspect -u "mysql://user:pass@localhost:3306" --exclude=dbname
```

### DIFF
--- a/cmd/atlas/internal/cmdapi/schema.go
+++ b/cmd/atlas/internal/cmdapi/schema.go
@@ -470,10 +470,6 @@ func schemaInspectRun(cmd *cobra.Command, _ []string, flags schemaInspectFlags) 
 		return err
 	}
 	defer r.Close()
-	s, err := r.ReadState(ctx)
-	if err != nil {
-		return err
-	}
 	client, ok := r.Closer.(*sqlclient.Client)
 	if !ok && dev != nil {
 		client = dev
@@ -484,6 +480,7 @@ func schemaInspectRun(cmd *cobra.Command, _ []string, flags schemaInspectFlags) 
 			return fmt.Errorf("parse log format: %w", err)
 		}
 	}
+	s, err := r.ReadState(ctx)
 	return format.Execute(cmd.OutOrStdout(), &cmdlog.SchemaInspect{
 		Client: client,
 		Realm:  s,
--- a/cmd/atlas/internal/cmdapi/schema.go
+++ b/cmd/atlas/internal/cmdapi/schema.go
@@ -395,7 +395,8 @@ func schemaDiffRun(cmd *cobra.Command, _ []string, flags schemaDiffFlags) error 
 }
 
 type schemaInspectFlags struct {
-	url       string   // URL of database to apply the changes on.
+	url       string   // URL of resource to inspect.
+	devURL    string   // URL of the dev database.
 	logFormat string   // Format of the log output.
 	schemas   []string // Schemas to take into account when diffing.
 	exclude   []string // List of glob patterns used to filter resources from applying (see schema.InspectOptions).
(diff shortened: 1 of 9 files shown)


cmd/atlas/internal/cmdapi/schema_test.go
```diff
@@ -592,6 +592,28 @@ func TestSchema_InspectLog(t *testing.T) {
 	require.Equal(t, `{"schemas":[{"name":"main","tables":[{"name":"t1","columns":[{"name":"id","type":"INTEGER","null":true}],"primary_key":{"parts":[{"column":"id"}]}},{"name":"t2","columns":[{"name":"name","type":"TEXT","null":true}]}]}]}`, s)
 }
 
+func TestSchema_InspectFile(t *testing.T) {
+	p := filepath.Join(t.TempDir(), "schema.sql")
+	os.WriteFile(p, []byte(`create table t1 (id integer primary key);create table t2 (name text);`), 0600)
+	db := openSQLite(t, "")
+	cmd := schemaCmd()
+	cmd.AddCommand(schemaInspectCmd())
+	s, err := runCmd(
```
masseelch, line 601: Why this?

internal/integration/testdata/postgres/cli-migrate-apply.txt
masseelch, line 1: This changes mean the change you did so `inspect` command is backwards incompatible. We need to decide how to handle that: 

1. Consider the old behavior a bug and communicate such
2. Add code to support the old behavior
3. ???

cmd/atlas/internal/cmdapi/schema.go
```diff
@@ -445,19 +447,37 @@ flag.
...
+	})
 	if err != nil {
 		return err
 	}
-	defer client.Close()
-	schemas := flags.schemas
-	if client.URL.Schema != "" {
-		schemas = append(schemas, client.URL.Schema)
+	defer r.Close()
+	client, ok := r.Closer.(*sqlclient.Client)
```
a8m, line 473: Keep it close to its usage below? (After reading the state).

cmd/atlas/internal/cmdapi/schema_test.go
```diff
@@ -592,6 +592,28 @@ func TestSchema_InspectLog(t *testing.T) {
...
+	)
+	require.NoError(t, err)
+	require.Equal(t, strings.Join([]string{
+		`-- Create "t1" table`,
+		"CREATE TABLE `t1` (`id` integer NULL, PRIMARY KEY (`id`));",
+		`-- Create "t2" table`,
+		"CREATE TABLE `t2` (`name` text NULL);",
+		"",
+	}, "\n"), s)
+}
```
a8m, line 615: Add another case without —dev-url and expect failure?

giautm: Yes, but I will replace this with testscript

cmd/atlas/internal/cmdapi/schema.go
giautm: Humm. Usually, we keep the call to close() close to the open function, no?

masseelch (approved):

cmd/atlas/internal/cmdapi/schema.go
a8m: I meant the definition of the `client` variable.

a8m (approved):
Looks good, but see my only comment there.

PR #1588, opened 2023-04-15
sql/migrate: increase error verbosity

masseelch:


sql/migrate/lex.go
```diff
@@ -174,10 +174,15 @@ func (l *lex) addPos(p int) {
 }
 
 func (l *lex) skipQuote(quote rune) error {
+	pos := l.pos
 	for {
 		switch r := l.next(); {
 		case r == eos:
-			return fmt.Errorf("unclosed quote %q", quote)
+			end := pos + 15
```
giautm, line 181: Why this number? Can you please put some comment?

masseelch: Arbitrary number to provide some more feedback to where the unclosed quote is

What do you think is a reasonable length?

rotemtam: From the tests it looks good . Just put it in a const with some descriptive name with some comment

a8m: Position is not that good. Change it to line number + column number. Then, adding this string prefix is not needed.

sql/migrate/lex.go
```diff
@@ -174,10 +174,21 @@ func (l *lex) addPos(p int) {
...
+				s    = l.input[:pos]
+				col  = strings.LastIndex(s, `\n`)
+				line = 1 + strings.Count(s, `\n`)
+			)
+			if line == 1 {
+				col = pos
+			} else {
+				col = pos - col - 2
+			}
+			return fmt.Errorf("unclosed quote %q at %d:%d", quote, line, col)
```
a8m, line 191: Maybe add an `error/f` method and report this format for all errors:

```go
file:%d:%d: <message>
```
`file` can be constant or omitted as we don't have this info.

sql/migrate/lex.go
```diff
@@ -281,3 +282,18 @@ func (l *lex) setDelim(d string) error {
 	l.delim = strings.NewReplacer(`\n`, "\n", `\r`, "\r", `\t`, "\t").Replace(d)
 	return nil
 }
+
+func (l *lex) error(pos int, format string, args ...any) error {
+	format = "%d:%d: " + format
+	var (
+		s    = l.input[:pos]
+		col  = strings.LastIndex(s, `\n`)
+		line = 1 + strings.Count(s, `\n`)
```
a8m, line 291:
```suggestion
		col  = strings.LastIndex(s, "\n")
		line = 1 + strings.Count(s, "\n")
```

sql/migrate/lex_test.go
```diff
@@ -105,3 +105,57 @@ cmd7;
...
+			err:  `1:1: unclosed quote '"'`,
+		},
+		{
+			name: "unclosed double at 2:2",
+			stmt: `unclosed double at 2:2\n "`,
+			err:  `2:2: unclosed quote '"'`,
+		},
+		{
+			name: "unclosed double at 5:5",
+			stmt: `unclosed double at 2:2\n\n\n\n    "`,
```
a8m, line 140:
```suggestion
			stmt: "unclosed double at 2:2\n\n\n\n    "",
```
\n inside raw strings are escaped

a8m (approved):
💯

PR #1593, opened 2023-04-17
cmd/atlas: allow attaching identifiers to revisions-table

a8m:


giautm (approved):
💯

a8m:
I removed the `entgo.io/ent/entc/integration` tests, as they were added before the Ent integration was official and they exist now inside the Ent repository (`entgo.io/ent/entc/integration/migrate`).

PR #1638, opened 2023-05-08
cmd/atlas: fallback to 'default' project name if was not provided

a8m:


rotemtam (approved):

cmd/atlas/internal/cmdext/cmdext.go
```diff
@@ -317,6 +317,9 @@ type AtlasConfig struct {
 	Project string           // Optional project.
 }
 
+// DefaultProjectName is the default name for prjects.
```
masseelch, line 320:
```suggestion
// DefaultProjectName is the default name for projects.
```

PR #1648, opened 2023-05-11
sql/postgres: make index storage check continue

### LINKED ISSUE #1472: Index Storage Parameter - Validation issue
This line [right here](https://github.com/ariga/atlas/blob/master/sql/postgres/inspect.go#L1024) checks for validity of a PSQL Index Storage Parameter during inspection. I'm no PSQL expert, but it seems like this storage parameter is just a generic string and only has meaning in the context of what ever index type is using it? Is there any way to treat these as opaque blobs in Atlas, or MUST they be parsed?

We use [ZomboDB](https://www.zombodb.com/) which has very bespoke syntax for defining cross-indexes via the Index Storage Parameters, and Atlas chokes with the error `invalid index storage parameter: options=customers:(customer_id=<public.customers.idx_zdb_customers>id)` 

Replicating this would require you setup both ZomboDB and an Elastic instance, so it's not super straightforward. But the below SQL would cause the issue:
```sql
-- Simple parent tables
create table parent (id serial primary key);

-- Single child table for a 1:m relationship
create table child (
    id serial primary key,
    parent_id integer constraint fk_parent_id references parent
);

-- Index for both parent and child tables individually
create index idx_zdb_parent on parent using zombodb ((parent.*)) with (url='http:dokku.elasticsearch.stage:9200/');
create index idx_zdb_child on child using zombodb ((child.*)) with (url='http:dokku.elasticsearch.stage:9200/');

-- The problematic cross-join index, it doesn't like the two `=` occurrences.
alter index idx_zdb_parent set (options='children:(parent_id=<public.parent.idx_zdb_parent>id)');
```

> Note: this issue is from [this](https://discord.com/channels/930720389120794674/933360123357839360/1073043916271652895) Discord thread.

AThilenius:
Fixes #1472 I have no idea if this causes issues downstream, but it seems to fix the immediate issue I have with Atlas. Looking at the logic itself, I don't see a good reason to fail the param parsing though; non-kvp-pairs in storage params doesn't indicate a bad state, as evidenced by our use case. They just aren't the specific params that are trying to be parsed.

### DIFF
--- a/sql/postgres/inspect.go
+++ b/sql/postgres/inspect.go
@@ -1024,7 +1024,7 @@ func newIndexStorage(opts string) (*IndexStorageParams, error) {
 	for _, p := range strings.Split(strings.Trim(opts, "{}"), ",") {
 		kv := strings.Split(p, "=")
 		if len(kv) != 2 {
-			return nil, fmt.Errorf("invalid index storage parameter: %s", p)
+			continue
 		}
 		switch kv[0] {
 		case "autosummarize":


AThilenius:
🫤

PR #1655, opened 2023-05-13
sql/mysql: print non-default storage engines

a8m:


sql/mysql/sqlspec.go
```diff
@@ -255,6 +256,17 @@ func tableSpec(t *schema.Table) (*sqlspec.Table, error) {
 	if c, ok := hasCollate(t.Attrs, t.Schema.Attrs); ok {
 		ts.Extra.Attrs = append(ts.Extra.Attrs, schemahcl.StringAttr("collate", c))
 	}
+	// Marshal the engine attribute only if it is not InnoDB (default).
```
masseelch, line 259: why?

a8m: Because it will change the generated schema for users, and we also don't marshal default attributes today (charset, collation, BTREE, etc) in all drivers

masseelch (approved):

PR #1674, opened 2023-05-19
cmd/atlas/internal/cmdext: use dev-database schema in ent loader as default schema

a8m:
image


a8m:
Currently, Atlas returns an error in case tables don't have schemas. Soon, we'll add support for generating HCL schemas without `schema` blocks - make it more dynamic.

giautm (approved):
😍

PR #1697, opened 2023-05-31
sql/migrate: fixed archive logic on sum file

giautm:


sql/migrate/dir.go
```diff
@@ -561,8 +561,11 @@ func ArchiveDir(dir Dir) ([]byte, error) {
 	tw := tar.NewWriter(&buf)
 	defer tw.Close()
 	sumF, err := dir.Open(HashFileName)
-	if err != nil && !errors.Is(err, fs.ErrNotExist) {
-		return nil, err
+	if err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
+			return nil, err
+		}
+		sumF = nil
```
a8m, line 568: Isn't this the default returned from `dir.Open` in case of an error?

masseelch: If it is not, the fix should be in dir.Open, not here.

a8m (approved):

PR #1716, opened 2023-06-04
merge libsql

a8m:


cmd/atlas/internal/sqlparse/sqlparse.go
```diff
@@ -52,5 +52,4 @@ func init() {
 	Register(mysql.DriverName, &myparse.Parser{})
 	Register(postgres.DriverName, &pgparse.Parser{})
 	Register(sqlite.DriverName, &sqliteparse.FileParser{})
-	Register(sqlite.LibsqlDriverName, &sqliteparse.FileParser{})
```
haaawk: Isn't this needed @a8m ?

sql/sqlite/driver.go
```diff
@@ -60,6 +60,16 @@ func init() {
 			return uc
 		})),
 	)
+	sqlclient.Register(
+		"libsql",
+		sqlclient.DriverOpener(Open),
+		sqlclient.RegisterTxOpener(OpenTx),
+		sqlclient.RegisterCodec(MarshalHCL, EvalHCL),
+		sqlclient.RegisterFlavours("libsql+wss", "libsql+https"),
```
haaawk, line 68: At the moment +https won't work as it does not support interactive transactions and is missing some features. We're working on making it better but maybe it's worth to keep https out of scope for this PR?

doc/md/concepts/url.mdx
```diff
@@ -101,6 +101,14 @@ Connecting to an in-memory SQLite database (ephemeral). Useful for `--dev-url`:
 sqlite://file?mode=memory&_fk=1
 ```
 
+Atlas also supports WebSocket and HTTPS connections to remote `libsql` databases:
+
+```shell
+libsql+wss://database-url
+
+libsql+https://database-url
```
haaawk, line 109: https support is under development so it might be a good idea to keep it out of scope of this PR

haaawk (approved):
I've tested the PR against Turso (managed Libsql) and updated https://github.com/chiselstrike/atlas-integration-ci to use the `haaawk-libsql` branch for testing. All tests are passing.

cmd/atlas/internal/sqlparse/sqlparse.go
a8m: No. For `migrate lint` we recommend users to pass a URL to an in-memory SQLite database. The semantics are the same

PR #1736, opened 2023-06-11
integration: fix run `go mod tidy`

giautm:
This keep `atlas-cli` in-sync with atlas package (via replace). Also allow to freely run `go mod tidy` inside `integration` directory


internal/integration/tools.go
```diff
@@ -0,0 +1,9 @@
+//go:build tools
+// +build tools
+
+package main
+
+import (
+	_ "ariga.io/atlas/cmd/atlas"
+	_ "entgo.io/ent"
```
giautm, line 8: This `tools.go` is needed for keep ent and cmd/atlas as directly deps.

a8m (approved):

PR #1737, opened 2023-06-13
cmd/atlas: initial support for external schema data sources

a8m:


cmd/atlas/internal/cmdext/cmdext.go
```diff
@@ -441,6 +443,57 @@ func SchemaHCL(ctx *hcl.EvalContext, block *hclsyntax.Block) (cty.Value, error)
...
+	if len(attrs) > 0 {
+		return cty.NilVal, errorf("unexpected attributes: %v", attrs)
+	}
+	if len(args.Program) == 0 {
+		return cty.NilVal, errorf("program cannot be empty")
+	}
+	cmd := exec.Command(args.Program[0], args.Program[1:]...)
+	if args.Dir != "" {
+		cmd.Dir = args.Dir
+	}
```
masseelch, line 472: What about env variables?

a8m: The default behavior is the environment variables defined in the process. Do you mean updating the PWD if Dir was set?

cmd/atlas/internal/cmdext/cmdext.go
```diff
@@ -441,6 +443,57 @@ func SchemaHCL(ctx *hcl.EvalContext, block *hclsyntax.Block) (cty.Value, error)
 	}), nil
 }
 
+// SchemaExternal is a data source that reads an Atlas SQL/HCL schema state from external program.
```
rotemtam, line 446:
```suggestion
// SchemaExternal is a data source that reads a SQL schema state from external program.
```

rotemtam (approved):

giautm (approved):
💯

Issue #1727, opened 2023-06-08
MySQL performance issue with lint command with some tables


jaugustin:
Hello,

I hit a performance issue when running the lint command.

I created a repo to reproduce the issue https://github.com/jaugustin/atlas_slow_lint

If I run the `atlas migrate lint --latest 1 --dev-url docker://mysql/8/dev` then it get stuck for 15 minutes

The issue is from a mysql command executed to get tables / foreign keys that get stuck in `checking permissions` for some time

I can try to provide more information if needed

jaugustin:
@a8m I tag you as asked on discord :wink:

jaugustin:
here is the explain analyse if it can help
```
mysql> explain analyze SELECT 
    -> t1.CONSTRAINT_NAME, 
    -> t1.TABLE_NAME, 
'tab    -> t1.COLUMN_NAME,
    ->  t1.TABLE_SCHEMA, 
     -> t1.REFERENCED_TABLE_NAME, 
    -> t1.REFERENCED_COLUMN_NAME, 
    -> t1.REFERENCED_TABLE_SCHEMA, 
    -> t3.UPDATE_RULE, t3.DELETE_RULE 
    -> FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE AS t1 
    -> JOIN INFORMATION_SCHEMA.TABLE_CONSTRAINTS AS t2 
    -> JOIN INFORMATION_SCHEMA.REFERENTIAL_CONSTRAINTS AS t3 ON t1.CONSTRAINT_NAME = t2.CONSTRAINT_NAME AND t1.CONSTRAINT_NAME = t3.CONSTRAINT_NAME AND t1.TABLE_SCHEMA = t2.TABLE_SCHEMA AND t1.TABLE_SCHEMA = t3.CONSTRAINT_SCHEMA 
    -> WHERE t2.CONSTRAINT_TYPE = 'FOREIGN KEY' AND t1.TABLE_SCHEMA = 'dev' AND t1.TABLE_NAME 
[...]

PR #1740, opened 2023-06-13
Remove orderby to prevent performance issue in MySQL inspect queries

jaugustin:
I try to fix the performace issue https://github.com/ariga/atlas/issues/1727

For that I removed the order by clause in the FksQuery.

On my test it fix the performance issue, but I am not sure if it can cause behavior issue

@a8m can you look at it ?


jaugustin:
@a8m do you think this PR could be accepted soon ? it block us from using atlas for now ;)

a8m:
Hey, @jaugustin!

Thanks for the contribution. I'll give this a look later today/week.

sql/mysql/inspect.go
```diff
@@ -789,10 +789,7 @@ FROM
 WHERE
 	t2.CONSTRAINT_TYPE = 'FOREIGN KEY'
 	AND t1.TABLE_SCHEMA = ?
-	AND t1.TABLE_NAME IN (%s)
-ORDER BY
-	t1.CONSTRAINT_NAME,
-	t1.ORDINAL_POSITION`
```
a8m, line 795: This is problematic because this ensures we get the columns in the right order. See: https://github.com/ariga/atlas/blob/master/sql/internal/sqlx/sqlx.go#L130-L135

a8m: We can order this in the application code, but I want to make sure before this actually causes the slowness in the inspection.

a8m:
Hey, @jaugustin! Just to update you that I was able to resolve the issue and I'll push it later today. I'll keep you updated.

sql/mysql/inspect.go
jaugustin: @a8m I think you figure out the root cause. The order was causing an issue with MySQL planner, I shared the explain in this ticket https://github.com/ariga/atlas/issues/1727

I found another way to resolve the issue by spliting the query and do the join in Go, but I am not a go Expert :wink:

a8m:
Issue was resolved with https://github.com/ariga/atlas/pull/1748 (see the diff there). Would you mind giving the canary version a try, and report if that solves your issue? Thanks 🙏🏻

jaugustin:
I will test it now.

rotemtam:
Thanks for your detailed report and help investors this issue @jaugustin , well done !

PR #1745, opened 2023-06-15
sql/sqlite: extract table options (without rowid, strict) from sql

noamcattan:


sql/sqlite/inspect.go
```diff
@@ -386,14 +386,26 @@ func (i *inspect) tables(ctx context.Context, opts *schema.InspectOptions) ([]*s
...
-			t.Attrs = append(t.Attrs, &WithoutRowID{})
-		}
+		t.Attrs = append(t.Attrs, tableOptions(stmt)...)
 		tables = append(tables, t)
 	}
 	return tables, nil
 }
 
+// Parse the table options from a creation statement (without rowid, strict).
+func tableOptions(stmt string) (opts []schema.Attr) {
```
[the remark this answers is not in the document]
a8m: But note this:
> The table_list pragma first appeared in SQLite version 3.37.0 (2021-11-27)

That's fine because STRICT is also relatively new.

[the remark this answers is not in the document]
a8m: Also, ensure you do it once (not query the database for each table). e.g., 

```sql
select sqlite_master.name, sql, wr, strict from sqlite_master join pragma_table_list(sqlite_master.name);
```

sql/sqlite/inspect.go
```diff
@@ -715,7 +727,7 @@ const (
 	databasesQuery     = "SELECT `name`, `file` FROM pragma_database_list() WHERE `name` <> 'temp'"
 	databasesQueryArgs = "SELECT `name`, `file` FROM pragma_database_list() WHERE `name` IN (%s)"
 	// Query to list database tables.
-	tablesQuery = "SELECT `name`, `sql` FROM sqlite_master WHERE `type` = 'table' AND `name` NOT LIKE 'sqlite_%' AND `name` NOT LIKE 'libsql_%'"
+	tablesQuery = "SELECT sqlite_master.name, sqlite_master.sql, `wr`, `strict` FROM sqlite_master join(pragma_table_list(sqlite_master.name)) WHERE sqlite_master.type = 'table' AND sqlite_master.name NOT LIKE 'sqlite_%' AND sqlite_master.name NOT LIKE 'libsql_%'"
```
a8m, line 730: Use `pragma_table_list` only in new versions of SQLite. See how we do it in MySQL: https://github.com/ariga/atlas/blob/master/sql/mysql/inspect.go#L695-L696 (we have 2 different statements).

sql/sqlite/inspect.go
```diff
@@ -375,8 +375,11 @@ func (i *inspect) tables(ctx context.Context, opts *schema.InspectOptions) ([]*s
 	defer rows.Close()
 	var tables []*schema.Table
 	for rows.Next() {
-		var name, stmt string
-		if err := rows.Scan(&name, &stmt); err != nil {
+		var (
+			name, stmt string
+			wr, strict int
```
a8m, line 380: These can be boolean. Also, use `sql.NullBool` as we have 2 different queries.

sql/sqlite/inspect.go
noamcattan: I played with it a bit and it seems to be backwards compatible. since strict was also introduced in version 3.37, databases created by older client will always show strict=0.

a8m, line 730: The query became a bit complex, add line breaks.
```suggestion
	tablesQuery = `
SELECT
	sqlite_master.name, sqlite_master.sql, wr, strict
FROM
	sqlite_master
	JOIN pragma_table_list(sqlite_master.name)
WHERE
	sqlite_master.type = 'table'
	AND sqlite_master.name NOT LIKE 'sqlite_%'
	AND sqlite_master.name NOT LIKE 'libsql_%'`
```

a8m (commented):
Looks good. See minor comments

a8m:
Please rebase this branch.

a8m (approved):
🚀

PR #1755, opened 2023-06-18
sql/sqlx: move Charset/Collate to sqlx

giautm:


sql/mysql/sqlspec.go
```diff
@@ -257,10 +257,10 @@ func tableSpec(t *schema.Table) (*sqlspec.Table, error) {
 	if err != nil {
 		return nil, err
 	}
-	if c, ok := hasCharset(t.Attrs, t.Schema.Attrs); ok {
+	if c, ok := sqlx.Charset(t.Attrs, t.Schema.Attrs); ok {
```
a8m, line 260: HasCharset*

Same for collate.

a8m (approved):

PR #1756, opened 2023-06-18
sql/schema: added Scale attribute for TimeType

giautm:


a8m (commented):
Add DSL option + test: https://github.com/ariga/atlas/blob/master/sql/schema/dsl.go#L466

a8m (approved):

PR #1763, opened 2023-06-20
sql/sqlite: sqlspec for table options

noamcattan:


sql/sqlite/sqlspec.go
```diff
@@ -62,7 +62,29 @@ func MarshalSpec(v any, marshaler schemahcl.Marshaler) ([]byte, error) {
...
+	if attr, ok := spec.Extra.Attr("strict"); ok {
+		b, err := attr.Bool()
+		if err != nil {
+			return nil, err
+		}
+		if b {
+			tab.AddAttrs(&Strict{})
+		}
+	}
+	return tab, nil
```
a8m, line 87:
```suggestion
	t, err := specutil.Table(spec, parent, convertColumn, specutil.PrimaryKey, convertIndex, specutil.Check)
	if err != nil {
		return nil, err
	}
	if attr, ok := spec.Attr("without_rowid"); ok {
		b, err := attr.Bool()
		if err != nil {
			return nil, err
		}
		if b {
			tab.AddAttrs(&WithoutRowID{})
		}
	}
	if attr, ok := spec.Attr("strict"); ok {
		b, err := attr.Bool()
		if err != nil {
			return nil, err
		}
		if b {
			tab.AddAttrs(&Strict{})
		}
	}
	return t, nil
```

sql/sqlite/sqlspec.go
```diff
@@ -114,14 +136,26 @@ func schemaSpec(s *schema.Schema) (*specutil.SchemaSpec, error) {
...
+		return nil, err
+	}
+	for _, a := range tab.Attrs {
+		switch a.(type) {
+		case *WithoutRowID:
+			spec.Extra.SetAttr(schemahcl.BoolAttr("without_rowid", true))
+		case *Strict:
+			spec.Extra.SetAttr(schemahcl.BoolAttr("strict", true))
+		}
+	}
```
a8m, line 157: Embed it like [this](https://github.com/ariga/atlas/blob/master/sql/internal/specutil/convert.go#L394-L400) to ensure it's added **after** the columns.

sql/sqlite/sqlspec_test.go
```diff
@@ -423,6 +431,43 @@ schema "test" {
...
+						Type: &schema.ColumnType{Type: &schema.IntegerType{T: "int"}},
+					},
+				},
+				Attrs: []schema.Attr{
+					&WithoutRowID{},
+					&Strict{},
+				},
+			},
+		},
+	}
```
a8m, line 452:
```suggestion
	s := &schema.Schema{
		Name: "test",
		Tables: []*schema.Table{
			{
				Name: "users",
				Columns: []*schema.Column{
					{
						Name: "id",
						Type: &schema.ColumnType{Type: &schema.IntegerType{T: "int"}},
					},
				},
				Attrs: []schema.Attr{
					&WithoutRowID{},
					&Strict{},
				},
			},
		},
	}
	s := schema.New("test").
		AddTables(
			schema.NewTable("users").
				AddColumn(
					schema.NewIntColumn("id", "int"),
				).
				AddAttrs(
					&WithoutRowID{},
					&Strict{},
				),
		)
```

a8m (approved):

PR #1765, opened 2023-06-20
sql/mssql: inspect columns and more tests

giautm:
- sql/mssql: inspect columns and tests


sql/mssql/inspect.go
```diff
@@ -367,6 +391,17 @@ ORDER BY
 )
 
 type (
+	// MemoryOptimized attribute describes if the table is memory-optimized or disk-based.
+	MemoryOptimized struct {
+		schema.Attr
+		V bool
```
a8m, line 397:
```suggestion
```

I think you can remove `V`. Attaching this attribute to a table indicates that it's memory-optimized.

a8m (approved):

a8m (approved):

PR #1782, opened 2023-06-24
sql/mssql: inspect column default value

giautm:
- sql/mssql: inspect checks
- sql/mssql: inspect computed column
- sql/mssql: inspect column default value

### DIFF
--- a/sql/mssql/inspect.go
+++ b/sql/mssql/inspect.go
@@ -280,14 +280,15 @@ func (i *inspect) addColumn(s *schema.Schema, rows *sql.Rows, scope queryScope) 
 		nullable, userDefined                     sql.NullInt64
 		identity, identitySeek, identityIncrement sql.NullInt64
 		size, precision, scale, isPersisted       sql.NullInt64
-		genexpr                                   sql.NullString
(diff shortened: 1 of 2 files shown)


sql/mssql/inspect.go
```diff
@@ -373,6 +444,29 @@ func nArgs(start, n int) string {
 	return b.String()
 }
 
+// defaultExpr returns the default expression of the given column.
+//
+// https://learn.microsoft.com/en-us/sql/relational-databases/tables/specify-default-values-for-columns
+func (i *inspect) defaultExpr(_ *schema.Column, x string) schema.Expr {
+	// Remove the parenthesis from the expression.
+	x = mayUnwrap(x)
```
a8m, line 452: Why do you remove this if it’s returned like this from the database? Also, sqlx has some code for checking balanced parens, you might want to use it instead.

giautm: MS-SQL alway wrap with double parenthesis. For example, `6` will become `((6))`, a string `'6'` become `('6')`, and a function `getdate()` become `(getdate())`.

So, unwrap for checking the expression type.

PR #1788, opened 2023-06-26
sql/mssql: driver diff implement

giautm:


sql/mssql/diff.go
```diff
@@ -17,51 +22,300 @@ var DefaultDiff schema.Differ = &sqlx.Diff{DiffDriver: &diff{}}
...
+func (d *diff) IsGeneratedIndexName(t *schema.Table, i *schema.Index) bool {
+	// MS-SQL generates index names for primary keys.
+	return fmt.Sprintf("PK_%s", t.Name) == i.Name
 }
 
 // IndexAttrChanged reports if the index attributes were changed.
-// The default type is BTREE if no type was specified.
-func (*diff) IndexAttrChanged(_, _ []schema.Attr) bool {
-	// Not implemented yet.
+// The default type is CLUSTERED if no type was specified.
```
a8m, line 87: The comment is not correct.

sql/mssql/diff.go
```diff
@@ -17,51 +22,300 @@ var DefaultDiff schema.Differ = &sqlx.Diff{DiffDriver: &diff{}}
...
+			}
+		}
+	case fromC.V != toC.V:
+		return &schema.ModifyAttr{
+			From: &fromC,
+			To:   &toC,
+		}
+	}
+	return noChange
+}
```
a8m, line 148: Did you verify this behavior in SQL Server? Because the code was taken from the current drivers.

a8m (approved):

sql/mssql/diff.go
giautm: Checked with their document, SQL Server has three objects that has collation attribute are: server, database, and column. And because we only managed schema, table and column. Only thing we can update here is the column.

PR #1844, opened 2023-07-10
internal/integration: support running testsciprt in ent builds

a8m:


masseelch (approved):

sql/postgres/diff.go
```diff
@@ -336,6 +336,12 @@ func (d *diff) typeChanged(from, to *schema.Column) (bool, error) {
 			return false, err
 		}
 		changed = t1 != t2
+	case *UserDefinedType:
```
masseelch, line 339: Does this belong in this PR?

PR #1861, opened 2023-07-16
sql/sqlx: add TableResource and Quote helpers

giautm:


sql/internal/sqlx/sqlx.go
```diff
@@ -261,9 +261,20 @@ func (b *Builder) Table(t *schema.Table) *Builder {
...
 // with the schema name if exists.
-func (b *Builder) mayQualify(s *schema.Schema, ident string) *Builder {
+func (b *Builder) TableResource(t *schema.Table, r any) *Builder {
+	switch c := r.(type) {
+	case *schema.Column:
+		return b.mayQualify(t.Schema, t.Name, c.Name)
+	case *schema.Index:
+		return b.mayQualify(t.Schema, t.Name, c.Name)
+	default:
+		return b
```
a8m, line 273:
```suggestion
		panic("unexpected table resource: %T", r)
```

sql/internal/sqlx/sqlx.go
```diff
@@ -277,7 +288,12 @@ func (b *Builder) mayQualify(s *schema.Schema, ident string) *Builder {
 		b.Ident(s.Name)
 		b.rewriteLastByte('.')
 	}
-	b.Ident(ident)
+	for i, ident := range idents {
```
a8m, line 291: idents cannot be an empty slice, because we rely on it to have at least one (we call `rewriteLastByte` above). 

What do you say about changing its signature to: 

```go
mayQualify(s *schema.Schema, top string, children ...string) *Builder
```

a8m (commented):
See small comments

a8m (approved):

PR #1868, opened 2023-07-16
web/projects-amount-prop

dimagreat:
Allow to use prop to change project amount number


doc/website/src/pages/index.js
```diff
@@ -21,6 +21,7 @@ export default function () {
             ],
             isHidden: false,
         }}
+        projectsAmount={3000}
```
rotemtam, line 24:
```suggestion
        projectsAmount={2000}
```

PR #1878, opened 2023-07-19
schemahcl: add tf function for simple reading files

a8m:


rotemtam (approved):

AlexandreBiguet:
Hello 👋 
thanks a lot for this P.R 🙏 (I just started with atlas, and I was looking for a way to read the content of a file to integrate with docker secrets) 

I was wondering if this will be part of the next release, and if it is when do you plan to do the release? 

Thanks again :)

PR #1889, opened 2023-07-24
internal/migrate: fixed path on Windows

giautm:


giautm:
I will add the test after we have runner on Windows

a8m (approved):

PR #1893, opened 2023-07-24
sql/mysql: inspect fulltext index parsers

a8m:


sql/mysql/migrate.go
```diff
@@ -587,7 +587,12 @@ func index(b *sqlx.Builder, idx *schema.Index) {
 	switch t := indexType(idx.Attrs); {
 	case idx.Unique:
 		b.P("UNIQUE")
-	case t.T == IndexTypeFullText || t.T == IndexTypeSpatial:
+	case t.T == IndexTypeFullText:
+		if p := (&IndexParser{}); sqlx.Has(idx.Attrs, p) {
+			defer func() { b.P("WITH PARSER").Ident(p.P) }()
```
giautm, line 592: I refer to handle options as the latest step, instead of using defer()

giautm (approved):

PR #1894, opened 2023-07-25
cmd/atlas/internal/cloudapi: add visualize stubs

masseelch:


cmd/atlas/internal/cloudapi/client_oss.go
```diff
@@ -0,0 +1,10 @@
+//go:build !ent
+
+package cloudapi
+
+import "context"
+
+// Visualize a schema and return a URL to the visualization.
+func (c *Client) Visualize(ctx context.Context, input VisualizeInput) (string, error) {
+	return "", nil // unimplemented
```
giautm, line 9: Return error instead? Just say this is Ent feature

Issue #1896, opened 2023-07-25
Unable to apply changes if schema contains ENUM


MrWitold:
Unable to apply changes if schema contains ENUM

TL;DR: First apply to database is working, however next update is never working because every time atlas is trying to delete enum which is used by table. 

**schema.sql before the edit**
```
CREATE TYPE e_status AS ENUM (
    'new',
    'created'
    );


CREATE TABLE IF NOT EXISTS user
(
    id            SERIAL PRIMARY KEY,
    status        e_status        DEFAULT 'new'::e_status NOT NULL,
    email         VARCHAR      NOT NULL
);
```

```
atlas migrate diff init \
  --dir "file://resources/migrations" \
  --to "file://resources/schema.sql" \
  --dev-url "postgres://postgres:docker@localhost:5432/docker?search_path=public&sslmode=disable"
```

Then this file is generated `20230725095338_init.sql`
```
-- Create enum type "e_status"
CREATE TYPE "public"."e_status" AS ENUM ('new', 'created');
-- Create "evaluations" table
CREATE TABLE "public"."evaluations" ("id" serial NOT NULL, "status" "public"."e_status" NOT NULL DEFAULT 'new', "email" character varying NOT NULL, PRIMARY KEY ("id"));
```

```
atlas migrate apply \
  --url "postgres://postgres:docker@localhost:5432/postgres?search_path=public&sslmode=disable" \
  --dir "file://resources/migrations"
Migrating to version 20230725095338 (1 migrations in total):

  -- migrating version 20230725095338
    -> CREATE TYPE "public"."e_status" AS ENUM ('new', 'created');
    -> CREATE TABLE "public"."evaluations" ("id" serial NOT NULL, "status" "public"."e_status" NOT NULL DEFAULT 'new', "email" character varying NOT NULL, PRIMARY KEY ("id"));
  -- ok (18.17857ms)

  -------------------------
  -- 70.28426ms
  -- 1 migrations
  -- 2 sql statements
```
So until this moment everything goes fine.
___

**schema.sql after the edit**
Only column `login` is added
```
CREATE TYPE e_status AS ENUM (
    'new',
    'created'
    );


CREATE TABLE IF NOT EXISTS user
(
    id            SERIAL PRIMARY KEY,
    status        e_status        DEFAULT 'new'::e_status NOT NULL,
    email         VARCHAR      NOT NULL,
    login         VARCHAR         NOT NULL
);
```

```
atlas migrate diff add_login_column \
[...]

PR #1900, opened 2023-07-26
cmd/atlas: apply only revisions-table changes on automatic migration

a8m:
Since Atlas now handles enums independently, running Ent automatic migration will try to drop them. Luckily, operations will fail because other resources depend on them. 

Fixed: #1896


cmd/atlas/internal/migrate/migrate_test.go
```diff
@@ -24,6 +24,8 @@ func TestNewEntRevisions(t *testing.T) {
 	require.NoError(t, err)
 	r, err := NewEntRevisions(ctx, c)
 	require.NoError(t, err)
+	_, err = c.DB.Exec("CREATE VIEW v1 AS SELECT 1;")
+	require.NoError(t, err)
 	require.NoError(t, r.Migrate(ctx))
```
masseelch, line 29: Shouldn't you check the view was not changed by the migrate here?

masseelch (approved):

PR #1944, opened 2023-08-06
Remove executable bits in `os.WriteFile`

ramikg:
I've changed the file mode passed to `os.WriteFile` from 0777 to 0666.
That is, I've removed the executable bits, which are unnecessary and may lead to undesirable consequences.


rotemtam (approved):

rotemtam:
Thanks for your contribution @ramikg !

PR #1971, opened 2023-08-10
sql/migrate: add support for baseline/tags files 

a8m:
This is currently a work in progress and should not be used until the documentation will be updated with this info.

### DIFF
--- a/sql/migrate/dir.go
+++ b/sql/migrate/dir.go
@@ -38,12 +38,6 @@ type (
 		Checksum() (HashFile, error)
 	}
 
-	// Formatter wraps the Format method.
-	Formatter interface {
-		// Format formats the given Plan into one or more migration files.
-		Format(*Plan) ([]File, error)
-	}
-
 	// File represents a single migration file.
 	File interface {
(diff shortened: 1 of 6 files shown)


sql/migrate/dir.go
```diff
@@ -59,6 +53,50 @@ type (
...
+		WriteCheckpoint(name, tag string, content []byte) error
+
+		// CheckpointFiles returns a set of checkpoint files stored in this Dir,
+		// ordered by name.
+		CheckpointFiles() ([]File, error)
+
+		// FilesFromCheckpoint returns the files to be executed on a database from
+		// the given checkpoint file, including it. An ErrCheckpointNotFound if the
+		// checkpoint file is not found in the directory.
+		FilesFromCheckpoint(string) ([]File, error)
```
giautm, line 78: The input is the full name of the file or just the version?

giautm (approved):

sql/migrate/dir.go
a8m: Currently, the file name.

PR #1998, opened 2023-08-17
Build arm64 mysql and mariadb Docker images

jacobwgillespie:
Closes #1997 

This PR builds `arm64` Docker images for mysql and mariadb, excluding the mysql image versions that don't have an upstream arm64 base image.

PR #2001, opened 2023-08-17
.github/ops/mysql: fix Dockerfile builds, fix mariadb:latest

jacobwgillespie:
Followup from #1998, this fixes two issues with the current workflow:

1. The workflow filename has changed, so builds aren't triggered on merge anymore
2. The `mariadb:latest` image does not have `mysqld`, so this adds a new build arg to override to `mariadbd` for that image build. This unblocks the regular cron builds.


giautm (approved):

giautm:
Thank you for your contribution. <3

PR #2004, opened 2023-08-17
Fix generation of initialized DB in Dockerfile

jacobwgillespie:
Followup to #2003.

I was able to get this to work through some trial and error at https://github.com/jacobwgillespie/atlas-dockerfile-repro. Rather than overriding the `datadir`, which for some reason isn't working as expected when building in GitHub Actions, this PR updates the Dockerfile to just run the server once, then copy `/var/lib/mysql` to the final image.

This is successfully building at the test repo above.

cc @giautm


giautm:
Thank you for your effort here. 😍😍😍

jacobwgillespie:
Awesome, I was able to run `atlas schema diff --dev-url "docker://mysql/8/schema"` from my Mac and it used the new `arm64` image, it's a _lot_ faster now 🙂

PR #2025, opened 2023-08-26
Fix cascade behavior when foreign keys with same name

remen:


rotemtam:
Sorry didn’t mean to unmark draft. Tried to click “run tests” with my chubby fingers

remen:
> Sorry didn’t mean to unmark draft. Tried to click “run tests” with my chubby fingers

No worries.

I updated the title & description with what I think this solves. I'm a bit unsure on how to take it from here though. Should I add a test case? Should I try to find out why it breaks for cockroach db?

Happy to continue the conversation on discord if that's better for you.

remen:
Latest commit should make it compatible with cockroachdb. Feel free to run tests again.

sql/postgres/inspect.go
```diff
@@ -1281,7 +1281,21 @@ SELECT
...
+					WHEN 'r' THEN 'RESTRICT'
+					WHEN 'a' THEN 'NO ACTION'
+				END AS update_rule,
+				CASE con.confdeltype
+					WHEN 'c' THEN 'CASCADE'
+					WHEN 'n' THEN 'SET NULL'
+					WHEN 'd' THEN 'SET DEFAULT'
+					WHEN 'r' THEN 'RESTRICT'
+ 					WHEN 'a' THEN 'NO ACTION'
+ 				END AS delete_rule
```
a8m, line 1298: I prefer to append the `con.confupdtype`,  `con.confdeltype` to the query and do this mapping in Go.

There are small test cases to change then in `inspect_test.go` and maybe add another one in `internal/integration/testdata/postgres`.

a8m (commented):
Thanks for the contribution, @remen. See my comment and let me know if that requires too much time from your side and I can take it.

sql/postgres/inspect.go
remen: Thanks for the feedback @a8m! I had a look at doing it this way, but the problem is that both MySQL and Postgres go through the same code path in `sqlx.SchemaFKs`, and I couldn't find a way to resolve this without more refactoring than I'm comfortable with.

I think handing it over to you is better at this point.

a8m:
Thanks for your contribution, @remen. Your commits were added to https://github.com/ariga/atlas/pull/2029.

PR #2027, opened 2023-08-26
feat/Table Inheritance (Mixins)

### LINKED ISSUE #1317: Table inheritance
Suppose I have some fields that I want to have in all my tables. I can declare those once and then I can inherit them in other tables.

```hcl
/// base table
table "base" {
  schema = schema.public
  column "created_at" {
    null    = false
    type    = timestamptz
    default = sql("now()")
  }
  column "updated_at" {
    null    = false
    type    = timestamptz
    default = sql("now()")
  }
}

/// Main table
table "projects" {
  schema = schema.public
 inherits = table.base
  column "id" {
    null = false
    type = bigserial
  }
}

```

Is this supported? This is a good feature.

patrykgz:
Fixes #1317. 

Mixins are implemented as a new block to the HCL schema, and when parsed are merged with the tables which embed them.

Example Schema:
```hcl
schema "a" {}

table "a" "users" {
  schema = schema.a
  embed = mixin.base
}

mixin "base" {
  column "id" {
    type = int
  }
  column "created_at" {
    type = timestamp
    default = now()
  }
}
```
Is exactly equivalent to:
```hcl
schema "a" {}

table "a" "users" {
  schema = schema.a
  column "id" {
    type = int
  }
  column "created_at" {
    type = timestamp
    default = now()
  }
}
```

---
DISCUSSION: with this change, we've introduced a HCL-only abstraction as Mixins are not represented in the database. Thereby if a user applied this diff, and then inspected the database the generated hcl will be different. Is this acceptable? Or should we store mixins in an atlas table (e.g. `atlas_mixins`), so on Inspect we can retrieve the user's configured mixins?

### DIFF
--- a/doc/md/atlas-schema/hcl.mdx
+++ b/doc/md/atlas-schema/hcl.mdx
@@ -207,6 +207,59 @@ table "b" "users" {
 }
 ```
 
+## Mixin
+
+In order to avoid code duplication, Atlas supports mixins. A mixin is a set of table properties that can be included in
+other tables. Mixins are defined using the `mixin` block, and they can be included in other tables using the `embed`
+statement.
+
+```hcl
+schema "a" {}
+
+table "a" "users" {
+  schema = schema.a
+  embed = mixin.base
+}
+
+mixin "base" {
+  column "id" {
+    type = int
+  }
+  column "created_at" {
+    type = timestamp
+    default = now()
+  }
+}
+```
+
+Multiple Mixins can be embedded in a table.
+
+```hcl
+schema "a" {}
+
+table "a" "users" {
+  schema = schema.a
+  embed = [mixin.base, mixin.deletable]
+}
+
+mixin "base" {
+  column "id" {
+    type = int
+  }
+  column "created_at" {
+    type = timestamp
+    default = now()
+  }
+}
+
+mixin "deletable" {
+  column "deleted_at" {
+    type = timestamp
+    null = true
+  }
+}
+```
+
 ## View
 
 A `view` is a virtual table in the database, defined by a statement that queries rows from one or more existing
--- a/internal/integration/hclsqlspec/hclsqlspec_test.go
+++ b/internal/integration/hclsqlspec/hclsqlspec_test.go
@@ -524,6 +524,294 @@ schema "c" {
 
 }
 
+func TestTable_Basic(t *testing.T) {
+	f := `schema "account_a" {
+}
+table "t1" {
+  schema = schema.account_a
+  column "id" {
+	type = int
+  }
+  column "name" {
+	type = text
+  }
+}
+`
+
+	for _, tt := range dialects {
+		t.Run(tt.name, func(t *testing.T) {
+			var r schema.Realm
+			err := tt.Eval([]byte(f), &r, nil)
+			require.NoError(t, err)
+			exp := &schema.Realm{
+				Schemas: []*schema.Schema{
+					{
+						Name:  "account_a",
+						Realm: &r,
+						Tables: []*schema.Table{
+							{
+								Name: "t1",
+								Columns: []*schema.Column{
+									{Name: "id", Type: &schema.ColumnType{Type: &schema.IntegerType{T: "int"}}},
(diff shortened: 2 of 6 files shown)


schemahcl/extension.go
```diff
@@ -376,11 +376,12 @@ func setPtr(field reflect.Value, cv cty.Value) error {
 // setSliceAttr sets the value of attr to the slice field. This function expects both the target field
 // and the source attr to be slices.
 func setSliceAttr(field reflect.Value, attr *Attr) error {
-	if !attr.V.Type().IsListType() && !attr.V.Type().IsTupleType() {
-		return fmt.Errorf("schemahcl: field is of type slice but attr %q is type: %s", attr.K, attr.V.Type().FriendlyName())
```
MutatePat, line 380: This is the only change I am concerned about - to me it seems logical this accepts either a value or an iterable, and if it's a single value it is converted into an iterable of one.  However it is a major change!

sql/internal/specutil/convert.go
```diff
@@ -191,6 +213,56 @@ func Table(spec *sqlspec.Table, parent *schema.Schema, convertColumn ConvertTabl
 	return tbl, nil
 }
 
+// Mixin applies a sqlspec.Mixin to a schema.Table.
+func Mixin(spec *sqlspec.Mixin, target *schema.Table, convertColumn ConvertTableColumnFunc,
```
MutatePat, line 217: Any advice on how we can reuse some logic from Table?

sql/sqlspec/sqlspec.go
```diff
@@ -19,14 +19,26 @@ type (
...
+	}
+
+	// Mixin holds a specification for common table attributes.
+	Mixin struct {
+		Name        string        `spec:",name"`
+		Columns     []*Column     `spec:"column"`
+		PrimaryKey  *PrimaryKey   `spec:"primary_key"`
+		ForeignKeys []*ForeignKey `spec:"foreign_key"`
+		Indexes     []*Index      `spec:"index"`
+		Checks      []*Check      `spec:"check"`
```
MutatePat, line 41: Should we create a struct with the  shared fields (cols, keys, index, etc) that is embedded in Table and Mixin?

internal/integration/hclsqlspec/hclsqlspec_test.go
```diff
@@ -524,6 +524,323 @@ schema "c" {
 
 }
 
+func TestTable_Basic(t *testing.T) {
```
MutatePat, line 527: is this the best plase for these tests?

doc/md/atlas-schema/hcl.mdx
MutatePat, line 1: is this enough  docs?

MutatePat:
I can't run the full regression suite locally, only the postgres test I created which passed - keen to get the CI approved so I can validate this doesn't break anything!

a8m:
Hey @MutatePat, thanks a lot for your contribution 🙏

I'd prefer for this logic to be implemented only in the `schemahcl` package without the other packages being aware of it. A `mixin` can be used by any block, including mixins themselves. As we're set to release our official LSP for both IntelliJ and VSCode this week, I don't see a way to support this feature without affecting the type-safety; it requires more design and research from our end (cc @giautm). We didn't spend time on this feature because it was mostly "nice to have" improvement without a critical need.

Thanks again for spending time on this and contributing to Atlas.

MutatePat:
> I'd prefer for this logic to be implemented only in the schemahcl package without the other packages being aware of it. A mixin can be used by any block, including mixins themselves.

Understood I'll make the change - regarding type safety, is this PR blocked until we assess LSP support?

a8m:
Yes. I believe a "mixin schema" should be somehow defined in order to provide both type-safety and editor support for it. As commented in https://github.com/ariga/atlas/issues/1317#issuecomment-1339151355, I think this can be useful for defining any reusable blocks and expanding them into other blocks. Maybe type-safe macros are the way to go, not sure - this requires more research and attention.

MutatePat:
Please re-review, the new solution is more generic.

Regarding the LSP - is this available on a feature branch in Atlas CE yet, or is this a private feature? If we must solve type safety before this can get merged, I wouldn't want this PR to go stale

PR #2030, opened 2023-08-28
sql/internal/sqlx: consistent order from topsort

a8m:
Although both (t1,t2)/(t2,t1) are valid orders when t1 and t2 do not depend each other, we want topsort to return consistent order by fallback to their names

### DIFF
--- a/sql/internal/sqlx/plan.go
+++ b/sql/internal/sqlx/plan.go
@@ -204,8 +204,8 @@ func sortMap(changes []schema.Change) (map[string]int, error) {
 		sorted[name] = len(sorted)
 		return false
 	}
-	for node := range deps {
-		if visit(node) {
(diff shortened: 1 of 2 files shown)


rotemtam (approved):

sql/internal/sqlx/plan.go
```diff
@@ -359,3 +359,24 @@ func CheckChangesScope(opts migrate.PlanOptions, changes []schema.Change) error
 	}
 	return nil
 }
+
+// byKeys sorts a map by keys.
+func byKeys[T any](m map[string]T) []struct {
```
masseelch, line 364:
```suggestion
func byStringKeys[T any](m map[string]T) []struct {
```
or
```suggestion
func byKeys[K comparable, V any](m map[K]V) []struct {
```

PR #2032, opened 2023-08-29
cmd/ext: create gcp_cloudsql_token datasource

giautm:


cmd/atlas/internal/cmdext/cmdext.go
```diff
@@ -31,6 +32,10 @@ import (
 	"ariga.io/atlas/sql/sqltool"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/feature/rds/auth"
+	"golang.org/x/exp/slices"
```
a8m, line 35: We can upgrade go to latest version in this module and use stdlib instead.

a8m (approved):

rotemtam (approved):

PR #2038, opened 2023-09-01
adding a guide on deloying to kubernetes with the atlas operator and flux-cd

jmushiri:


rotemtam:
This will land a bit modified in https://github.com/ariga/atlas/pull/2092

PR #2050, opened 2023-09-06
cmd/atlas: support `build-tags` for entloader

giautm:


cmd/atlas/internal/cmdext/cmdext.go
```diff
@@ -810,7 +810,17 @@ func (EntLoader) tables(u *url.URL) ([]*entschema.Table, error) {
 	if err != nil {
 		return nil, err
 	}
-	graph, err := entc.LoadGraph(abs, &gen.Config{})
+	opts := []entc.Option{}
+	if tags, ok := u.Query()["build-tags"]; ok {
```
a8m, line 814: I would prefer go-tags or just tags

a8m (approved):

cmd/atlas/internal/cmdext/cmdext.go
giautm: let me check other tools

PR #2108, opened 2023-09-21
cmd/atlas: omit url strings from parse errors

a8m:
Before this PR, passing Atlas an invalid URL returned an error that includes the problematic URL (in `url.Error.URL`):

```shell
$ atlas schema inspect -u "  mysql://user:pass@:3306"

Error: parse "  mysql://user:pass@:3306": first path segment in URL cannot contain colon
```

This might be useful in local development, but it should be avoided, as Atlas is mostly used in CI/CD environments. This PR changes the behavior by introducing a new function (`sqlclient.ParseURL`) that returns parse errors without the raw URLs, and it is used by the various cmd packages. For example:

```
$ atlas schema inspect -u "  mysql://user:pass@:3306"

Error: first path segment in URL cannot contain colon
```

### DIFF
--- a/cmd/atlas/internal/cloudapi/client.go
+++ b/cmd/atlas/internal/cloudapi/client.go
@@ -12,12 +12,12 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"net/url"
 	"runtime"
 	"strings"
 	"time"
 
 	"ariga.io/atlas/sql/migrate"
+	"ariga.io/atlas/sql/sqlclient"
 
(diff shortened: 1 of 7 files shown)


giautm (approved):

giautm:
I think we should add golint rule to block url.Parse for all places

a8m:
> I think we should add golint rule to block url.Parse for all places

It’s only for database URLs. For other URLs, like directory path, we use url.Parse and return the full error.

PR #2109, opened 2023-09-22
sql: add support for materialized views in sqlspec

a8m:
Adding another type to HCL, `materialized`. The reason we don't use `view` is that, in different databases (PostgreSQL, MSSQL, DB2, etc.), these two don't have the same set of attributes and we want our LSP to enforce type safety without executing custom logic.

### DIFF
--- a/sql/internal/specutil/convert.go
+++ b/sql/internal/specutil/convert.go
@@ -41,9 +41,10 @@ type (
 type (
 	// ScanDoc represents a scanned HCL document.
 	ScanDoc struct {
-		Schemas []*sqlspec.Schema
-		Tables  []*sqlspec.Table
-		Views   []*sqlspec.View
+		Schemas      []*sqlspec.Schema
+		Tables       []*sqlspec.Table
+		Views        []*sqlspec.View
+		Materialized []*sqlspec.View
 	}
 	// ScanFuncs represents a set of scan functions
 	// used to convert the HCL document to the Realm.
@@ -51,6 +52,14 @@ type (
 		Table ConvertTableFunc
 		View  ConvertViewFunc
 	}
+)
+
+const (
+	typeView         = "view"
+	typeTable        = "table"
+	typeColumn       = "column"
+	typeSchema       = "schema"
+	typeMaterialized = "materialized"
 )
 
 // Scan populates the Realm from the schemas and table specs.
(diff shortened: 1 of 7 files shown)


sql/internal/specutil/convert.go
```diff
@@ -53,6 +54,14 @@ type (
 	}
 )
 
+const (
+	typeView         = "view"
+	typeTable        = "table"
+	typeColumn       = "column"
+	typeSchema       = "schema"
+	typeMaterialized = "materialized"
```
masseelch, line 62:
```suggestion
	typeMaterialized = "materialized view"
```

Think this is more readable / correct.

sql/internal/specutil/convert.go
```diff
@@ -284,9 +331,9 @@ func Index(spec *sqlspec.Index, parent *schema.Table, partFns ...func(*sqlspec.I
 			part := &schema.IndexPart{SeqNo: i, Desc: p.Desc}
 			switch {
 			case p.Column == nil && p.Expr == "":
-				return nil, fmt.Errorf(`"column" or "expr" are required for index %q at position %d`, spec.Name, i)
+				return nil, fmt.Errorf(`typeColumn or "expr" are required for index %q at position %d`, spec.Name, i)
```
masseelch, line 334: ?

masseelch: Think you forgot to %s this.

sql/internal/specutil/convert.go
```diff
@@ -284,9 +331,9 @@ func Index(spec *sqlspec.Index, parent *schema.Table, partFns ...func(*sqlspec.I
...
 			case p.Column != nil && p.Expr != "":
-				return nil, fmt.Errorf(`cannot use both "column" and "expr" in index %q at position %d`, spec.Name, i)
+				return nil, fmt.Errorf(`cannot use both typeColumn and "expr" in index %q at position %d`, spec.Name, i)
```
masseelch, line 336: ditto

sql/schema/dsl.go
```diff
@@ -231,6 +231,12 @@ func NewView(name, def string) *View {
 	return &View{Name: name, Def: def}
 }
 
+// NewMaterialized creates a new materialized View.
+func NewMaterialized(name, def string) *View {
+	return NewView(name, def).
+		SetMaterialized(true)
```
masseelch, line 237:
```suggestion
	return NewView(name, def).SetMaterialized(true)
```
Not really neede, just a suggestion.

sql/schema/dsl.go
```diff
@@ -231,6 +231,12 @@ func NewView(name, def string) *View {
 	return &View{Name: name, Def: def}
 }
 
+// NewMaterialized creates a new materialized View.
+func NewMaterialized(name, def string) *View {
```
masseelch, line 235:
```suggestion
func NewMaterializedView(name, def string) *View {
```
Keeps it more closed to SQL imho.

masseelch (approved):

PR #2160, opened 2023-10-05
Add support for postgis in --dev-url

anthonybishopric:
Postgis is a commonly used GIS extension for Postgres and ships with its own published Docker images. Everything about Atlas works except for image name parsing for the Postgis images. This PR adds support for Postgis in the URL scheme.


a8m (approved):
Thanks for the contribution, @anthonybishopric.

PR #2219, opened 2023-10-25
cmd/atlas: init stubs

rotemtam:


a8m (approved):
I'd use context.WithTimeout instead and merge checkForUpdate into one function and split it to oss and ent

PR #2240, opened 2023-10-31
support autoIncColumnChange

feitianbubu:


a8m:
Thanks for the contribution, @feitianbubu.

The AUTO_INCREMENT clause is a table attribute (not a column attribute) and it’s already implemented here: https://github.com/ariga/atlas/blob/master/sql/mysql/diff.go#L394. Do you have any issues with it?

feitianbubu:
Thank you for your reply. @a8m 
The `autoIncChange` func only effect where the autoincrement value is modified. However, when the user adds or removes the autoincrement attribute on a field, it does not take effect. The `autoIncColumnChange` I added will solves this issue.

example, when user change a `AUTO_INCREMENT` on id filed
`CREATE TABLE `account` (
  `id` bigint NOT NULL AUTO_INCREMENT,  // will not be change
  PRIMARY KEY (`id`)
) AUTO_INCREMENT=2;` //  will change by autoIncChange

a8m:
Thanks for the response, @feitianbubu. Can you please describe the use case with "before" and "after" schemas? Do you try to add it to an existing column? Note that adding the AUTO_INCREMENT to a table without adding it to a column is ignored.

feitianbubu:
Thanks for the response, @a8m ,
This is  a example:
1. before hcl:  
```
table "account" {
  schema = schema.test
  column "id" {
    null = false
    type = bigint
  }
  primary_key {
    columns = [column.id]
  }
}
```
2. after hcl:
```
table "account" {
  schema = schema.test
  column "id" {
    null           = false
    type           = bigint
    auto_increment = true
  }
  primary_key {
    columns = [column.id]
  }
}
```
3. current diff result:  Schemas are synced, no changes to be made.
4. need result: 
```
-- Modify "account" table
ALTER TABLE `account` MODIFY COLUMN `id` bigint NOT NULL AUTO_INCREMENT;
```

feitianbubu:
commit pr will gen:
```
-- Modify "account" table
ALTER TABLE `account` MODIFY COLUMN `id` bigint NOT NULL AUTO_INCREMENT;
```

Issue #1678, opened 2023-05-22
supporting custom docker images in --dev-url


krilor:
Hi! And thanks for making this cool tool!

I'm trying to use Atlas on databases where we need to have some extensions enabled and am really excited that #1601 is happening.

While using a [dev database](https://atlasgo.io/concepts/dev-database), it seems like we have to use a "vanilla" postgres docker image, tho. I'm working around that by spinning up a ephemeral database "on the side", and using the standard `postgres://` url.

I'm reading https://github.com/ariga/atlas/blob/master/cmd/atlas/internal/docker/docker.go#L72
and thinking about how to support this within the current docker url format (docker://image/tag)

The only suggestion I can come up with is something similar to the "git+ssh://" format that I sometimes see with git (ref https://github.com/IonicaBizau/git-url-parse/blob/master/test/index.js#L12 ). So say that I compile an extension on top of the official postgres image (that preserves all the env vars, config etc that Atlas requires), then I could do: `docker+postgres://<my-image>/<my-tag>`

Would there be any interest in this or are there other options that would solve the same need?

a8m:
@krilor, thanks for reporting this, and sorry for the delayed response. The issue was resolved with #2249

Please, see https://atlasgo.io/concepts/dev-database#introduction for more info.

krilor:
Very nice :) Thanks for making it happen

PR #2249, opened 2023-11-03
cmd/atlas/internal/docker: allow custom images in docker driver

a8m:
Fixed https://github.com/ariga/atlas/issues/1678
- Support custom docker images
- Remove the implicit "docker pull" 

For example:

```shell
$ atlas schema inspect \
  -u file://schema.sql \
  --dev-url "docker+postgres://ankane/pgvector/dev"

table "t" {
  schema = schema.public
  column "c" {
    null = true
    type = integer
  }
}
schema "public" {
  comment = "standard public schema"
}
```
```shell
$ atlas schema inspect \
  -u file://schema.sql \
  --dev-url "docker+postgres://ankane/pgvector/dev?search_path=public" \
  --format="{{ sql . \"  \" }}"

-- Create "t" table
CREATE TABLE "t" (
  "c" integer NULL
);
```

### DIFF
--- a/cmd/atlas/internal/docker/docker_test.go
+++ b/cmd/atlas/internal/docker/docker_test.go
@@ -58,10 +58,11 @@ func TestFromURL(t *testing.T) {
 	cfg, err := FromURL(u)
 	require.NoError(t, err)
 	require.Equal(t, &Config{
-		Image: "arigaio/mysql",
-		Env:   []string{"MYSQL_ROOT_PASSWORD=pass"},
-		Port:  "3306",
-		Out:   io.Discard,
+		driver: "mysql",
+		Image:  "arigaio/mysql",
+		Env:    []string{"MYSQL_ROOT_PASSWORD=pass"},
+		Port:   "3306",
+		Out:    io.Discard,
 	}, cfg)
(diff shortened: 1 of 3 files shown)


cmd/atlas/internal/docker/docker_test.go
```diff
@@ -105,10 +109,101 @@ func TestFromURL(t *testing.T) {
...
 		Database: "postgres",
 		Env:      []string{"POSTGRES_PASSWORD=pass"},
 		Port:     "5432",
 		Out:      io.Discard,
 	}, cfg)
 }
+
+func TestFromURL_CustomImage(t *testing.T) {
+	for _, tt := range []struct {
+		url, image, db string
```
masseelch, line 123:
```suggestion
		url, image, db, dlct string
```

masseelch (approved):

cmd/atlas/internal/docker/docker_test.go
masseelch: Prefer to check for the correct driver chosen as well here. Just a minor suggestion, as the testscript is there as well.

PR #2260, opened 2023-11-06
Fixes a bug when rendering multi-word types in Mermaid ERDs

jerluc:
# Overview

When using the `{{ mermaid . }}` template with a simple table schema containing a data type that has multiple words, I noticed that the generated Mermaid syntax was broken.

For example, given the MRE `test.sql`:

```sql
-- test.sql --
CREATE TABLE "users" (
  oops DOUBLE PRECISION
)
```

After running:

```shell
atlas schema inspect --url file://test.sql --dev-url docker://postgres/latest/dev --format="{{ mermaid . }}"
```

We can see that it outputs:

```mermaid
erDiagram
    users {
      double precision oops
    }
```

This breaks the Mermaid syntax, as type names cannot contain whitespace.

# Proposed fix

I've come up with a pretty basic fix for this issue by simply replacing all whitespace characters with an underscore (`_`). I've also added a unit test that breaks before this bugfix to prove the solution at least works in this scenario.

Re-running the same `atlas schema inspect` command as above using this new build produces:

```mermaid
erDiagram
    users {
      double_precision oops
    }
```

### DIFF
--- a/cmd/atlas/internal/cmdlog/cmdlog.go
+++ b/cmd/atlas/internal/cmdlog/cmdlog.go
@@ -318,7 +318,7 @@ Migrating to version {{ cyan .Target }}{{ with .Current }} from {{ cyan . }}{{ e
   {{ yellow "--" }} {{ dec $files }} migrations ok (1 with errors)
   {{ yellow "--" }} {{ dec $stmts }} sql statements ok (1 with errors)
 {{- else }}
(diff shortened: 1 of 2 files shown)


a8m:
Thanks for the fix, @jerluc. The commit was added with https://github.com/ariga/atlas/pull/2261.

PR #2287, opened 2023-11-15
sql/migrate: support controlling the execution order

a8m:
Also, changing the default behavior to error (instead of skip) in case non-linear history

(cmd tests will be pushed soon)


cmd/atlas/internal/cmdapi/migrate_test.go
```diff
@@ -713,6 +713,95 @@ func TestMigrate_ApplyTxModeDirective(t *testing.T) {
 	require.Equal(t, s, `unknown txmode "unknown" found in file directive "20220925094021_second.sql"`)
 }
 
+func TestMigrate_ApplyExecOrder(t *testing.T) {
```
masseelch, line 716: Add test that checks the new default is applied.

a8m: There's one for this (L750~).

masseelch (approved):

PR #2288, opened 2023-11-17
cmd/atlas/internal/cmdlog: indent new line in apply log

masseelch:
Previously, if applying migration files which statements that contained new lines, it looked like this:
image

After this change it is properly indented:
image


cmd/atlas/internal/cmdlog/cmdlog.go
```diff
@@ -291,6 +291,7 @@ var (
 		"upper":      strings.ToUpper,
 		"json":       jsonEncode,
 		"json_merge": jsonMerge,
+		"indentLn":   indentLn,
```
giautm, line 294: why not `indent_ln`? to consistent with `json_merge`

giautm (approved):

PR #2306, opened 2023-11-22
cmd/atlas: allow extending config hcl parse options

a8m:


doc/md/versioned/diff.mdx
```diff
@@ -854,8 +854,8 @@ env "local" {
   diff {
     // By default, indexes are not created or dropped concurrently.
     concurrent_index {
-      create = true
-      drop   = true
+      add  = true
```
masseelch, line 857: This is not backwards compatible, is it?

masseelch (approved):

PR #2336, opened 2023-12-05
cmd/atlas: allow run sqlserver docker

giautm:


cmd/atlas/internal/docker/docker.go
```diff
@@ -95,8 +100,7 @@ func FromURL(u *url.URL) (*Config, error) {
 	case n == 2:
 		parts[0] = fmt.Sprintf("%s/%s", parts[0], parts[1])
 	}
-	switch u.Scheme {
-	case "docker+postgres", "docker+mysql", "docker+maria":
+	if slices.Contains(flavours, u.Scheme) {
```
a8m, line 103: It feels less safe. Maybe add `docker+sqlserver` instead?

cmd/atlas/internal/docker/docker.go
```diff
@@ -131,6 +135,19 @@ func FromURL(u *url.URL) (*Config, error) {
...
 		cfg, err = PostgreSQL(tag, opts...)
+	case DriverMSSQL:
+		if len(parts) > 1 {
+			if db := parts[1]; db != "master" {
+				opts = append(opts, Setup(fmt.Sprintf("CREATE DATABASE [%s]", db)))
+			}
+		}
+		if tag == "" {
+			tag = "2022-latest"
+		}
```
a8m, line 146: Th tag is not required, you can still call `docker pull mcr.microsoft.com/mssql/server` no?

cmd/atlas/internal/docker/docker.go
```diff
@@ -73,6 +77,7 @@ const (
 	DriverMySQL    = "mysql"
 	DriverMariaDB  = "mariadb"
 	DriverPostgres = "postgres"
+	DriverMSSQL    = "sqlserver"
```
a8m, line 80: qq, why not name it as the docker image (mssql)? This is also a valid flavour

PR #2347, opened 2023-12-07
internal/ci: run dialect integration test only after general ci finished

ronenlu:


.github/workflows/ci-dialect_oss.yaml
```diff
@@ -3,18 +3,9 @@
...
-  push:
-    branches:
-      - master
-    paths-ignore:
-      - 'doc/**'
-      - 'ops/**'
+  workflow_call:
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}-dialect
```
masseelch, line 8: This is not good. It means that if dialect is running and you push a new commit, the dialect tests will not be stopped until the new one starts.

Worst case, they run through if the unit tests fail. Correct?

masseelch (approved):

PR #2349, opened 2023-12-08
schemahcl: reduce eval mem usage

a8m:
```
benchmark              old ns/op     new ns/op     delta
BenchmarkSchema-16     569319394     238469401     -58.11%

benchmark              old allocs     new allocs     delta
BenchmarkSchema-16     1327927        1174085        -11.59%

benchmark              old bytes     new bytes     delta
BenchmarkSchema-16     293956792     186829161     -36.44%
```

### DIFF
--- a/schemahcl/context.go
+++ b/schemahcl/context.go
@@ -282,11 +282,18 @@ func setBlockVars(ctx *hcl.EvalContext, b *hclsyntax.Body) (*hcl.EvalContext, er
 	if err != nil {
 		return nil, err
 	}
-	if ctx.Variables == nil {
-		ctx.Variables = make(map[string]cty.Value)
-	}
-	for k, v := range vars {
-		ctx.Variables[k] = v
+	switch {
+	case ctx.Variables == nil:
+		ctx.Variables = vars
+	case len(ctx.Variables) >= len(vars):
+		for k, v := range vars {
+			ctx.Variables[k] = v
+		}
+	default:
+		for k, v := range ctx.Variables {
+			vars[k] = v
+		}
+		ctx.Variables = vars
 	}
 	return ctx, nil
 }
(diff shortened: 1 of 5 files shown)


schemahcl/context.go
```diff
@@ -402,13 +409,11 @@ func blocksOfType(blocks hclsyntax.Blocks, typeName string) []*hclsyntax.Block {
...
 func attrMap(attrs hclsyntax.Attributes) map[string]cty.Value {
-	out := make(map[string]cty.Value)
+	out := make(map[string]cty.Value, len(attrs))
 	for _, v := range attrs {
-		value, diag := v.Expr.Value(nil)
-		if diag.HasErrors() {
-			continue
+		if value, diag := v.Expr.Value(nil); !diag.HasErrors() {
+			out[v.Name] = value
 		}
```
masseelch, line 416: What kind of perf impact has this?

schemahcl/extension.go
```diff
@@ -547,10 +547,16 @@ func scanAttr(key string, r *Resource, field reflect.Value) error {
 	return nil
 }
 
+// specCache holds a cache for type-spec fields descriptions.
+var specCache sync.Map
+
```
masseelch, line 552: Is this called concurrently? A simple map is faster.

schemahcl/schemahcl.go
```diff
@@ -479,28 +480,11 @@ func (s *State) mayScopeContext(ctx *hcl.EvalContext, scope []string) *hcl.EvalC
...
-	// Override the parent context with the scoped variables and functions.
-	for n, v := range vars {
-		nctx.Variables[n] = v
-	}
-	for n, f := range funcs {
-		nctx.Functions[n] = f
-	}
-	// A patch from the past. Should be moved
-	// to specific scopes in the future.
-	nctx.Functions["sql"] = rawExprFunc
```
masseelch, line 502: Where is this? No longer needed since we don't copy to a new map?

sql/postgres/sqlspec.go
```diff
@@ -900,7 +900,6 @@ var TypeRegistry = schemahcl.NewRegistry(
 		schemahcl.NewTypeSpec(TypeDateRange),
 		schemahcl.NewTypeSpec(TypeDateMultiRange),
 		schemahcl.NewTypeSpec("hstore"),
-		schemahcl.NewTypeSpec("sql", schemahcl.WithAttributes(&schemahcl.TypeAttr{Name: "def", Required: true, Kind: reflect.String})),
```
masseelch, line 903: Why remove?

schemahcl/context.go
a8m: no perf. just avoiding the unnecessary ifs.

schemahcl/schemahcl.go
a8m: global definitions (see stdFuncs).

schemahcl/extension.go
a8m: Yes - our tests :D

PR #2359, opened 2023-12-11
sql: implement migrate.StmtScanner in drivers

a8m:


sql/migrate/lex.go
```diff
@@ -54,9 +54,40 @@ func Stmts(input string) ([]*Stmt, error) {
 	}).Scan(input)
 }
 
+// FileStmtDecls scans atlas-format file statements using
+// the Driver implementation, if implemented.
+func FileStmtDecls(drv Driver, f File) ([]*Stmt, error) {
```
masseelch, line 59: Not sure I like the naming. I see `StmtDecls` is taken, but why `File`? We can extend the `StmtDecls` with a spread and use options like `Driver()` maybe?

sql/mysql/mysqlcheck/mysqlcheck.go
```diff
@@ -142,12 +142,16 @@ func addNotNull(p *datadepend.ColumnPass) (diags []sqlcheck.Diagnostic, err erro
...
 	})
 	if !ok {
 		return false
 	}
-	filled, _ := p.ColumnFilledAfter(pass.File, pass.Table, pass.Column, pass.Change.Stmt.Pos, matchValue)
+	stmts, err := migrate.FileStmtDecls(pass.Dev, pass.File)
+	if err != nil {
+		return false
+	}
+	filled, _ := p.ColumnFilledAfter(stmts, pass.Table, pass.Column, pass.Change.Stmt.Pos, matchValue)
```
masseelch, line 154: Ignore error?

rotemtam (approved):

sql/migrate/lex.go
a8m: There's a method under `File` named `StmtDecls`. This is the same as calling File.StmtDecls, but accepts Driver as the first argument. Feels simpler and clear enough to me. Especially if I want to have both Stmts and StmtDecls.

sql/mysql/mysqlcheck/mysqlcheck.go
a8m: Yes, parsing can fail and we ignore this "enrichment" in that case (that logic was not changed).

PR #2426, opened 2024-01-08
doc: add computed triggers to hcl examples

a8m:


cemo:
> for_each = toset([table.users, table.orders, table.payments])

This part is not working for us. Only one table is supported for now.

a8m:
What version do you use?

PR #2447, opened 2024-01-16
Support for PgBouncer transaction pooling 

vgogolev1703:
We've had some issues using atlasgo with PgBouncer, which are caused by the fact that our PgBouncer runs in transactional mode, which is widely used in production (as we do in our production).

Sometimes, after successful migration, atlasgo can't release advisory lock. We migrate our schema using the same connection that the application uses (throught pgbouncer)

> `Error: sql/postgres: failed releasing lock 2686470616`

We found out, that atlasgo using session-level advisory locks, which is not correctly working with transactional mode.. 

To prevent such behavior, atlasgo can use pg_advisory_xact_lock (which is transactional) with some bonus - you don't need to release it, cause it unlocks right after transaction ends. Also, it should work correctly with PgBouncer in session mode and without PgBouncer at all.
https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS

Here's the same problem that flyway had a few years ago, which was resolved by using advisory locks in transactional mode by default. 
https://github.com/flyway/flyway/issues/2895

### DIFF
--- a/sql/postgres/driver.go
+++ b/sql/postgres/driver.go
@@ -164,16 +164,7 @@ func (d *Driver) Lock(ctx context.Context, name string, timeout time.Duration) (
 	}
 	return func() error {
 		defer conn.Close()
-		rows, err := conn.QueryContext(ctx, "SELECT pg_advisory_unlock($1)", id)
-		if err != nil {
-			return err
-		}
-		switch released, err := sqlx.ScanNullBool(rows); {
-		case err != nil:
-			return err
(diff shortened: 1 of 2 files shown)


a8m:
Thanks for the contribution, @vgogolev1703. Unfortunately, I cannot easily merge this because locks in Atlas are taken in multiple places, including both by sessions/connections and transactions. Therefore, the code there should be more dynamic.

I will try to give it a look in the upcoming days.

Issue #2455, opened 2024-01-17
Statement-level granularity: Infinitely pending migration


ruslic19:
As described in [atlas documentation here](https://atlasgo.io/blog/2023/04/10/troubleshooting-migrations#statement-level-granularity), migrations can partially fail and in provided there case in order to fix it, we add one more statement to our migration and then rerun it again.

**Expected:**
There is `atlas migrate status` which returns amount of pending migrations and when let's say our last migration partially fails, status command returns us `pending: 1` which is supposed to become `0` when we will successfully fix and execute our partially failed migration.

**Actual:**
When our migration partially fails, we get `pending: 1` as expected, but then when we fix it and execute it again, even if it executes successfully, `pending` remains to be `1`, because of this, we can now execute our migrations multiple times and it never tells us that there is no migrations to execute anymore.

**Steps to reproduce:**
In general you can even use guide from statement level granularity itself or you can do it like this
- Declaratively define `users` table schema with nullable name
- Generate migration from defined schema and execute it
- Insert couple of rows into `users` table with name being `null`
- Change `users` table schema to have name `NOT NULL`
- Generate migration from updated schema and try to execute it
- Migration will be marked as partially failed, because table already contains `null` values
- Manually edit migration to add sql statement to update all nulls with some default value before altering table
- Recalculate hash of the migrations directory
- Execute migration again, now it succeeds, but migration will still be marked as `pending`
- Now you can try to get migration status or execute migrations again and it will always say that there is 1 pending migration
- `atlas_schema_revisions` table now has `total = 1` and `applied = 2` and this is the cause of the issue

**Reason:**
As I understand it is related to the fact that we determine if migration is pending by comparing applied statements to the total statements in a given migration. Migration is considered to be complete only if amount of applied statements strictly equals to amount of total statements in the migration. For example, [here in the code](https://github.com/ariga/atlas/blob/d9bae0967b1845623c52cb54c8d38c1e1b439783/sql/migrate/migrate.go#L738-L740). 

Whereas, in this case, when we update our migration to add new statement and execute it now successfully, row in the table `atlas_schema_revisions` responsible for this migration now contains `total=value_before_we_added_new_statement` and `applied=all_executed_statements_including_new_one`, so it results in something like this: `total=1` `applied=2`.

[...]

PR #2456, opened 2024-01-17
fix: update total statements in the migration when it is being executed

ruslic19:
The reason for this is that users are required to edit existing migration files to resolve partial failure issues. Without keeping total in sync, such migration would be considered still pending, because total != applied. 

This fix was NOT tested in all possible use cases, but only in the one described in the linked issue and it resolves it.

Issue: #2455

### DIFF
--- a/sql/migrate/migrate.go
+++ b/sql/migrate/migrate.go
@@ -812,6 +812,12 @@ func (e *Executor) Execute(ctx context.Context, m File) (err error) {
 			Hash:        hash,
 		}
 	}
+
+	// Update total statements in the existing revision to support
+	// statement-level granularity in cases when users add new statements to
+	// existing migrations to resolve partial failures.
+	r.Total = len(stmts)
+
 	// Save once to mark as started in the database.
 	if err = e.writeRevision(ctx, r); err != nil {
 		return err


a8m:
Thanks for opening this PR, @ruslic19. It requires additional work because in cases like the one mentioned in the issue, the "partial hashes" (stored in the revisions table) need to be updated as well.

I suggest using `migrate set` for now to unblock yourself, and we will provide updates on this matter - we encountered this in the past, and have some ideas how to handled this.

ruslic19:
Hi again, @a8m, I just had on my opinion more serious issue than previous one. My partially failed migration was marked as successful even though it didn't successfully end. The reason is still the same. 

Steps to reproduce:
- Declaratively define `users` table schema with nullable `name` and `age`
- Generate migration from defined schema and execute it
- Insert couple of rows into `users` table with `name` and `age` being `null`
- Change `users` table schema to have `name` and `age` `NOT NULL`
- Generate migration from updated schema and try to execute it
- Migration will be marked as partially failed, because table already contains `null` values
- Manually edit migration, but add only one query to update values for `name` to some default one
- Recalculate hash of the migrations directory
- Execute migration again, it fails again, because `age` still contains `null` values, but at this point one statement to update `name` values succeeded
- Now, revisions's `applied` statements equal to `1` and `total` statements are also `1`
- So, migration is now marked as successful, even though it didn't actually complete

So, because of this I decided to take a look on what you said and it seems to me that partial hashes are actually getting updated already, take a look [here in the code](https://github.com/ariga/atlas/blob/d9bae0967b1845623c52cb54c8d38c1e1b439783/sql/migrate/migrate.go#L845-L847).

Also checked it manually by inspecting `atlas_schema_revisions` and every time I execute migration with additional successful query to fix partial failure, new partial hash is getting added to the `atlas_schema_revisions`.

As I understand, this should be good enough to only add new partial hashes, because previous query should not be edited anyway, because partially failed migration will anyway skip successful queries during next execution, so only queries added after successful one matter.

I can be wrong here of course, so, sorry for bothering in such case.

PR #2499, opened 2024-02-02
sql/sqlite: Fix DROP TABLE IF EXISTS

eric:
I found that the syntax for the `DROP TABLE IF EXISTS` is not correct for sqlite.

https://www.sqlite.org/lang_droptable.html


giautm:
Thank @eric for this fix.

giautm (approved):

PR #2505, opened 2024-02-02
sq/postgres: normalize realm after convert

giautm:


sql/postgres/driver_oss.go
```diff
@@ -221,6 +221,10 @@ func convertSequences(_ []*sqlspec.Table, seqs []*sequence, _ *schema.Realm) err
 	return nil
 }
 
+func normalizeRealm(_ *schema.Realm) error {
+	return nil
+}
```
a8m, line 226:
```suggestion
func normalizeRealm(*schema.Realm) error {
	return nil
}
```

PR #2515, opened 2024-02-06
sql/sqlite: add stubs for verifying changes with context

noamcattan:


sql/sqlite/driver_oss.go
```diff
@@ -58,3 +58,7 @@ func (*state) addTrigger(*schema.AddTrigger) error {
 func (*state) dropTrigger(*schema.DropTrigger) error {
 	return nil // unimplemented.
 }
+
+func verifyChanges(context.Context, []schema.Change) error {
+	return nil
```
a8m, line 63:
```suggestion
	return nil // unimplemented.
```

a8m (approved):

PR #2549, opened 2024-02-17
sql/sqlx: drop the table before drop UDTs

giautm:
When table's triggers referenced to other types, it should drop first then types.


a8m (approved):
💯

PR #2556, opened 2024-02-19
all: upgrade ent to latest version

giautm:


cmd/atlas/tools.go
```diff
@@ -0,0 +1,5 @@
+package main
+
+import (
+	_ "entgo.io/ent/entc"
```
giautm, line 4: Without this, `go mod tidy` always remove `x/tools` package

cmd/atlas/internal/migrate/ent/generate.go
```diff
@@ -4,4 +4,4 @@
 
 package ent
 
-//go:generate go run entc.go
```
giautm, line 7: cc @masseelch

a8m (approved):

PR #2573, opened 2024-02-24
sql/specutil: custom object reference for `depends_on`

giautm:


a8m (approved):

giautm:
We still need another method for `fromDependsOn()` works

sql/internal/specutil/convert.go
```diff
@@ -71,6 +73,14 @@ type (
 		Func  func(*schema.Func) (*sqlspec.Func, error)
 		Proc  func(*schema.Proc) (*sqlspec.Func, error)
 	}
+	// RefNamer is an interface for objects that can
+	// return their reference and type name.
+	RefNamer interface {
+		// Ref returns the reference to the object.
+		Ref() *schemahcl.Ref
+		// TypeName returns the type and name of the object.
+		TypeName() (string, string)
```
a8m, line 82: I think better to be explicit that these are spec and type names. So either SpecTypeName, or SpecType+SpecName, or return specutil.TypeName instead.

sql/internal/specutil/convert.go
```diff
@@ -71,6 +73,20 @@ type (
...
+		Ref() *schemahcl.Ref
+	}
+	// ObjectSpec is an interface for objects that can
+	// return their spec type and name.
+	ObjectSpec interface {
+		// SpecType returns the spec type of the object.
+		SpecType() string
+		// SpecName returns the spec name of the object.
+		SpecName() string
+	}
```
a8m, line 89:
```suggestion
	// SpecTypeNamer is an interface for objects that can
	// return their spec type and name.
	SpecTypeNamer interface {
		// SpecType returns the spec type of the object.
		SpecType() string
		// SpecName returns the spec name of the object.
		SpecName() string
	}
```

a8m (approved):

PR #2585, opened 2024-03-03
doc/mssql: add document for `type_alias` and `type_table`

giautm:


doc/md/atlas-schema/hcl-types.md
```diff
@@ -1170,6 +1170,104 @@ table "t" {
 }
 ```
 
+### User-defined types
+
+#### Alias Types
```
a8m, line 1175: Wrap it with driver tabs maybe? Because I will add pg to it as well

giautm: I think this document is per driver already.

rotemtam (approved):

PR #2617, opened 2024-03-14
cmd/atlas/internal: fix IAM Authentication for RDS MySQL

ronenlu:


cmd/atlas/internal/cmdext/cmdext.go
```diff
@@ -152,6 +156,24 @@ func AWSRDSToken(ctx *hcl.EvalContext, block *hclsyntax.Block) (cty.Value, error
...
+	}
+	pem, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return cty.NilVal, errorf("reading ca bundle: %v", err)
+	}
+	rootCertPool := x509.NewCertPool()
+	if ok := rootCertPool.AppendCertsFromPEM(pem); !ok {
+		return cty.NilVal, errorf("appending certs from pem: %v", err)
+	}
+	err = mysql.RegisterTLSConfig("rds", &tls.Config{RootCAs: rootCertPool, InsecureSkipVerify: true})
```
giautm, line 173: I don't think this fix is right. The user should download it by their own and update its location to the driver.

a8m, line 173: I don't like working with drivers directly in packages that are not main (atlas packages don't aware to the underlying driver). I prefer creating a driver for `awsmysql`. e.g., https://github.com/google/go-cloud/blob/master/mysql/awsmysql/awsmysql.go

rotemtam: Is it necessary? I'm guessing its posible to provide the CA path to the driver. 
we shouldn't do the magic of forcing the user to trust some cert. its an advanced use case and the user should manage their own CAs.

masseelch: > Is it necessary? I'm guessing its posible to provide the CA path to the driver. we shouldn't do the magic of forcing the user to trust some cert. its an advanced use case and the user should manage their own CAs.

Yes. It is needed. This is how it works with mysql go sql driver sadly.

a8m (approved):

PR #2620, opened 2024-03-15
allow libsql to work with versioned migrations

golanglemonade:
**Issue**: 
- When using `libsql` with versioned migrations I would get an "unsupported dialect" error from Ent 
- The old `libsql` driver did not support `time.Time`

**PR Includes**:

- Updates the `libsql` client to the `github.com/tursodatabase/libsql-client-go`
- Uses the `dialect.Sqlite`  for ent versioned migrations when using `libsql` driver

This allows the use of the following in migrations: 
```
env "turso" {
  url     = "libsql://<REPLACE WITH YOUR SUBDOMAIN>?authToken=${var.token}"
  exclude = ["_litestream*"]
}
```

### DIFF
--- a/cmd/atlas/internal/migrate/migrate.go
+++ b/cmd/atlas/internal/migrate/migrate.go
@@ -76,10 +76,13 @@ func NewEntRevisions(ctx context.Context, ac *sqlclient.Client, opts ...Option) 
 			return nil, err
 		}
 	}
+
+	entDialect := r.getEntDialect()
+
 	// Create the connection with the underlying migrate.Driver to have it inside a possible transaction.
-	entopts := []ent.Option{ent.Driver(sql.NewDriver(r.ac.Name, sql.Conn{ExecQuerier: r.ac.Driver}))}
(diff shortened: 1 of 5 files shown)


cmd/atlas/internal/migrate/migrate.go
```diff
@@ -376,3 +381,12 @@ func DirURL(ctx context.Context, u *url.URL, create bool) (migrate.Dir, error) {
...
 	return d, err
 }
+
+// getEntDialect returns the corresponding ent dialect of the underlying Atlas client
+func (r *EntRevisions) getEntDialect() string {
+	if r.ac.Name == "libsql" {
+		return dialect.SQLite
+	}
+
+	return r.ac.Name
```
a8m, line 391: Thanks for the contribution, but I prefer not to mix driver logic with cmdapi.

Drivers can implement their own RevisionsReadWriter, and I need to move this implementation out so the libsql driver one can reuse it.

PR #2633, opened 2024-03-20
cmd/atlas: labeled envs should take precedence on non-labeled ones

a8m:
The flag below should evaluate only the first block, not both.
```
--env dev
```
```hcl
env "dev" {
  ...
}

env {
  name = atlas.env
  ...
}
```


masseelch:
Yes please. Awesome, I need this!

masseelch (approved):

masseelch (approved):

PR #2643, opened 2024-03-25
cmd/atlas/internal/cmdapi: to flag has precedence

masseelch:


cmd/atlas/internal/cmdapi/schema.go
```diff
@@ -669,10 +669,6 @@ func setSchemaEnvFlags(cmd *cobra.Command, env *Env) error {
 		if err := maySetFlag(cmd, flagFormat, env.Format.Schema.Diff); err != nil {
 			return err
 		}
-	case "push":
-		if err := maySetFlag(cmd, flagTo, strings.Join(srcs, ",")); err != nil {
-			return err
-		}
```
a8m, line 675: why is that? revert

a8m (approved):

cmd/atlas/internal/cmdapi/schema.go
masseelch: This case is executed before the switch. The change I did is equal to this line. No longer needed.

PR #2651, opened 2024-03-31
 cmd/atlas/internal: show more detailed error msg when scheme is not provided to --dir

ronenlu:
image

### DIFF
--- a/cmd/atlas/internal/migrate/migrate.go
+++ b/cmd/atlas/internal/migrate/migrate.go
@@ -332,7 +332,7 @@ const (
 // DefaultDirName is the default directory name.
 const DefaultDirName = "migrations"
 
-// DirURL returns a migrate.Dir to use as migration directory. For now only local directories are supported.
+// DirURL returns a migrate.Dir to use as migration directory.
 func DirURL(ctx context.Context, u *url.URL, create bool) (migrate.Dir, error) {
 	p := filepath.Join(u.Host, u.Path)
 	switch u.Scheme {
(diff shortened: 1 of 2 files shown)


cmd/atlas/internal/migrate/migrate.go
```diff
@@ -344,6 +344,8 @@ func DirURL(ctx context.Context, u *url.URL, create bool) (migrate.Dir, error) {
 		}
 	case DirTypeAtlas:
 		return openAtlasDir(ctx, u)
+	case "":
+		return nil, fmt.Errorf("missing scheme for dir url. Did you mean %q? ", fmt.Sprintf("file://%s", u.Path))
```
a8m, line 348:
```suggestion
	case "":
		return nil, fmt.Errorf("missing scheme for dir url. Did you mean %q? ", fmt.Sprintf("%s://%s", DirTypeFile, u.Path))
```

cmd/atlas/internal/migrate/migrate_test.go
```diff
@@ -71,6 +72,61 @@ func TestNewEntRevisions(t *testing.T) {
...
+			name:     "Valid file URL",
+			u:        &url.URL{Scheme: "file", Path: localDir},
+			create:   false,
+			expected: must(migrate.NewLocalDir(localDir)),
+		},
+		{
+			name:     "Create local dir",
+			u:        &url.URL{Scheme: "file", Path: "new/dir"},
+			create:   true,
+			expected: must(migrate.NewLocalDir("new/dir")),
```
a8m, line 94: Don't create non-temp directories. Use filepath.Join(localDir, "new/dir") instead

cmd/atlas/internal/migrate/migrate_test.go
```diff
@@ -71,6 +72,61 @@ func TestNewEntRevisions(t *testing.T) {
...
+			name:        "No scheme",
+			u:           &url.URL{Path: localDir},
+			create:      false,
+			expectedErr: fmt.Errorf("missing scheme for dir url. Did you mean %q? ", "file://"+localDir),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			dir, err := DirURL(context.Background(), tt.u, tt.create)
```
a8m, line 112: Test `Dir` instead - it calls DirURL and will cover both functions.

a8m (approved):

PR #2706, opened 2024-04-21
sql/sqlx: fix miss makezero bug with `byKeys` sort

alingse:
when I want to add feature for linter [makezero](https://github.com/ashanbrown/makezero) , so I run some github action for real world repos, and the action report this bug.


a8m (approved):

a8m:
Thanks for the fix, @alingse

PR #2758, opened 2024-05-08
chore: update go version to 1.22.3

noamcattan:


a8m (approved):

jerbob92:
Just out of curiosity, what is the reason for bumping the `go` version this specifically? Since `Ent` imports this package this bumps the minimal Go version of other projects as well, shouldn't the `toolchain` directive be used here and the `go` directive to indicate the minimal supported version?

a8m:
https://pkg.go.dev/vuln/GO-2024-2824
https://groups.google.com/g/golang-announce/c/wkkO4P9stm0

We'll add additional context to the next PRs like this one in the future.
cc @noamcattan

jerbob92:
@a8m I get that it's good to make sure you use a new Go version in builds, but isn't that what the toolchain directive is used for? I thought the `go` directive is to indicate which version is required minimally to use the package.

a8m:
You're right, but how does it affect you if the ent package uses a pinned version of Atlas? (0.19.1~).

jerbob92:
It's a bit of an edge case where we import Atlas ourselves as well for schema generation, and Dependabot is failing to automatically create MR's for the update because of the Go directive because it's using Go 1.22.2 internally.

So not a really big deal, I just saw some errors passing by and was curious what was causing them.

PR #2776, opened 2024-05-12
sql/postgres: unset default search_path

a8m:
When operating on realm-scope we expect all objects to be qualified. However, the default configuration of pg sets the search_path to `$user, public` causes all references to public objects to return from pg_catalog without qualifiers.

### DIFF
--- a/sql/postgres/inspect.go
+++ b/sql/postgres/inspect.go
@@ -29,7 +29,12 @@ type inspect struct{ *conn }
 var _ schema.Inspector = (*inspect)(nil)
 
 // InspectRealm returns schema descriptions of all resources in the given realm.
-func (i *inspect) InspectRealm(ctx context.Context, opts *schema.InspectRealmOption) (*schema.Realm, error) {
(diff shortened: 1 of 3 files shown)


giautm (approved):

sql/postgres/inspect.go
```diff
@@ -86,6 +91,30 @@ func (i *inspect) InspectRealm(ctx context.Context, opts *schema.InspectRealmOpt
 	return schema.ExcludeRealm(r, opts.Exclude)
 }
 
+// noSearchPath ensures the session search_path is clean when inspecting realms to ensures all
+// referenced objects in the public schema (or any other default search_path) are returned
+// qualified in the inspection.
+func (i *inspect) noSearchPath(ctx context.Context) (func() error, error) {
+	rows, err := i.QueryContext(ctx, "SELECT current_setting('search_path'), set_config('search_path', '', false)")
```
giautm, line 98: cc @datdao

PR #2781, opened 2024-05-15
Fix typo ('Doctrine' instead of 'Doctine')

groovytron:
It is a simple typo fix ('Doctrine' instead of `Doctine').

Thanks for your review.


a8m (approved):
Thanks, @groovytron

PR #2791, opened 2024-05-20
sql/mysql: support inet4 + inet6 for MariaDB 10.10+

luantranminh:
Fixed: https://github.com/ariga/atlas-provider-gorm/issues/46

### DIFF
--- a/sql/mysql/driver.go
+++ b/sql/mysql/driver.go
@@ -431,6 +431,9 @@ const (
 	TypeGeometryCollection = "geometrycollection" // Geometry_type::kGeometrycollection
 
 	TypeUUID = "uuid" // MariaDB supported uuid type from 10.7.0+
+
+	TypeInet4 = "inet4" // MariaDB type for storage of IPv4 addresses, from 10.10.0+.
+	TypeInet6 = "inet6" // MariaDB type for storage of IPv6 addresses, from 10.10.0+.
 )
 
 // Additional common constants in MySQL.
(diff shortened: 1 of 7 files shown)


.github/workflows/ci-dialect_oss.yaml
```diff
@@ -170,6 +170,33 @@ jobs:
         env:
           MYSQL_DATABASE: test
           MYSQL_ROOT_PASSWORD: pass
+  integration-maria1010:
```
a8m, line 173: No need for this. Remove it

sql/mysql/inspect.go
```diff
@@ -911,6 +911,12 @@ type (
 		Values []string
 	}
 
+	// NetworkType stores an IPv4 or IPv6 address.
+	NetworkType struct {
+		schema.Type
+		T string
+	}
```
a8m, line 918: Please keep only parsing/types tests and remove the integrations ones. No need to run all our tests on another database just for this new type.

sql/mysql/driver.go
```diff
@@ -431,6 +431,9 @@ const (
 	TypeGeometryCollection = "geometrycollection" // Geometry_type::kGeometrycollection
 
 	TypeUUID = "uuid" // MariaDB supported uuid type from 10.7.0+
+
+	TypeINET4 = "inet4" // MariaDB type for storage of IPv4 addresses, from 10.10.0+.
+	TypeINET6 = "inet6" // MariaDB type for storage of IPv6 addresses, from 10.10.0+.
```
a8m, line 436: Use these instead: TypeInet4, TypeInet6

a8m (approved):

PR #2793, opened 2024-05-21
sql/postgres: remove dynamic default comparison

a8m:
As dev-url is not a standard option, this comparison is not needed anymore.


giautm (approved):

a8m:
This breaks the compatibility with Cockroach. I'll skip this for now.

Issue #2472, opened 2024-01-23
`schema inspect` with MySQL Unix socket uses socket path as the schema name


ariel-co:
I'm trying to connect via Unix socket while specifying the password. This works with the command line client:

```
$ mysql -s -u root -S /san/mysql-fs/mysql/mysql.sock -p
Enter password:
mysql>

```

With Atlas, putting the password (after urlencoding a '=' in the middle) in the URL, this is what I get:
```

$ ./atlas schema inspect -u "mysql+unix://root:mysikrit%3Dpasswd@/san/mysql-fs/mysql/mysql.sock?database=mydatabase" > mydatabase.hcl
Error: mysql: schema "san/mysql-fs/mysql/mysql.sock" was not found

```

Notice how it strips the first `/` from the socket path and takes that as the schema name. Removing `?database=mydatabase` has no effect.

    $ ./atlas version
    atlas version v0.18.1-60783c4-canary

xaverhimmelsbach:
This is broken for us as well. It should have been fixed in https://github.com/ariga/atlas/issues/1587 but looking at the tests in https://github.com/ariga/atlas/blob/master/sql/mysql/driver_test.go for ParseURL, everything seems to work fine.

Maybe the leading slash gets stripped and the schema is extracted in an earlier, step. Sadly, I don't know the code well enough to find any place like that. @a8m do you have any input on this?

c24-jens-albert:
We encountered the same problem, but not only with `schema`, but also with `migrate`:
```
atlas migrate apply --dir file://db/migration --url "mysql+unix://user:$DB_PW@/run/proxysql/proxysql.sock?database=dbname"

Error: add new schema named "run/proxysql/proxysql.sock": Error 1044 (42000): Access denied for user 'user'@'10.10.10.100' to database 'run/proxysql/proxysql.sock'

```
But according to the [Docs](https://atlasgo.io/concepts/url) and #1587 it should work? Is this a known problem?

a8m:
<img width="537" alt="image" src="https://github.com/ariga/atlas/assets/7413593/12ce88eb-3eaf-4b5b-9b74-91e357becffd">

This won't make me support you faster, sorry.

a8m:
@ariel-co, I just tested it locally and it seems to work without any issues. You can try upgrading your Atlas version (download  it and don't compile from source) and report if you face any issues with it. 

[...]

PR #2847, opened 2024-06-07
Add schema parsing for unix scheme urls

LudvigLundberg:
Fix for #2472

Faced the same issue (altough with atlas migrate) when trying to migrate through a unix socket, seems the issue is here: 

```go
return &sqlclient.URL{URL: u, DSN: dsn(u), Schema: strings.TrimPrefix(u.Path, "/")}
```

Since the url path for unix sockets can contain multiple `/` characters.

Steps to reproduce issue (tested on macOS):

```bash
brew install mysql
mysql.server start
mysql -u root

# mysql 
create schema test;

# in atlas cmd directory, current master branch
go build .
./atlas schema inspect -u "mysql+unix://root@/tmp/mysql.sock?database=test"

# Error: mysql: schema "tmp/mysql.sock" was not found

git checkout fix-unix-schema-parse
go build .
./atlas schema inspect -u "mysql+unix://root@/tmp/mysql.sock?database=test"

# schema "test" {
#   charset = "utf8mb4"
#   collate = "utf8mb4_0900_ai_ci"
# } 
```

There is a function in the go-sql-driver/mysql that could be used to parse the DSN and extract the schema, but didn't want to add another dependency to this repo without approval so kept it simple here! 

Feel free to suggest improvements or comments on the code, I don't have all the context of atlas so might have missed something! @ttc0419 could maybe have a look and see if it solved the issue for them as well!

### DIFF
--- a/sql/mysql/driver.go
+++ b/sql/mysql/driver.go
@@ -318,7 +318,7 @@ func (parser) ParseURL(u *url.URL) *sqlclient.URL {
 	v := u.Query()
 	v.Set("parseTime", "true")
 	u.RawQuery = v.Encode()
-	return &sqlclient.URL{URL: u, DSN: dsn(u), Schema: strings.TrimPrefix(u.Path, "/")}
+	return &sqlclient.URL{URL: u, DSN: dsn(u), Schema: parseSchema(u)}
 }
 
 // ChangeSchema implements the sqlclient.SchemaChanger interface.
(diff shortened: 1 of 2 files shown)


binarious:
@a8m Any news? Still having trouble with this one.

a8m:
Thanks for your contribution, @LudvigLundberg. Your commit was added with #3195 
image

PR #2863, opened 2024-06-15
sql/postgres: add support for row level security

### LINKED ISSUE #1071: Row level security policies
Hello!

I would love to see [RLS ](https://www.postgresql.org/docs/current/ddl-rowsecurity.html) added to Atlas! 
I currently make a lot of use of them, especially when using Supabase!

The proposal would be to have the following spec:

```hcl
table "example" {

	column "user_id" {
		type = uuid
	}

       row_level_security = true

	policy "admin" {
		select {
			using = "auth.uid() = user_id"
		}
		delete {
			using = "auth.uid() = user_id"
		}
		update {
			check = "auth.uid() = user_id"
			using = "auth.uid() = user_id"
			to = "authentication" // specify schema or default to table schema
		}
		insert {
			check = "auth.uid() == user_id"
		}
	}
}
```

- `row_level_security (bool)` would allow us to enable / disable RLS on the table
- `policy` would be an object that represents the policy rules. There can be many policies on a table

the name of the policy for each operation would be `{table_name}_{policyname}_operation` for instance, given the example above atals should migrate:

```sql
ALTER TABLE example ENABLE ROW LEVEL SECURITY;

CREATE POLICY example_admin_select ON example
    FOR SELECT
    USING (auth.uid() = user_id);

CREATE POLICY example_admin_delete ON example
    FOR DELETE
    USING (auth.uid() = user_id);

CREATE POLICY example_admin_update ON example
    FOR UPDATE
    WITH CHECK ( auth.uid() = user_id )
    USING (auth.uid() = user_id);

CREATE POLICY example_admin_insert ON example
    FOR INSERT
    WITH CHECK ( auth.uid() = user_id )
```

Any thoughts?

meringu:
Hello, first time contributor, so please let me know if I'm going about this the wrong way. I also haven't used RLS and with Postgres extensively.

- I modelled the HCL based on discussions in #1071 which mimics the SQL statements used to create policies. See the .mdx change for how to use.
- I've tested this manually against a Postgres DB in a docker container and seems to behave as expected including idempotency.
- Unit tests covering the changes.
- I couldn't get the integration tests running on either my branch or master.
- I've included some fixes to some broken links in the contributing document.

resolves #1071


a8m:
Hey @meringu, and thanks for your contribution.

Unfortunately, I can't accept this PR because we've already added support for [RLS](https://atlasgo.io/atlas-schema/hcl#row-level-security-policy) internally (the design is slightly different than yours) and we exposed it today for global use. You're welcome to give it a try and share your impressions 🙏 

Sorry that you spent time on it and we can't accept these changes - I communicated in our Discord that this should be out in this release (or the one after). Roles and users are going to be joined to this feature.

If you still want to contribute the changes to the contributing document, you're welcome.

PR #2889, opened 2024-06-25
sql/postgres: dont run default change query if no connection available 

ronenlu:


sql/postgres/diff.go
```diff
@@ -119,7 +119,7 @@ func (d *diff) defaultChanged(from, to *schema.Column) (bool, error) {
 	//	SELECT ARRAY[1] = '{1}'::int[]
 	//	SELECT lower('X'::text) = lower('X')
 	//
-	if (fromX || toX) && d.conn.ExecQuerier != nil {
+	if (fromX || toX) && d.conn.ExecQuerier != nil && d.conn.ExecQuerier != sqlx.NoRows {
```
dorav, line 122: Is there a way to add a test for this?

a8m (approved):

PR #2898, opened 2024-06-26
Support risingwave schema inspect

neverchanje:
Hi, atlasgo devs,

I want to contribute the integration for RisingWave, a pgwire-compatible database primarily targeting streaming. We've heard from several users praising the usability of atlasgo, and we believe integrating both tools would streamline the user experience.

At first, I tried directly connecting RisingWave via the Postgres connector, but atlasgo threw serveral errors due to the lack of some SQL features at the RisingWave side. These features are currently unnecessary in RisingWave's context, so I had to tweak atlasgo to work properly with RisingWave.

A major modification is that I introduced a separate branch to check if atlas is connecting to RisingWave:

```go
	c.risingwave, err = c.isRisingWaveConn()
	if err != nil {
		return nil, fmt.Errorf("postgres: failed checking if connected to RisingWave: %w", err)
	}
	if c.risingwave {
		return noLockDriver{
			&Driver{
				conn:        c,
				Differ:      &sqlx.Diff{DiffDriver: &risingwaveDiff{diff{c}}},
				Inspector:   &risingwaveInspect{inspect{c}},
				PlanApplier: &planApply{c},
			},
		}, nil
	}
```
 

If it is, all further execution will go to RisingWave-specific implementation.

Another tricky part is:

```diff
-	LEFT JOIN pg_depend AS t5 ON t5.classid = 'pg_catalog.pg_class'::regclass::oid AND t5.objid = t3.oid AND t5.deptype = 'e'
+	LEFT JOIN pg_depend AS t5 ON t5.classid::text = 'pg_catalog.pg_class' AND t5.objid = t3.oid AND t5.deptype = 'e'
```

I slightly adjusted the comparison, removing `::regclass::oid`. I used `text` instead of `oid` for comparison. This is because RisingWave doesn't support `regclass` nor `oid`. I thin using `text` would achieve better compatibility with other PG-like databases. 😄 I hope this modification would not break anything.

### DIFF
--- a/sql/postgres/driver.go
+++ b/sql/postgres/driver.go
@@ -38,7 +38,10 @@ type (
 		schema string
 		// System variables that are set on `Open`.
 		version int
-		crdb    bool
+		// Whether the connected database is a CockroachDB.
+		crdb bool
+		// Whether the connected database is a RisingWave.
+		risingwave bool
 	}
 )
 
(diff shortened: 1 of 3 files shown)


a8m:
Hey! Thanks for the contribution, @neverchanje.

Unfortunately, we cannot accept these changes and commit to support this new database. If the database is compatible with the PostgreSQL information schema, it should work with the current driver as is. 

If you are working at RisingWave, you are welcome to send me a message on our Discord server and we can discuss this partnership.

PR #2954, opened 2024-07-11
sql/migrate: ExecuteTo should respect versions before last checkpoint

a8m:


giautm (approved):

sql/migrate/migrate.go
```diff
@@ -919,18 +919,48 @@ func (e *Executor) ExecuteN(ctx context.Context, n int) (err error) {
...
+		if err != nil {
+			return err
+		}
 	default:
-		pending = pending[:idx+1]
+		if pending, err = e.Pending(ctx); err != nil {
+			return err
+		}
+		// This code is relevant only
+		// Strip pending files greater given version.
```
ronenlu, line 955:
```suggestion
		// This code is relevant only if
		// Strip pending files greater given version.
```

PR #3027, opened 2024-08-06
sql/migrate: add FormatFile for TemplateFormatter

a8m:
Avoid `len` checks whenever `Format` is called on the default template.

### DIFF
--- a/cmd/atlas/internal/migratelint/run.go
+++ b/cmd/atlas/internal/migratelint/run.go
@@ -16,7 +16,6 @@ import (
 	"text/template"
 	"time"
 
-	"ariga.io/atlas/cmd/atlas/internal/cmdlog"
 	"ariga.io/atlas/sql/migrate"
 	"ariga.io/atlas/sql/sqlcheck"
 	"ariga.io/atlas/sql/sqlclient"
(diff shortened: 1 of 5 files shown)


rotemtam (approved):

cmd/atlas/internal/migratelint/run.go
```diff
@@ -226,7 +225,12 @@ var (
 			}
 			return append(lines, strings.Join(words[j:], " "))
 		},
-	})
+		"cyan":         color.CyanString,
+		"green":        color.HiGreenString,
+		"red":          color.HiRedString,
+		"redBgWhiteFg": color.New(color.FgHiWhite, color.BgHiRed).SprintFunc(),
+		"yellow":       color.YellowString,
+	}
```
masseelch, line 233: Why? This isn't DRY anymore.

a8m: I'm moving all serialized logged objects to cmdlog, and new parts depend on `migratelint` report objects, which creates me circular dep.

a8m: Eventually, the `migratelint` templates will be moved to this as well.

PR #3119, opened 2024-09-14
Fixed documentation typo in `projects.mdx`

Maxim-Durand:
A documentation example had a typo referencing `src = data.hcl_schema.app.url` when it should be `src = data.hcl_schema.app.path`


a8m (commented):
This data source exports a URL to load it from. Thank you for your willingness to contribute to Atlas.

PR #3238, opened 2024-11-29
sql/migrate: support custom Delimiter in Plan

giautm:


luantranminh (approved):

sql/migrate/dir.go
```diff
@@ -528,11 +528,22 @@ func CheckVersion(v string) error {
...
 }
 
+// delim returns a directive with the given delimiter.
+func delim(s string) string {
+	if s == "" {
+		return ""
+	}
+	// Escape delimiters. e.g. "\n" => "\\n".
+	s = strings.NewReplacer("\n", `\n`, "\r", `\r`, "\t", `\t`).Replace(s)
+	return fmt.Sprintf("-- atlas:%s %s\n", directiveDelimiter, s)
```
a8m, line 538: I would return the directive as-is (without extra \n), and handle the double \n in the template.

giautm: It make the template more complex, because I don't add the directive when we don't use custom delimiter

a8m: The reason I suggest this is because I want to add support for extra file directives by the template/planner (e.g., txmode).

a8m:
```
{{- with delim .Delimiter }}
  {{- printf "%s\n\n" . }}
{{- end }}
{{- range .Changes }}{{ with .Comment }}{{ printf "-- %s%s\n" (slice . 0 1 | upper ) (slice . 1) }}{{ end }}{{ printf "%s%s\n" .Cmd (or $.Delimiter ";") }}{{ end }}
```

a8m (approved):

PR #3241, opened 2024-12-05
sql/migrate: add `Locker` as require implement by all drivers

giautm:


cmd/atlas/internal/migratelint/lint.go
```diff
@@ -400,7 +400,9 @@ func (d *DevLoader) inspect(ctx context.Context) (*schema.Realm, error) {
 func (d *DevLoader) lock(ctx context.Context) (schema.UnlockFunc, error) {
 	l, ok := d.Dev.Driver.(schema.Locker)
 	if !ok {
-		return nil, errors.New("driver does not support locking")
+		// Some drivers do not support advisory locks.
+		// In this case, we just return a no-op unlock function.
+		return func() error { return nil }, nil
```
a8m, line 405: I'd prefer the driver to implement the no-op lock rather than skipping it here.

giautm: In that case, the `Locker` should be require to implement by all drivers. Currently it's optional.

giautm: Then, I add the Locker first, we can continues add other interface late.

a8m (approved):

PR #3251, opened 2024-12-10
cmd/atlas/docker: allow provide `User` via options

giautm:


cmd/atlas/internal/docker/docker_test.go
```diff
@@ -416,14 +433,20 @@ func TestImageURL(t *testing.T) {
...
+		Config: Config{
+			driver: "postgres",
+			User:   url.UserPassword("postgres", "pass"),
+		},
+		Port: "5432",
+	}
 	u, err := c.URL()
 	require.NoError(t, err)
-	require.Equal(t, "postgres://postgres:pass@localhost:5432/?sslmode=disable", u.String())
+	require.Equal(t, "postgres://postgres:pass@localhost:5432?sslmode=disable", u.String())
```
a8m, line 445: Why omitting the /?

giautm: Is it important? because the database name is empty. So, I think no-path and `/` will have same effect.

a8m: I don’t know tbh. If we change it, we need to test it on the different database drivers.

giautm: Reverted this change,

a8m (approved):

PR #3253, opened 2024-12-11
cmd/atlas: attr to control color

rotemtam:


cmd/atlas/internal/migratelint/run.go
```diff
@@ -397,6 +411,9 @@ type (
 		Text   string      `json:"Text,omitempty"`   // Step description.
 		Error  string      `json:"Error,omitempty"`  // Error that cause the execution to halt.
 		Result *FileReport `json:"Result,omitempty"` // Result of the step. For example, a diagnostic.
+
+		// Text formatting
+		NameColor string `json:"NameColor,omitempty"` // Color of the name field when printed to the terminal.
```
a8m, line 416:
```suggestion
		NameColor string `json:"-"` // Color of the name field when printed to the terminal.
```
My mistake. This is supposed to be dropped in marshalling

a8m (approved):

a8m (approved):

PR #3281, opened 2024-12-22
cmd/atlas/internal/cmdapi: new "schema diff" flag to error out on diff.

pylaligand:
This makes it easier to programmatically verify that migrations are up-to-date with e.g. an ORM-generated schema in a CI context by observing the command's return code rather than parsing its output.

### DIFF
--- a/cmd/atlas/internal/cmdapi/cmdapi.go
+++ b/cmd/atlas/internal/cmdapi/cmdapi.go
@@ -271,6 +271,7 @@ const (
 	flagDryRun         = "dry-run"
 	flagEnv            = "env"
 	flagExclude        = "exclude"
+	flagFailOnDiff     = "fail-on-diff"
 	flagFile           = "file"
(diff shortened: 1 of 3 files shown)


pylaligand:
Happy new year! Anyone to review this? @a8m maybe?

a8m:
Thanks for the contribution, @pylaligand. Currently, in order to implement something like this in CI, you can do it this way:

```bash
atlas schema diff \
  --from "file://migrations" \
  --to env://orm \
  --env local \
  --format '{{ if .Changes }}The failure message{{ end }}' | grep '.' && exit 1
```

Or, convert it to `if [ -n "$output" ]; then .. fi`. 

This is flexible enough because users can control the error message and the format of the failure. If we'll add such flag, we also need to support a way to customize its message, which I prefer to avoid.

pylaligand:
Makes sense, thanks.

PR #3302, opened 2025-01-09
cmd/atlas/internal: add function that return the URL used for inspection

ronenlu:


cmd/atlas/internal/cmdapi/schema.go
```diff
@@ -447,8 +447,9 @@ func schemaInspectRun(cmd *cobra.Command, _ []string, flags schemaInspectFlags,
 	if err != nil {
 		return err
 	}
-	maySuggestUpgrade(cmd)
```
masseelch, line 450: Why did you remove this?

cmd/atlas/internal/cmdlog/cmdlog.go
```diff
@@ -785,6 +786,15 @@ func (s *SchemaInspect) MarshalHCL() (string, error) {
 	return string(spec), nil
 }
 
+// RedactedURL returns the inspected url redacted.
+func (s *SchemaInspect) RedactedURL() (string, error) {
```
masseelch, line 790: It is dangerous to have the URL still be un-redacted. I advise to have it `URL`, unexport the `url` and add it as either arg or optional arg to the constructor.

a8m (approved):

PR #3309, opened 2025-01-14
cmd/atlas/docker: allow set ENV when run the container

giautm:


kevbook:
Is there docs for this?

a8m:
There's a new `env` attribute in the `docker` block: https://atlasgo.io/concepts/dev-database#baseline-schema

A new doc should be generated for this object soon. cc @giautm

PR #3327, opened 2025-01-20
sql/migrate: remove partial hashes on successful apply

### LINKED ISSUE #3311: The memory usage is growing unlimitedly
# issue

The memory usage is growing unlimitedly as the revision table grows.

# usecase

Especially when using atlas for DML frequently, `atlas_schema_revisions` table grows rapidly and so does the memory usage, which is not scalable.

# code

https://github.com/ariga/atlas/blob/5bb5954fa4b91abb11dad4dd95d87164896bd59e/cmd/atlas/internal/migrate/migrate.go#L148-L151

masseelch:
Close #3311


a8m (approved):
💯

PR #3333, opened 2025-01-21
cmd/atlas/internal/cmdext: move external_schema to ent repo

a8m:
It relies now on non-oss logic


giautm (approved):

MRocholl:
Hello!
Thank you for the amazing work on Atlas. 
Im sad to see this change. We use the OSS version and rely on this feature.
How and why did this happen?  This is also not reflected in the docs at all. 
https://atlasgo.io/atlas-schema/external
Am I reading the docs wrong? Is there a different way to achieve this now? 
Thank you for clarifying.

MRocholl:
Aha! Thanks for the explanation.
What is the purpose of the OSS version then? 
I'll go with the official binary then.

PR #3334, opened 2025-01-21
sql/migrate: fixed delimiter on the same line with the line comment

giautm:


sql/migrate/migrate.go
```diff
@@ -59,6 +59,19 @@ type (
 	}
 )
 
+// Statement format the change's command
+func (p Plan) Statement(c *Change) string {
```
giautm, line 63: Not sure this should a templace function or method on plan.

a8m, line 63: sqltool contains different formatter. The plan is agnostic to the actual format.

a8m: Maybe make it a template function?

Issue #2509, opened 2024-02-03
atlas migrate apply --url $DATABASE_URL fails


domenkozar:
```
Error: sql/migrate: execute: read revisions: pq: relation "atlas_schema_revisions.atlas_schema_revisions" does not exist
```

Using Atlas 0.18.

a8m:
Hey, @domenkozar 👋 

`migrate apply` is one of the most frequently used Atlas commands, so I suspect there may be something incorrect in your setup unless I'm overlooking something.

If the `atlas_schema_revisions` table resides in its own schema, it means you are operating on the entire database (or multiple schemas), not just a single one - is it correct? Could you please provide a few commands that reproduce this issue, similar to the one below?

<img width="1131" alt="image" src="https://github.com/ariga/atlas/assets/7413593/5255bb65-a55c-4269-bac8-bca0d8d852f5">

domenkozar:
I'm not sure what else to provide, I'm using https://neon.tech/ for postgres store.

`--revisions-schema public` workarounds the issue.

a8m:
Hey! Just fyi, I found the issue for this - https://github.com/neondatabase/neon/issues/3148. TL;DR, Neon does not respect the `search_path` parameter and we'll need to work around this.

jakubno:
@a8m I have the same problem. On a brand new database created on Supabase I am getting:
```
Error: upsert revision-table id: pq: relation "atlas_schema_revisions.atlas_schema_revisions" does not exist
```

But `atlas_schema_revisions` table is present in `public` schema.

I am running:
```
atlas migrate apply --env local --url $(POSTGRES_CONNECTION_STRING) --baseline "20000101000000"
```

I am on:
```
atlas version v0.27.0
```

and my `atlas.hcl`:
```hcl
// Define an environment named "local"
env "local" {
  // Declare where the schema definition resides.
  // Also supported: ["file://multi.hcl", "file://schema.hcl"].
  src = "ent://pkg/schema/"

  // Define the URL of the Dev Database for this environment
  // See: https://atlasgo.io/concepts/dev-database
  dev = "docker://postgres/15/dev"

  migration {
    // Define the path to the migration directory.
    // See: https://entgo.io/docs/migrate/#migration-directory
    dir = "file://migrations"
  }

  diff {
    skip {
      drop_func = true
      drop_trigger = true
      drop_schema = true
    }
  }
}
```

If you need anything else, let me know, happy to provide more info.

[...]

PR #3378, opened 2025-02-11
cmd/atlas/internal/migrate: use schema qualifier for pooler-based pg services

a8m:
Currently, this change applies only to neon and supabase and fixes: https://github.com/ariga/atlas/issues/2509. However, we plan to migrate away from entmigrate to driver-specific store. This way, we can detect other poolers such as pgbouncer.

### DIFF
--- a/cmd/atlas/internal/migrate/migrate.go
+++ b/cmd/atlas/internal/migrate/migrate.go
@@ -13,6 +13,8 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"slices"
+	"strings"
 	"time"
 
 	"ariga.io/atlas/cmd/atlas/internal/migrate/ent"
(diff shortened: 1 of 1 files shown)


cmd/atlas/internal/migrate/migrate.go
```diff
@@ -257,6 +265,21 @@ func (r *EntRevisions) Migrate(ctx context.Context) (err error) {
 	}))
 }
 
+// maySetSchemaQualifier sets the schema on the atlas_schema_revisions table
+// for cases the migration should use qualified identifiers due to some limitation
+// in PostgreSQL services that use "connection pooler" and do not support the "search_path"
+// parameter. See: https://github.com/ariga/atlas/issues/2509.
+func (r *EntRevisions) maySetSchemaQualifier(t *schema.Table) {
```
ronenlu, line 272: add test?

a8m: Mocking the `database/sql/driver` in unit tests is really cumbersome (I just tried it), and in integration tests, it requires changing `/etc/hosts`.  

It's not really worth it. I'll let it run this way for some time, and then we can test it on all PGs.

PR #3406, opened 2025-02-27
sql/schema: added `IsType` to compare between two types

giautm:


sql/schema/schema.go
```diff
@@ -780,3 +780,20 @@ func UnderlyingType(t Type) Type {
 	}
 	return t
 }
+
+// IsType reports if t1 is the same as t2 or if t1 is an alias of t2.
+func IsType(t1, t2 Type) bool {
```
a8m, line 785: Maybe name SameType instead?

sql/schema/schema.go
```diff
@@ -780,3 +780,20 @@ func UnderlyingType(t Type) Type {
...
+	// Comparing might panic due to mismatch types.
+	defer func() { recover() }()
```
a8m, line 787: We can avoid this defer call by using reflect.TypeOf

sql/schema/schema.go
```diff
@@ -780,3 +780,20 @@ func UnderlyingType(t Type) Type {
...
+	if t1 == nil || t2 == nil {
+		return t1 == t1
+	}
```
a8m, line 790: Typo (t1 == t1)

sql/schema/schema.go
```diff
@@ -780,3 +780,20 @@ func UnderlyingType(t Type) Type {
...
+	t1, t2 = UnderlyingType(t1), UnderlyingType(t2)
+	if t1 == t2 {
```
a8m, line 792: If we compare the underlying types, the logic is not the same anymore (as dependsOnT).

sql/schema/schema.go
giautm: I still prefer `IsType` or `IsSameType`

a8m (approved):

PR #3448, opened 2025-04-02
cmd/atlas: add `--dry-run` flag to migrate diff cmd

ronenlu:


a8m (approved):

masseelch (commented):
The command should return that the flag is not supported if it is present here.

cmd/atlas/internal/cmdapi/cmdapi_oss.go
```diff
@@ -214,6 +214,9 @@ func migrateLintRun(cmd *cobra.Command, _ []string, flags migrateLintFlags, env
 }
 
 func migrateDiffRun(cmd *cobra.Command, args []string, flags migrateDiffFlags, env *Env) error {
+	if flags.dryRun {
+		return errors.New("--dry-run not supported by the community version")
```
masseelch, line 218:
```suggestion
		return errors.New("'--dry-run' is not supported in the community version")
```

PR #3475, opened 2025-04-27
cmd/atlas/internal/cmdapi: add command line flag for lock name

hanurimies:
In our application we use single database and different schemas for each microservice. There are several reasons for using this kind of setup.

The migrate command used a fixed name for the migration lock (PG advisory lock). That caused issues in our application: one microservice running a lenghty migration caused all other microservices to be waiting for the lock to check and run their migrations.

To remedy the situation, this commit makes the lock name to be configurable using a command line flag. Condirered also making the locking to use schema name as lock name, but that might not be what other applications want. Introducing a separate command line flag for the lock name offers the most flexible solution.

The default behaviour is the same as before. If the lock name flag is not given, the old fixed name is used for lock.

### DIFF
--- a/cmd/atlas/internal/cmdapi/cmdapi.go
+++ b/cmd/atlas/internal/cmdapi/cmdapi.go
@@ -283,6 +283,7 @@ const (
 	flagGitDir         = "git-dir"
 	flagLatest         = "latest"
 	flagLockTimeout    = "lock-timeout"
+	flagLockName       = "lock-name"
 	flagLog            = "log"
 	flagPlan           = "plan"
 	flagRevisionSchema = "revisions-schema"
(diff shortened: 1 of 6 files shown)


a8m:
Thanks for the contribution, @hanurimies.

I'm not sure this is how we want to resolve this issue. We already have a unique resource per directory/migration, which is the revisions table. We can handle it without exposing this to the user and making the API error-prone.

I suggest you open an issue with this context, and the team will discuss it. We won't be able to accept external contributions in this area because it touches some parts that involve our internal codebase.

hanurimies:
Thanks for the quick response!

I'll open an issue. Internally we tried this solution and it works, but sure, better to use the more robust solution that you referred to.

hanurimies:
Issue: https://github.com/ariga/atlas/issues/3476

PR #3490, opened 2025-05-08
internal/integration: allow passing custom envs when when doing > in …

ronenlu:
…tescript cli


internal/integration/script_test.go
```diff
@@ -446,6 +446,12 @@ func cmdCLI(ts *testscript.TestScript, neg bool, args []string, dbURL, devURL, c
 			"PATH="+ts.Getenv("PATH"),
 			"DOCKER_HOST="+ts.Getenv("DOCKER_HOST"),
 		)
+		for _, env := range envs {
+			if env == "" {
+				continue
+			}
```
a8m, line 452: don't pass  empty strings

a8m (approved):

PR #3501, opened 2025-05-12
cmd/atlas/docker: let Docker Engine allocate a free port

giautm:
This fixed a case in which the port was free on the machine running atlas-cli but unavailable on the machine running the Docker engine. (remote docker engine via `DOCKER_HOST`)

Because the `freePort()` function only able to checks free ports on the current machine.


masseelch (approved):

a8m:
@masseelch, we still didn’t find why it’s not working on Windows.

giautm:
> @masseelch, we still didn’t find why it’s not working on Windows. 

You may missed it, but I found. Please re-check messages in the thread.

a8m (approved):

luantranminh (approved):

PR #3577, opened 2025-08-11
all: upgrade go to v1.24.6

noamcattan:


giautm (approved):

DaenaBradfield:
@noamcattan this will fix a few issues with container scanning issues we are having. We are currently getting `database/sql: Postgres Scan Race Condition` and the proposed solution is to upgrade the go version to 1.24.6. Once this is in we would really appreciate the updated tag 👍

PR #3612, opened 2025-10-30
fix: allow not having an empty line after directives

jay-babu:


jay-babu:
Any update on this? it is a simple fix it seems like?

PR #3619, opened 2025-11-14
all: bump Go version to 1.25.4

igolaizola:
Make it compatible with latest go version.

### DIFF
--- a/cmd/atlas/go.mod
+++ b/cmd/atlas/go.mod
@@ -1,6 +1,6 @@
 module ariga.io/atlas/cmd/atlas
 
-go 1.24.9
+go 1.25.4
 
 replace ariga.io/atlas => ../..
 
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module ariga.io/atlas
 
-go 1.24.9
+go 1.25.4
 
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.0
--- a/internal/integration/go.mod
+++ b/internal/integration/go.mod
@@ -1,6 +1,6 @@
 module ariga.io/atlas/internal/integration
 
-go 1.24.9
+go 1.25.4
 
 replace ariga.io/atlas => ../../
 


igolaizola:
I have a project stucked in version 1.24 due to this dependency. It will be great if we can have a tagged version that is compatible with go 1.25

cc @giautm

giautm:
> I have a project stucked in version 1.24 due to this dependency. It will be great if we can have a tagged version that is compatible with go 1.25
> 
> 
> 
> cc @giautm

What is the error you got on the application? The old version of Go never causing issue on the newer version of Go.

igolaizola:
> > I have a project stucked in version 1.24 due to this dependency. It will be great if we can have a tagged version that is compatible with go 1.25
> > cc @giautm
> 
> What is the error you got on the application? The old version of Go never causing issue on the newer version of Go.

I'm using atlas in combination with Ent, and it fails with the following error when running `atlas migrate diff`

```
Error: loading schema: entc/load: parse schema dir: /path/ent/schema/entity.go:1:1: package requires newer Go version go1.25 (application built with go1.24)
```

If I downgrade my project to go 1.24 it works fine.

giautm:
Which version of atlas you have?

giautm:
Also entgo

igolaizola:
entgo.io/ent v0.14.4
atlas version v0.38.1-0f81633-canary

masseelch:
@igolaizola Can you share how you are loading the schema for ent into atlas?

igolaizola:
> @igolaizola Can you share how you are loading the schema for ent into atlas?

```
atlas migrate diff my_diff_name --dir "file://ent/migrate/migrations" --to "ent://ent/schema?globalid=1" --dev-url "postgres://user:pass@localhost:5432/test?sslmode=disable&search_path=public"
```

giautm:
That's weird. We also using Go 1.25 in our backend. And we used both atlas, entgo and setup like you did. Can you verify that's coming from sth else? Because, change the Go version like this PR won't resolve the issue you facing.

igolaizola:
What is the best way to compile atlas binary locally, so I can test it myself? I'm trying with a simple `go build cmd/atlas/main.go` but that doesn't seem to work.

giautm:
Hey @igolaizola Please re-download atlas and give a try, we've upgraded atlas-cli to use Go 1.25 and config it fixed this issue. But, I recommend you to use the new way to load Ent's schema to avoid this issue again.

https://atlasgo.io/blog/2025/03/10/v032-atlas-ai-sql-imports#ent-loader

igolaizola:
I've tested with the latest atlas binary and it works now.

Thanks for your help and for the recommendation.

PR #3650, opened 2026-01-15
postgres: add Aurora DSQL driver support

### LINKED ISSUE #3649: Aurora DSQL: datatype json not supported error during schema introspection
Following up on #3353 (advisory lock issue, resolved with `--skip-lock`), there's another blocker when using Atlas with AWS Aurora DSQL.

### Problem

Aurora DSQL does not support the `json`/`jsonb` data types or JSON functions like `json_agg()`. Atlas uses these internally for:

1.  **Schema introspection queries** - Atlas uses `json_agg()` or similar functions when inspecting the database schema.
2.  **Revision table** - The `atlas_schema_revisions` table has a `partial_hashes` column of type `jsonb`.

### Steps to Reproduce

```bash
atlas schema inspect \
    -u "postgres://admin:TOKEN@ENDPOINT:5432/postgres?sslmode=require"

atlas migrate apply \
    --url "postgres://admin:TOKEN@ENDPOINT:5432/postgres?sslmode=require" \
    --dir "file://migrations" \
    --skip-lock
```

### Error

```text
Error: sql/schema: pq: datatype json not supported
```

### Investigation

I confirmed DSQL doesn't support JSON:

```sql
SELECT json_agg(table_name) FROM information_schema.tables WHERE table_schema = 'public';
-- ERROR: datatype json not supported
```

I also tried pre-creating the revision table with `TEXT` instead of `jsonb` for `partial_hashes`, but the error persists because Atlas uses JSON functions in its introspection queries.

### Environment

- **Database:** Amazon Aurora DSQL (PostgreSQL-compatible, distributed)
- **Atlas CLI Version:** v1.0.1
- **Region:** eu-central-1

### Root Cause

The problematic code is in sql/postgres/inspect_oss.go lines 1698-1702:

```sql
  LEFT JOIN (
      select conindid, jsonb_object_agg(conname, contype) AS nametypes
      from pg_constraint
      group by conindid
  ) con ON con.conindid = idx.indexrelid
```

This subquery aggregates index constraints into a JSON object, which is then parsed in Go at line 567-574:

```go
  if sqlx.ValidString(constraints) {
      var m map[string]string
      if err := json.Unmarshal([]byte(constraints.String), &m); err != nil {
          return fmt.Errorf("postgres: unmarshaling index constraints: %w", err)
      }
      // ...
  }
```

### Suggested Fix

Atlas already has a pattern for handling PostgreSQL-compatible databases with limited features: CockroachDB support (sql/postgres/crdb_oss.go). The same pattern could be applied for DSQL.

### References

- https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-unsupported-features.html
- #3353 (advisory lock issue)
- #3539 (related DSQL issue)

alessandrolattao:
Add support for AWS Aurora DSQL via the `auroradsql://` URL scheme.

Aurora DSQL is PostgreSQL-compatible but lacks support for:
- JSON/JSONB types and functions
- Advisory locks (pg_try_advisory_lock)
- regoper type (used by Ent)
- DDL and DML in the same transaction

This implementation:
- Reuses CockroachDB inspection queries (no JSON)
- Implements custom RevisionReadWriter to bypass Ent
- Provides no-op Lock() method

Usage: atlas migrate apply --url "auroradsql://..." --tx-mode none

Closes #3649

### DIFF
--- a/sql/postgres/aurora_dsql_oss.go
+++ b/sql/postgres/aurora_dsql_oss.go
@@ -164,8 +164,11 @@ func (r *dsqlRevisionReadWriter) ReadRevision(ctx context.Context, version strin
 
 // WriteRevision saves the revision to the storage.
 func (r *dsqlRevisionReadWriter) WriteRevision(ctx context.Context, rev *migrate.Revision) error {
-	partialHashes, _ := json.Marshal(rev.PartialHashes)
-	_, err := r.db.ExecContext(ctx, fmt.Sprintf(
+	partialHashes, err := json.Marshal(rev.PartialHashes)
+	if err != nil {
+		return fmt.Errorf("marshaling partial hashes: %w", err)
+	}
+	_, err = r.db.ExecContext(ctx, fmt.Sprintf(
 		`INSERT INTO %s (version, description, type, applied, total, executed_at, execution_time, error, error_stmt, hash, partial_hashes, operator_version)
 		 VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12)
 		 ON CONFLICT (version) DO UPDATE SET


a8m:
Hey, please see my comment here: https://github.com/ariga/atlas/issues/3649#issuecomment-3758165198

Thanks for the suggestion, and I appreciate your willingness to contribute to Atlas.

Issue #3688, opened 2026-03-09
schema apply fails on PostGIS geography with dev-url in canary


elitan:
### Summary

`atlas schema apply --dry-run` fails during plan analysis when the schema contains `create extension if not exists postgis` and a column typed as `geography(point, 4326)`.

On the same machine, same database setup, and same schema:

- `v1.1.7-e3ed86f-canary` fails
- `v0.37.0` passes

This looks like a regression in newer Atlas builds.

### Minimal schema

```sql
create extension if not exists postgis;

create table geocoding_cache (
  id bigserial primary key,
  location geography(point, 4326)
);
```

### Repro

I reproduced this against a local PostgreSQL 17 database with PostGIS available.

```bash
atlas schema apply \
  -u "postgres://postgres:postgres@localhost:5432/target?sslmode=disable" \
  --to "file://schema.sql" \
  --dev-url "postgres://postgres:postgres@localhost:5432/dev?sslmode=disable" \
  --schema public \
  --dry-run
```

### Actual result

On `v1.1.7-e3ed86f-canary`:

```text
Error: analyzing plan: running first analysis: executing statement: CREATE TABLE "public"."geocoding_cache" (
  "id" bigserial NOT NULL,
  "location" public.geography(Point,4326) NULL,
  PRIMARY KEY ("id")
);: pq: type "public.geography" does not exist
```

### Expected result

The dry run should succeed and plan the table creation.

### Notes

- PostGIS is available in the database environment.
- The same repro passes with `v0.37.0`.
- I hit this from a normal `schema apply` flow that uses a separate `--dev-url` database for analysis.

If useful, I can share more environment details or test a candidate fix.

a8m:
Hey @elitan,
What's your PostGIS version? Also, are you using Atlas Pro?

a8m:
Issue might be related to the fact that you use extensions without Atlas Pro.

<img width="1573" height="980" alt="Image" src="https://github.com/user-attachments/assets/509258d4-0970-4d3b-8564-2e2cf8347f0e" />

a8m:
Please, reopen if you still encounter issues after login.

elitan:
yea that was it, was signed out for some reason. don't know why.

added this that might be helpful: https://github.com/ariga/atlas/pull/3689

PR #3689, opened 2026-03-09
fix(postgres): fail early on unsupported sql state

elitan:
## Summary
- fail early in OSS when Postgres SQL state contains unsupported objects
- cover both single SQL files and schema directories
- keep the change scoped to SQL state input paths

## Why
- `#3688` is real on the official canary build, but this repo builds the OSS/community path
- in OSS, SQL state could continue past unsupported Postgres objects and fail later with a confusing error
- failing early is better than emitting a broken plan or replaying partial state

## Example
- input SQL contains `create extension if not exists postgis;`
- before: replay could fail later with `type "public.geography" does not exist`
- after: OSS fails early and tells the user to use the official Atlas build or avoid that object in SQL state

## Notes
- this is a guardrail for the community build
- it does not claim to fix the official canary bug behind `#3688`

## Testing
- `cd cmd/atlas && go test ./internal/cmdext -count=1`
- `go test ./sql/postgres -count=1`


luantranminh:
Sorry @elitan, I have to close this PR because the pro feature check should not exist in the community version.

